PR #4, opened 2018-02-28
Quiz text

stoeffel:
This adds the first `Text.Quiz` text style. @bendansby is planning to add more.

screen shot 2018-02-28 at 14 18 18

This pr also upgrades elm-css in the styleguide-app.


src/Nri/Ui/Text/Quiz/V1.elm
```diff
@@ -0,0 +1,67 @@
...
+    "Nri-Ui-Text-Quiz-"
+
+
+type CssClasses
+    = Footnote
+
+
+{-| -}
+styles : Nri.Ui.Styles.V1.Styles Never CssClasses msg
+styles =
```
brookeangel: ❓ Should we make this in Html.Styled to make future css conversion easier? (If now is not the time, that's okay too!)

src/Nri/Ui/Text/Quiz/V1.elm
```diff
@@ -0,0 +1,67 @@
...
+            [ makeQuizFont (px 13) gray45
+            , lineHeight (px 18)
+            , fontWeight (int 400)
+            , margin4 (px 5) (px 0) (px 0) (px 0)
+            ]
+        ]
+
+
+makeQuizFont : Css.FontSize a -> Css.ColorValue b -> Style
+makeQuizFont size fontColor =
```
brookeangel: Oh,  nice! One question - do we want to call this all "Writing" instead of quiz? I think that may be the new terminology for this style

bendansby:
Yeah, Writing is a better term

stoeffel:
@brookeangel thanks for the review ❤️. I've renamed it to Writing and changed it to use `Html.Styled`.

PR #5, opened 2018-03-01
Update textarea 

jwoudenberg:
This updates `Nri.Ui.TextArea.V1` with the latest code from the monolith, including the auto-expanding behavior Luke added.

Although this is a breaking change of the TextArea API I didn't create a `V2`, because at the moment this module is not yet used in the monolith. The plan is to open a PR against the monolith swapping out its TextArea implementation for this one after this PR gets merged.


src/Nri/Ui/CssFlexBoxWithVendorPrefix.elm
```diff
@@ -0,0 +1,292 @@
+module Nri.Ui.CssFlexBoxWithVendorPrefix
```
stoeffel: I think it would be better to publish this as it's own module. This could  be useful to others as well. and maybe name it `Css.FlexBox`.

jwoudenberg:
> I think it would be good to convert elm-css to html.styled while we are at it.

I'd like these PR's to contain the smallest possible amount of changes to the modules pulled out from the monolith, for three reasons:
- I don't have the capacity at the moment to do bigger updates on all our widgets, so I'd like to keep these PR's focused on addressing the code-sharing concerns.
- Having the smallest possible diff with the monolith makes it easier to carry over last-minute changes being made to the widget in the monolith.
- Making major changes to the API of a widget here (like the switch to Html.Styled) will require extensive changes in all the places in the monolith making use of the new API. This will make it harder to get the counterpart of this PR in the monolith merged, which in turn will make it more likely the implementation of TextArea in this PR will diverge from the one in the monolith.

src/Nri/Ui/CssFlexBoxWithVendorPrefix.elm
jwoudenberg: Happy to split this out! One question: What's your motivation to drop `...WithVendorPrefix` from the name? To me `Css.FlexBox` gives the impression elm-css itself doesn't when really it does, just not with prefixes.

stoeffel:
Yes, this is totally reasonable!

src/Nri/Ui/CssFlexBoxWithVendorPrefix.elm
jwoudenberg: cc @stoeffel

PR #6, opened 2018-03-07
Butteroala  ug text styles

tesk9:
Adds helpers for user-generated text styles.

screen shot 2018-03-07 at 2 41 57 pm

@bendansby mediumBody & smallBody have margins on them by default (`margin4 (px 10) (px 0) (px 0) (px 0)` and `margin4 (px 7) (px 0) (px 0) (px 0)` respectively). Would you like for ugMediumBody and ugSmallBody to follow this pattern?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,19 @@
 # noredink-ui
 
 UI widgets we use.
+
+## Examples
+
+This repo contains an app showcasing all of these UI widgets.
+
+To see them locally:
+
+```
+> cd styleguide-app
+> elm-make Main.elm --output=elm.js
+```
+
+Open `index.html` in your browser.
+
+Alternatively, you may use elm-reactor. Please be aware that you'll need to globally
+install fonts (in particular, Muli) if you go this route.
--- a/src/Nri/Ui/Text/V1.elm
+++ b/src/Nri/Ui/Text/V1.elm
@@ -165,6 +165,9 @@ ugMediumBody =
             , fontSize (px 18)
             , lineHeight (px 30)
             , color gray20
+            , margin4 (px 10) (px 0) (px 0) (px 0)
+            , firstChild [ margin zero ]
+            , firstOfType [ margin zero ]
             ]
         ]
 
@@ -179,6 +182,9 @@ ugSmallBody =
             , fontSize (px 16)
             , lineHeight (px 25)
             , color gray20
+            , margin4 (px 7) (px 0) (px 0) (px 0)
+            , firstChild [ margin zero ]
+            , firstOfType [ margin zero ]
             ]
         ]
 
--- a/src/Nri/Ui/Text/V1.elm
+++ b/src/Nri/Ui/Text/V1.elm
@@ -13,28 +13,31 @@ module Nri.Ui.Text.V1
         , styles
         , subHeading
         , tagline
+        , ugMediumBody
+        , ugSmallBody
         )
 
-{-| Text types:
-
-@docs caption
-@docs heading
-@docs mediumBody
-@docs smallBody
-@docs smallBodyGray
-@docs subHeading
-@docs smallHeading
-@docs tagline
-
-Text class strings:
-
-@docs captionClassString
-@docs mediumBodyClassString
-@docs smallBodyClassString
+{-|
 
 @docs styles
 
-Modifying strings to display nicely:
+
+## Semantic text types:
+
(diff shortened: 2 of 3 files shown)


bendansby:
I think the margin is good, but first-child-of-type should have no margin!

README.md
```diff
@@ -1,3 +1,19 @@
 # noredink-ui
 
 UI widgets we use.
+
```
juanedi: This was super helpful, thank you!

src/Nri/Ui/Text/V1.elm
```diff
@@ -13,28 +13,31 @@ module Nri.Ui.Text.V1
         , styles
         , subHeading
         , tagline
+        , ugMediumBody
+        , ugSmallBody
         )
 
-{-| Text types:
+{-|
```
juanedi: 😻

src/Nri/Ui/Text/V1.elm
```diff
@@ -150,6 +155,34 @@ classString classes =
     String.join " " (List.map (identifierToString namespace) classes)
 
 
+{-| User-generated text.
+-}
+ugMediumBody : List (Html.Styled.Html msg) -> Html.Styled.Html msg
+ugMediumBody =
+    Html.Styled.p
+        [ css
+            [ quizFont
```
juanedi: ❓ should `Nri.Fonts` expose a separate `ugFont` even if it's also Georgia?
Not a blocker since that's in `nri-elm-css`, but would like to know what you think.

src/Nri/Ui/Text/V1.elm
```diff
@@ -150,6 +155,34 @@ classString classes =
     String.join " " (List.map (identifierToString namespace) classes)
 
 
+{-| User-generated text.
+-}
+ugMediumBody : List (Html.Styled.Html msg) -> Html.Styled.Html msg
```
juanedi: For the QW grading page we are using `white-space: pre-line` to display newlines.
Does that make sense for your use cases? I'm thinking it's a reasonable default.

src/Nri/Ui/Text/V1.elm
tesk9: Ooo that's a good idea. 

Also -- I kinda feel like nri-elm-css should merge into this module. Otherwise, updating a font (if we care that everything gets the new font at the same time) means:

- PR -> Review -> Merge -> Release `nri-elm-css`
- PR to make new V2 noredink-ui -> Review -> Merge -> Release `noredink-ui`
- Upgrade to new noredink-ui version & new nri-elm-css version
- Hope that font kerning doesn't mean that we have to change all the hardcoded widths everywhere 😱... so probably use new V2 noredink-ui version & maintain two font styles from nri-eml-css simultaneously?
- PR 🤞  -> Deploy ?

Anyway I'm pro ugFont!

src/Nri/Ui/Text/V1.elm
tesk9: Yeah that makes sense to me! @bendansby is there any case where user generated text **shouldn't** be whitespace preline?

bendansby: Nothing comes to mindx

src/Nri/Ui/Text/V1.elm
```diff
@@ -150,6 +155,42 @@ classString classes =
...
+    Html.Styled.p
+        [ css
+            [ quizFont
+            , fontSize (px 16)
+            , lineHeight (px 25)
+            , whiteSpace preLine
+            , color gray20
+            , margin4 (px 7) (px 0) (px 0) (px 0)
+            , firstChild [ margin zero ]
+            , firstOfType [ margin zero ]
```
tesk9: @bendansby is this what you wanted? first-child or first-of-type?

bendansby: Whoops, guess I asked for a property that doesn't exist 😅 

That looks good, though!

tesk9: But one that should exist! I would like for first-of-class to be a thing too

src/Nri/Ui/Text/V1.elm
tesk9: Sweet! added :)

src/Nri/Ui/Text/V1.elm
juanedi: Sounds great! Maybe we should bring it up next 🤗& 🐛?

PR #9, opened 2018-03-16
Butteroala  condense with nri elm css

tesk9:
This PR does a coupla things:

1. Adds `Nri.Ui.Fonts.V1`
2. Adds `Nri.Ui.Colors.V1`
3. Adds `Nri.Ui.Colors.Extra` <- note, not versioned at the module level. Lmk if you think it should be! 
4. Removes `nri-elm-css` dependency
    - [Nri.Accessibility.invisibleText](https://github.com/NoRedInk/nri-elm-css/blob/5.0.1/src/Nri/Accessibility.elm#L22) <- place where I'm most likely to have made a mistake. Also this change led to an API change. It looked like we were ignoring a class that we were setting via a bool. ~~i thought it made sense to remove the unused bool. Lmk if you think this change should happen independently, or if you think I should use the module-level versioning!~~ used module level versioning and added `Nri.Ui.Checkbox.V2`
    - [Nri.Stylers.makeFont](https://github.com/NoRedInk/nri-elm-css/blob/5.0.1/src/Nri/Stylers.elm#L16) I replaced with multiple lines of fontsize/font family/color.
    - find and replace `Nri.Colors` -> `Nri.Ui.Colors.V1`
    - find and replace `Nri.Colors.Extra` -> `Nri.Ui.Colors.Extra`
    - find and replace `Nri.Fonts` -> `Nri.Ui.Fonts.V1`
5. Fixes double listing of elm-css in the styleguide app elm-package.json.
6. Bumps package version.


elm-package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "3.1.1",
+    "version": "4.0.0",
```
jwoudenberg: 💭 The last major version bump to 3.x.x prompted a short discussion on Slack: https://noredink.slack.com/archives/C0VVDLEES/p1520961722000597

The suggestion there was to reserve future major version bumps for language changes, or removal of widget versions that are no longer used anywhere in the NRI codebase. So I think this would be a good idea:

> Also this change led to an API change. It looked like we were ignoring a class that we were setting via a bool. i thought it made sense to remove the unused bool.

Also, if we feel the major versions policy above doesn't make sense we should change it, and if we do feel it makes sense we should document it. I'm happy to own either depending on your feeling on it with regards to this PR.

src/Nri/Ui/InputStyles.elm
```diff
@@ -106,61 +105,64 @@ styles =
...
             , class Label
-                [ backgroundColor Nri.Colors.white
+                [ backgroundColor white
                 , left (px 10)
                 , top (px 0)
                 , padding2 zero (px 5)
-                , Nri.Stylers.makeFont (px 12) navy
+                , fontSize (px 12)
+                , Nri.Ui.Fonts.V1.baseFont
+                , color navy
```
jwoudenberg: 😻

styleguide-app/Examples/Colors.elm
```diff
@@ -0,0 +1,116 @@
+module Examples.Colors exposing (example)
```
jwoudenberg: 😻

styleguide-app/elm-package.json
```diff
@@ -9,15 +9,13 @@
     ],
     "exposed-modules": [],
     "dependencies": {
-        "NoRedInk/nri-elm-css": "5.0.0 <= v < 6.0.0",
```
jwoudenberg: 🎉 🏆

elm-package.json
tesk9: Interesting!  🤔 it seems a little weird to limit versioning such that:

- we keep the unused bool in V1 & release without the bool as V2
- we swap all the monolith/content creation uses over to V2
- we kill the V1 version and do a major release

I think the major release at removal looks like work that won't ever happen/will be a chore. Like, after I update upgrade the monolith's dependency on this package, is it then on me to revisit this library and see if I can remove V1, even though I have no idea what content creation's needs are? I guess that's more of a question of general long term package management plans... heyo internal! 😄

For now, I'll add the bool back in and get that version number back to a minor change. 

Thanks!

tesk9: Done with the Checkbox.v2 change!

What do you think about Nri.Colors.Extra not having module-level versioning? should it?

jwoudenberg: Some context you might already have: I believe the primary concern behind having per-widget versioning and disallowing breaking changes is preventing this scenario:

1. I make a breaking change of noredink-ui. 
2. You'd like to make a minor change to some widget in noredink-ui you need for a story you're working on in the monolith. You're now blocked from using this change until the monolith has been changed to deal with the unrelated breaking changes introduced in (1). Ideally I already created a PR for that, but that PR might still take a while getting merged.

Granted, for this PR that doesn't seem a huge concern, as the 'breaking' change is super straightforward.

Good question with regards to who's responsible for taking old versions out of noredink-ui again. Perhaps we simply shouldn't care about those old versions sticking around. After all we also don't care about old versions of libraries sticking around on `elm package` or `npm`. Technically those repositories would be leaner if versions of packages no-one uses anymore get removed, but nobody seems to  think setting that up is worth the effort.

> What do you think about Nri.Colors.Extra not having module-level versioning? should it?

I think it's fine the way it is now. If we ever get to the point we want to make a breaking change in that module, we can decide at that point whether to start versioning the module after all or make a breaking release.

tesk9: I'm very happy to not care about old versions! Maybe we can remove old versions when there's a new elm version or only if we are working on managing dependencies?

Thanks! Will keep `Nri.Ui.Colors.Extra` as is.

tesk9:
I think I'm ready for another review, @jwoudenberg :)

PR #11, opened 2018-03-24
Add `SegmentedControl`

joneshf:
<details>
<summary>Diff from Monolith</summary>

```diff
1c1
< module Nri.SegmentedControl exposing (Config, Option, styles, view)
---
> module Nri.Ui.SegmentedControl.V1 exposing (Config, Option, styles, view)
12d11
< import Css.FlexBoxWithVendorPrefix as FlexBox
17,20c16,20
< import Nri.Colors as Colors
< import Nri.Colors.Extra exposing (withAlpha)
< import Nri.Stylers
< import Nri.Styles
---
> import Nri.Ui.Colors.Extra exposing (withAlpha)
> import Nri.Ui.Colors.V1 as Colors
> import Nri.Ui.CssFlexBoxWithVendorPrefix as FlexBox
> import Nri.Ui.Fonts.V1
> import Nri.Ui.Styles.V1
67c67
< styles : Nri.Styles.Styles Never CssClass msg
---
> styles : Nri.Ui.Styles.V1.Styles Never CssClass msg
69c69
<     Nri.Styles.styles "Nri-SegmentedControl-"
---
>     Nri.Ui.Styles.V1.styles "Nri-Ui-SegmentedControl-V1-"
77c77,79
<             , Nri.Stylers.makeFont (px 15) Colors.azure
---
>             , Nri.Ui.Fonts.V1.baseFont
>             , fontSize (px 15)
>             , color Colors.azure
```
</details>

<details>
<summary>`elm-package diff` output</summary>

```
Comparing NoRedInk/noredink-ui 3.2.0 to local changes...
This is a MINOR change.

------ Added modules - MINOR ------

    Nri.Ui.SegmentedControl.V1


```
</details>

<details>
<summary>Screenshot of style guide</summary>

screen shot 2018-03-23 at 19 06 51
</details>

##### Questions

1. We're not currently depending on `nri-elm-css`, so we don't have `Nri.Stylers`. I seemed to remember that we're moving away from that. Is that right? I inlined what `Nri.Stylers.makeFont` is in the meantime. If that's not the case, I'm down to add the dependency and use it instead. If we want `Nri.Ui.Stylers.V1` here, I'm down to move it here instead.
1. In the screenshot, the control has a good deal of height. The styles are the same as what's in the monolith, but the monolith's control doesn't have as much height. Any idea what might be up?

    <details>
    <summary>Monolith's height</summary>

    screen shot 2018-03-23 at 19 09 59
    </details>

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -21,6 +21,7 @@
         "Nri.Ui.Modal.V1",
         "Nri.Ui.Outline.V1",
         "Nri.Ui.Palette.V1",
+        "Nri.Ui.SegmentedControl.V1",
         "Nri.Ui.Styles.V1",
         "Nri.Ui.Tabs.V1",
         "Nri.Ui.Text.Writing.V1",
--- a/src/Nri/Ui/SegmentedControl/V1.elm
+++ b/src/Nri/Ui/SegmentedControl/V1.elm
@@ -91,6 +91,7 @@ styles =
             , border3 (px 1) solid Colors.azure
             , borderLeft (px 0)
             , borderBottom3 (px 3) solid Colors.azure
+            , boxSizing borderBox
             ]
         , Css.Foreign.class Focused
             [ color Colors.gray20
--- a/src/Nri/Ui/SegmentedControl/V1.elm
+++ b/src/Nri/Ui/SegmentedControl/V1.elm
@@ -1,0 +1,101 @@
+module Nri.Ui.SegmentedControl.V1 exposing (Config, Option, styles, view)
+
+{-|
+
+@docs Config, Option, styles, view
+
+-}
+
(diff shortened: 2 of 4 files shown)


jwoudenberg:
> We're not currently depending on nri-elm-css, so we don't have Nri.Stylers. I seemed to remember that we're moving away from that. Is that right? I inlined what Nri.Stylers.makeFont is in the meantime. If that's not the case, I'm down to add the dependency and use it instead. If we want Nri.Ui.Stylers.V1 here, I'm down to move it here instead.

`nri-elm-css` has been subsumed into `noredink-ui` as part of #9. `Stylers` didn't make the cut and has been split up into its components like you did here. In other words, everything is awesome.

PR #12, opened 2018-03-26
Add Nri.Ui.styled

brookeangel:
This adds Nri.Ui.styled, a handy helper for adding a data attribute to styled html!


rtfeldman:
Looks great! I agree that this shouldn't have a `V1`.

PR #13, opened 2018-03-26
Add forgotten module

brookeangel:
Forgot to add `Nri.Ui` to exposed modules! This PR should do the trick, so that `styled` is exposed.


rtfeldman:
Whoops! Thanks @brookeangel

PR #15, opened 2018-03-27
Add elm tests

jwoudenberg:
Of the widgets we have in here at the moment, `Checkbox` is the only one with a separate spec file in the monolith. This PR adds that suite and sets us up for adding more specs in the future.

This commit only touches specs, so a release after merging this is unnecessary.


jwoudenberg:
The build is failing because bintray is having a major outage.

PR #18, opened 2018-03-27
make Nri.Ui.styled compatible with Accessibility.Styled

brookeangel:
Before: could not use `Nri.Ui.styled` with `Accessibility.Styled`
Now: you can, because the type signature is compatible!


brookeangel:
Here's one more for you, @rtfeldman !

brookeangel:
Let me know if you don't have time for this and I'll assign this to someone else 😄 - maybe @avh4 since I paired with him on this fix!

rtfeldman:
Makes sense, and looks good! 🤘

PR #20, opened 2018-03-28
add a Makefile, speed up CI

BrianHicks:
how do you make your CI faster? You don't let it get slow in the first place! There are a few optimizations we can do the the travis configuration to speed it up a bit. This takes us from around 4 minutes to ~around 2~ 1m15s for a warm build.

It does this with a `Makefile` as well… now you can go from `git clone` to "I have working software" in one command, if you have `npm` and `make` installed.

(P.S. the branch is called makefile-uncontroversial because I did some weeeeird stuff in another branch! 😅)

| cold cache | hot cache | % decrease |
|--|--|--|
| 2m4s | 1m15s | 39.6% |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ node_js:
 
 cache:
   directories:
-    - node_modules
     - elm-stuff/build-artifacts
     - tests/elm-stuff/build-artifacts
     - styleguide-app/build-artifacts
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 sudo: false
 
+language: node_js
+node_js:
+  - node # latest stable
+
 cache:
   directories:
     - tests/elm-stuff/build-artifacts
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,20 +16,9 @@ before_install:
       make && make install;
       cd ..;
     fi
+
 install:
-  - npm install -g elm@0.18.0 elm-test elm-format@exp
-  - mv $(npm config get prefix)/bin/elm-make $(npm config get prefix)/bin/elm-make-old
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,8 +1,15 @@
 sudo: false
 
+language: node_js
```
michaelglass: what does this change do? the image apparently had npm before

.travis.yml
```diff
@@ -1,8 +1,15 @@
...
+node_js:
+  - node # latest stable
+
 cache:
   directories:
-    - tests/elm-stuff/build-artifacts
+    - node_modules
+    - elm-stuff
+    - tests/elm-stuff
+    - styleguide-app/elm-stuff
```
michaelglass: checks out.

how do we clear the cache? We've had issues with `elm-stuff` being corrupted before.  Curious about how we fix that problem.

package.json
```diff
@@ -0,0 +1,27 @@
...
+    "url": "git+https://github.com/NoRedInk/NoRedInk-ui.git"
+  },
+  "author": "",
+  "license": "BSD-3-Clause",
+  "bugs": {
+    "url": "https://github.com/NoRedInk/NoRedInk-ui/issues"
+  },
+  "homepage": "https://github.com/NoRedInk/NoRedInk-ui#readme",
+  "dependencies": {
+    "elm": "^0.18.0",
```
michaelglass: 👍

.travis.yml
BrianHicks: it has NPM now! But now it also runs in a Docker container where it won't try to install RVM and different versions of PHP and all that. This just specialized it to be a JavaScript build system.

.travis.yml
BrianHicks: https://travis-ci.org/NoRedInk/noredink-ui/caches (find it under the build page -> more options -> caches)

you can just drop 'em there with a click. Anyone who's a repo owner can do that AFAICT.

PR #23, opened 2018-03-29
Separate icon functions

jwoudenberg:
#### Changes
This introduces a new version of the `Icon` module in which the `IconType` union has been removed in favor of separate icon constructor functions. This has a number of advantages:

- Rendering a single Icon no longer requires a record containing the assets of all icons. This means services such as CCS only need to provide assets for Icons they actually use.
- Adding a new icon to the module no longer causes a breaking change in `noredink-ui`. This wasn't an issue while `Nri.Icons` lived in the monolith, but as part of a separate package adding an exposed constructor to a union type causes a breaking change.

This PR also introduces a new version of `SegmentedControl`, which is the only widget in `noredink-ui` currently using Icons.

#### Review
Asking a review from Luke because I know you had a large hand in setting up `Nri.Icon`. Also asking a review from Hardy, specifically on whether you feel this approach benefits CCS development. Please feel completely free to remove yourself from the review list for any reason!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,9 +91,6 @@ anaconda-mode/
 .AppleDouble
 .LSOverride
 
-# Icon must end with two \r
-Icon
-
 # Thumbnails
 ._*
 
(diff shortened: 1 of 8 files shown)


jwoudenberg:
Thanks for the reviews!

PR #43, opened 2018-04-25
Add guidedWrite icon

juanedi:
For https://www.pivotaltracker.com/story/show/156937580.

Asset added in monolith here: https://github.com/NoRedInk/NoRedInk/pull/19874

### DIFF
--- a/src/Nri/Ui/Icon/V2.elm
+++ b/src/Nri/Ui/Icon/V2.elm
@@ -43,6 +43,7 @@ module Nri.Ui.Icon.V2
         , gardening
         , gear
         , greenCheckMark
+        , guidedWrite
         , help
         , highFive
         , icon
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/Icon/V2.elm
```diff
@@ -533,6 +535,12 @@ greenCheckMark assets =
     ImgIcon assets.smallCheckmark_png
 
 
+{-| -}
+guidedWrite : { r | icons_guidedWrite_svg : Asset } -> IconType
+guidedWrite assets =
+    ImgIcon assets.icons_guidedWrite_svg
```
brookeangel: ❓ Looks like QuickWrite is also an ImgIcon... do you know what the pattern is for when something is supposed to be an image and when something is supposed to be an svg? I'm assuming that we want an image here because we want to use the given colors or something?

juanedi: That does sound like a good theory! TBH I'm not sure what the answer is (just copied from quick writes).

Will ask on Slack!

PR #44, opened 2018-04-25
Introduce Nri.Ui.Table.V2

jwoudenberg:
#### Changes
This adds a new version of the `Nri.Ui.Table` module with the following powers:

- It allows column widths to be specified using any css-length value. Previously we could only use pixels.
- Since we can now use relative widths in column sizes, tables take up the full available width by default.
- Styles have been ported to use `Html.Styled`.
- The default cell padding has been removed, because some designs like [this](https://paper.dropbox.com/doc/Product-guide-Content-creation-index-p1wpha9HkFp5HChwicL8h) do not want such large paddings, and overriding the default paddings in such instances is hard.

#### Review asks
- Is there a danger in allowing relative column sizes I'm overlooking?
- The `keyframes` styles are still exported separately. Can / should we include these with the other styles somehow?

### DIFF
--- a/src/Nri/Ui/Table/V2.elm
+++ b/src/Nri/Ui/Table/V2.elm
@@ -11,7 +11,20 @@ module Nri.Ui.Table.V2
         , viewWithoutHeader
         )
 
-{-|
+{-| Upgrading from V1:
+
+  - All the `width` fields in column configurations now take an elm-css length
+    value rather than an Integer. Change `width = 100` to `width = px 100` to get
+    the same widths as before.
+  - Tables now by default take the full width of the container they are placed in.
+    If this is not what you want, wrap the table in an element with a fixed width.
+  - The table module now makes use of `Html.Styled` and no longer exposes a
+    separate `styles` value.
+    Check out the [elm-css](http://package.elm-lang.org/packages/rtfeldman/elm-css/14.0.0/Html-Styled)
+    documentation on Html.Styled to see how to work with it.
+  - The default cell padding has been removed and content is not vertically
+    centered in its cell. If you need to overwrite this, wrap your cells in
+    elements providing custom styling to the cell.
 
 @docs Column, custom, string
 
--- a/src/Nri/Ui/Table/V2.elm
+++ b/src/Nri/Ui/Table/V2.elm
@@ -5,7 +5,6 @@ module Nri.Ui.Table.V2
         , keyframeStyles
         , keyframes
         , string
-        , styles
         , view
         , viewLoading
         , viewLoadingWithoutHeader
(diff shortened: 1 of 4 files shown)


src/Nri/Ui/Table/V2.elm
```diff
@@ -0,0 +1,281 @@
...
+type alias WidthStyle =
+    List ( String, String )
+
+
+{-| A column that renders some aspect of a value as text
+-}
+string :
+    { header : String
+    , value : data -> String
+    , width : LengthOrAuto compatible
```
ento: > Is there a danger in allowing relative column sizes I'm overlooking?

Not really a danger, but people might find that they now want to set `table-layout: fixed` on the table element to, perhaps tweak the layout to have a column auto-expand while keeping the widths of the rest fixed (["Example 3: Mixed unit widths" of this article](https://adamschwartz.co/magic-of-css/chapters/3-tables/)) - which I don't think is a need that should be addressed now.

src/Nri/Ui/Table/V2.elm
```diff
@@ -0,0 +1,281 @@
+module Nri.Ui.Table.V2
+    exposing
+        ( Column
+        , custom
+        , keyframeStyles
+        , keyframes
```
ento: > The keyframes styles are still exported separately. Can / should we include these with the other styles somehow?

I'm not sure if we can without wrapping the table and the style element in a div or something. [There doesn't seem to be a way of defining keyframes as an inline style](https://www.w3.org/TR/css-animations-1/#keyframes). Maybe a div wrapper isn't so bad?

I also feel like it can wait til [elm-css gains first-class animation support](https://github.com/rtfeldman/elm-css/issues/313)

src/Nri/Ui/Table/V2.elm
```diff
@@ -0,0 +1,281 @@
...
+    , color gray45
+    , pseudoClass "nth-child(odd)"
+        [ backgroundColor gray96 ]
+    ]
+
+
+cellStyles : List Style
+cellStyles =
+    [ verticalAlign middle
+    ]
```
ento: 💭 we could have a section in the doc comment at the top about how to migrate from V1 to V2: specify your own padding, perhaps wrap in a fixed-width container, look out for mismatch in vertical align, etc.

src/Nri/Ui/Table/V2.elm
jwoudenberg: Makes sense! Going with the option to hold of on this until elm-css has first-class support.

src/Nri/Ui/Table/V2.elm
jwoudenberg: I love this! Added a section like you suggested.

src/Nri/Ui/Table/V2.elm
jwoudenberg: Can see the use case for this, but going to hold of on this for now!

PR #48, opened 2018-05-02
bump elm to resolve security exception

michaelglass:
the old elm depended on a package that raised a security warning from github

the new elm does not, however, [its binwrap isn't compatible with node 10](https://github.com/avh4/binwrap/issues/11)

so: let's not merge this until the binwrap issue [is resolved upstream.](https://github.com/avh4/binwrap/pull/12)


stoeffel:
@michaelglass can we close this?

PR #49, opened 2018-05-02
test against node 9, too

michaelglass:
node 10 was just released but we use node 9 widely, still

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 
 language: node_js
 node_js:
+  - "9"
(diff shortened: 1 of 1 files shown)


jwoudenberg:
This change will run the same build using node 9 and 10, correct?

Considering we're currently only using `npm` to install some dependencies for CI (rather than dependencies for the package itself), I wonder if that's worth doubling CI times over.

What risk do we want to protect ourselves against with this change? Broken dev environments?

michaelglass:
I buy it.

PR #53, opened 2018-05-09
Add max width on details to improve readability

terezka:
The details in this page are for showing decoder errors and such when the page is broken. Before the text took up the whole screen while now, it is limited to a max with of 700px, so the sentences are easier to read.


michaelglass:
@terezka can you ferry this forward?

PR #54, opened 2018-05-10
Add link spa to icon module

terezka:
This adds an icon link option for SPAs.

Almost a copy paste of the `linkSpa` in `Button`!


terezka:
@jwoudenberg Do I just release it myself like normal or is there something I should do before?

jwoudenberg:
> Do I just release it myself like normal or is there something I should do before?

Nothing special!

PR #57, opened 2018-05-12
Add custom button

michaelglass:
for broad to text questions, we want to be able to focus on an element. The Dom package only focuses on elements with an ID. So I want to add the ability to pass Id (or random attributes) to button. Perhaps it'd be better to constrain this to just id.


jwoudenberg:
Published as part of 5.1.0

PR #59, opened 2018-05-17
Add V2 of Nri.Ui.Text using Html.Styled

tesk9:
- adds new version of the text module using Html.Styled instead of an elm css stylesheet
- updates the styleguide example app to use the new text module version

screen shot 2018-05-16 at 5 09 52 pm


jwoudenberg:
Published as part of 5.1.0

PR #63, opened 2018-05-22
Textarea v2 adding a minimum height

jwoudenberg:
#### What is this

This is a new version of the `TextArea` widget. It adds the option to set a minimum height for text areas and does some initial work in porting the widget to `Html.Styled` (API is changed, internals still rely on some old fashion styling).

### DIFF
--- a/src/Nri/Ui/InputStyles.elm
+++ b/src/Nri/Ui/InputStyles.elm
@@ -1,8 +1,13 @@
-module Nri.Ui.InputStyles exposing (Assets, CssClasses(..), styles)
+module Nri.Ui.InputStyles exposing (Assets, CssClasses(..), inputLineHeight, inputPaddingVertical, styles, textAreaHeight, writingLineHeight, writingMinHeight, writingPadding, writingPaddingTop)
 
 {-|
 
 @docs styles, CssClasses
+
+
+## Shared hardcoded values
+
+@docs inputPaddingVertical, inputLineHeight, textAreaHeight, writingLineHeight, writingPadding, writingPaddingTop, writingMinHeight
 
 -}
 
@@ -60,7 +65,7 @@ styles =
         inputStyle =
             [ border3 (px 1) solid gray75
             , borderRadius (px 8)
-            , padding2 (px 8) (px 14)
+            , padding2 inputPaddingVertical (px 14)
             , property "transition" "all 0.1s ease"
             , pseudoClass "placeholder"
                 [ color gray45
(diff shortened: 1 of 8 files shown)


src/Nri/Ui/InputStyles.elm
```diff
@@ -246,3 +251,38 @@ type alias Assets r =
         , icons_searchGray_svg : Asset
         , icons_xBlue_svg : Asset
     }
+
+
+inputPaddingVertical : Px
+inputPaddingVertical =
+    px 8
```
jwoudenberg: 😻

src/Nri/Ui/TextArea/V2.elm
```diff
@@ -50,26 +51,31 @@ type alias Model compatible msg =
...
     , label : String
     , showLabel : Bool
-    , minimumHeight : LengthOrMinMaxDimension compatible
+    , minimumHeight : MinimumHeight
     }
 
 
+type MinimumHeight
+    = DefaultHeight
+    | SingleLine
```
jwoudenberg: 💯

styleguide-app/assets/TextArea.js
```diff
@@ -43,7 +43,7 @@ Object.defineProperties(AutoresizeTextArea.prototype, {
   }
 });
 
-AutoresizeTextArea.prototype._onInput = function() {
+AutoresizeTextArea.prototype._resize = function() {
```
jwoudenberg: Thanks for pulling in these changes from the `TextArea.js` files from the monolith.

src/Nri/Ui/TextArea/V2.elm
```diff
@@ -45,15 +44,23 @@ type alias Model msg =
...
-    , minimumHeight : MinimumHeight
     }
 
 
-type MinimumHeight
+{-| Control whether to auto-expand the height.
+-}
+type HeightBehavior
+    = Fixed
+    | AutoResize Height
```
jwoudenberg: 😻 Even better!

src/Nri/Ui/TextArea/V2.elm
```diff
@@ -211,15 +211,7 @@ calculateMinHeight textAreaStyle specifiedHeight =
                     singleLineHeight
 
         DefaultHeight ->
-            case textAreaStyle of
-                DefaultStyle ->
-                    InputStyles.textAreaHeight
-
-                WritingStyle ->
-                    InputStyles.writingMinHeight
```
ento: A Writing textarea is still specified to have a min-height of 150px: https://github.com/NoRedInk/noredink-ui/pull/65/files#diff-603eb374d349e2a301c6d44684f6016fR144

Should we use that 150px here as well, or was the distinction of writing vs other styles no longer needed when it's set to auto-resize?

jwoudenberg: Good catch, thanks! Will push a fix.

jwoudenberg: Added back the height of 150px for text areas. Also checked the styleguide app to see the min-height is applied correctly. Seems to look good!

ento:
published as 5.3.0

PR #64, opened 2018-05-23
Styleguide app: load custom-elements's native-shim for TextArea.js

ento:
I was seeing this error when I checked the checkbox for making textareas autoresize in the styleguide app:

> Uncaught TypeError: Failed to construct 'HTMLElement': Please use the 'new' operator, this DOM object constructor cannot be called as a function.

The [README of custom-elements explains why this is happening](https://github.com/webcomponents/custom-elements#es5-vs-es2015). Following the advice there, this PR loads native-shim.js for ES5 <> custom elements compatibility.


michaelglass:
this seems fine. Consider having the dependency local so we can use this without internet?

ento:
> Consider having the dependency local so we can use this without internet?

hm, I buy that. Added back custom-elements.min.js. This could be npm/bowerified, but given the number and size of the external dependencies checked in, this may not yet be the time.

PR #69, opened 2018-06-02
WIP improve textarea and fix styling

lukewestby:
Instead of having the custom element create a new textarea and then copy all of the attributes we can just wrap an existing textarea with a custom element and optionally have it attach or not. this also degrades more gracefully as the textarea will show up no matter what and the wrapping custom element will just be ignored by browsers that can't even use the polyfill.


lukewestby:
I'm closing this in favor of #70 which includes distributing the custom element via npm

PR #70, opened 2018-06-04
Updates to TextArea

lukewestby:
With this PR we will:

1. Rework the TextArea implementation to be easier to style and just generally maintain
2. Publish the TextArea and any future custom elements so that they can be consumed by the monolith and CCS in a consistent way

This resolves the observed issue in #68. It turns out that it isn't necessary to manually create a `textarea` element and copy attributes to it. It's okay for a custom element to have children in the virtual DOM as long as the custom element does not modify any of the children in a way that causes a discrepancy between the virtual DOM's understanding and the actual DOM. Since the call to `Html.textarea` does not include a `style` attribute it's fine for the `nri-textarea-v3` custom element to modify its style directly to do the resizing.

The way that I tested this was by manually copying the `Nri.Ui.TextArea.V3` implementation to the monolith and then installing `@noredink/ui` from a git url. There is a (very large) GIF below for verification.

Follow up steps:
- Publish `@noredink/ui` to NPM
- Submit a PR to the monolith to install `@noredink/ui` and substitute any imports of `Nri/TextArea` with `@noredink/ui`

<details>
<summary>GIF</summary>

</details>


lib/TextArea.js
```diff
@@ -0,0 +1,53 @@
...
+    _updateListener: function() {
+      if (this._autoresize) {
+        this._textarea.addEventListener('input', this._resize)
+        this._resize()
+      } else {
+        this._textarea.removeEventListener('input', this._resize)
+      }
+    },
+
+    _resize: function() {
```
ento: I made changes to the resize logic here: https://github.com/NoRedInk/noredink-ui/pull/63/files#diff-b310ebe2d8a713781bb80d92660737f1R51 (which are missing from this version)

- look at the computed min-height in addition, as it's possible min-height is specified through a class
- reset to minHeight instead of `auto`, as `auto` insisted on adding some padding to `scrollHeight` when minHeight is only a single line's worth
  - set transition to none as I couldn't figure out why the textarea started doing animation after changing the above when it didn't before

lukewestby: included, thank you!

README.md
```diff
@@ -18,7 +18,7 @@ We try to avoid breaking changes and the associated major version bumps in this
 
 Suppose you just released version `5.0.1`, a small styling fix in the checkbox widget, for a story you're working on. If the project you're working in currently pulls in `noredink-ui` at version `4.x`, then getting to your styling fix means pulling in a new major version of `noredink-ui`. This breaks all `TextArea` widgets across the project, so those will need to be fixed before you can do anything else, potentially a big effort.
 
-To prevent these big Yaks from suddenly showing up in seemingly trivial tasks we prefer to avoid breaking changes in the package. Instead when we need to make a breaking change in a widget, we create a new module for it `Nri.Ui.MyWidget.VX`.
+To prevent these big Yaks from suddenly showing up in seemingly trivial tasks we prefer to avoid breaking changes in the package. Instead when we need to make a breaking change in a widget, we create a new module for it `Nri.Ui.MyWidget.VX`. Similarly, when we build custom elements in JavaScript we create a file `lib/MyWidget/VX.js` and define a custom element `nri-mywidget-vX`.
```
brookeangel: Read me to-do discussed IRL: add line about `rm styleguide-app/assets/javascript.js` to force a rebuild!

ento: Can be added to the `make clean` target as well: https://github.com/NoRedInk/noredink-ui/blob/15adb4bd2a517e8f7b7c27f851128eeba9b35e64/Makefile#L18-L19

(`make clean` will force a rebuild of everything though, as opposed to just the custom elements)

That reminds me: if we're checking in the build artifact, should we add a CI check that confirms `javascript.js` is up-to-date? something like what our internal build script does:

```sh
if git status --porcelain | grep . ; then
    echo "your build left non-gitignored trash behind"
    git diff --no-color
    exit 1
fi
```

brookeangel: Ooh, that sounds like a good idea!

src/Nri/Ui/TextArea/V3.elm
```diff
@@ -88,61 +88,38 @@ contentCreation model =
...
-                   the textarea that will trigger this event on the autoresize-textarea element itself. See AutoresizeTextArea.prototype._onInput
-                   and AutoresizeTextArea.prototype.connectedCallback for an example pertaining to the `input` event
-                   - When adding a new Html.Attribute that is an _attribute_, you don't have to do anything. All attributes are
-                   automatically reflected onto the textarea element via AutoresizeTextArea.prototype.attributeChangedCallback
-                -}
-                Html.node "autoresize-textarea"
-                    (sharedAttributes
-                        ++ [ -- setting the default value via a text node doesn't play well with the custom element,
-                             -- but we'll be able to switch to the regular value property in 0.19 anyway
-                             Attributes.defaultValue model.value
```
brookeangel: > setting the default value via a text node doesn't play well with the custom element, but we'll be able to switch to the regular value property in 0.19 anyway

❓ Oh! Is this no longer the case? (Seems fine!) What fixed it?

src/Nri/Ui/TextArea/V3.elm
```diff
@@ -88,61 +88,38 @@ contentCreation model =
...
+            autoresizeAttrs
+            [ Html.styled Html.textarea
                 [ InputStyles.input theme model.isInError
+                , Css.boxSizing Css.borderBox
+                , case model.height of
+                    AutoResize minimumHeight ->
+                        Css.minHeight (calculateMinHeight theme minimumHeight)
+
+                    Fixed ->
+                        Css.batch []
```
brookeangel: I think that this actually got introduced in a previous PR, but it looks like a height attribute got deleted here so the textareas are super squashy before they expand!

screen shot 2018-06-05 at 4 12 34 pm

In the monolith styles, the "Standard" and "ContentCreation" text areas are minheight 100px (see `Nri/Inputs/Styles.elm` line 97), and the "Writing" text area is 150px. Let me know if you wanna fix it here or if I should make a followup PR to fix!

brookeangel: Done over here: https://github.com/NoRedInk/noredink-ui/pull/71

brookeangel:
👋 Hi Marica & Luke! Are there any more fixes needed on this branch?

README.md
lukewestby: I actually think this artifact is fine to gitignore, and that seems like the least complicated way forward.

src/Nri/Ui/TextArea/V3.elm
lukewestby: In the previous setup we weren't inserting a textarea under the `autoresize-textarea` element, so we had to do something different in order to get the default value in the textarea that it eventually created. In the current setup the textarea is always created by the virtual DOM so we can use the text node in both cases.

lukewestby:
@brookeangel i think this is okay to merge now once it passes ci. marica is taking some time off this week

brookeangel:
Thanks!!!

PR #72, opened 2018-06-12
Render markdown in button labels

stoeffel:
We want to support markdown in buttons for Outline MC/Draggable questions. 

### Context
Currently they use `Nri.Button.submit` which takes html, to render markdown. We're moving away from submitting the answer as a form submit, that's way we need to switch to `Nri.Button.button`.

buttons-support-markdown

cc @bendansby

### DIFF
--- a/src/Nri/Ui/Button/V2.elm
+++ b/src/Nri/Ui/Button/V2.elm
@@ -71,6 +71,8 @@ import Html.Attributes exposing (..)
 import Html.Events exposing (onClick)
 import Html.Styled
 import Json.Decode
+import Markdown.Block
+import Markdown.Inline
 import Nri.Ui.AssetPath as AssetPath exposing (Asset)
 import Nri.Ui.Colors.Extra exposing (withAlpha)
 import Nri.Ui.Colors.V1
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/Button/V2.elm
```diff
@@ -595,17 +597,29 @@ buttonClass size width colorPalette =
...
-                , Html.text label
-                ]
+            [ span [ styles.class [ LabelIconAndText ] ]
+                (decorativeIcon iconType
+                    :: renderMarkdown label
+                )
+            ]
+
+
+renderMarkdown : String -> List (Html msg)
```
stoeffel: 📝 this is used in https://github.com/NoRedInk/elm-formatted-text/blob/master/src/FormattedText/Markdown.elm#L62 as well

jwoudenberg:
Wondering if we should allow buttons (or a button) to take any `Html` content instead.

stoeffel:
@jwoudenberg I think I like this better, because it ensures that markdown is always rendered correctly in buttons.

src/Nri/Ui/Button/V2.elm
brookeangel: 👍

PR #74, opened 2018-06-15
Ink  update checkbox

tesk9:
Updating the checkbox module to use Html.Styled.

I also 💀 trimmed out the themes that it didn't seem like we are using anymore. E.g., `Square Orange` and `Locked`. I think there's a lot of room for further cleaning up the API, but I think that might require a broader discussion.

Why did I do this?

For https://www.pivotaltracker.com/story/show/157794412, I was asked to opacify the checkbox but not the text. Going to the checkbox module, I found that we're mostly still using the deprecated monolith checkbox (😱). To use the noredink-ui checkbox and change the styles, I would have needed to add the noredink-ui checkbox styles to the compiled site stylesheet dealio. But those styles are sloooowwww and I didn't have an approach in mind for partially opacifying the checkmark background images anyway, so I thought nahhhh let's do the new thing that we want to have anyway!

As it turned out though, this was a pretty challenging module to upgrade to the new Html.Styled pattern! there was a lot of forking and a lotttt of nested selectors... logic in the stylesheets and such.

I did what I could!

But did I do it correctly??

Hard-to-impossible to say!

The checkbox images aren't in the Nri.Icon module, and we also don't have a styleguide example yet. Kind reviewer, do you have a suggestion for what should happen next here? I'm very aware that this PR has become MASSIVE and I don't want to make it worse. What would you like? how should this move forward?


src/Nri/Ui/Checkbox/V3.elm
```diff
@@ -529,13 +508,6 @@ type ColorTheme
     | Orange
 
 
-{-| -}
-styles : Nri.Ui.Styles.V1.StylesWithAssets Never CssClasses msg (Assets r)
-styles =
-    (\assets -> [] |> List.concat)
-        |> Nri.Ui.Styles.V1.stylesWithAssets "checkbox-"
```
jwoudenberg: 💀 😻 🎉 💯

src/Nri/Ui/Radio/V1.elm
```diff
@@ -0,0 +1,228 @@
+module Nri.Ui.Radio.V3
```
jwoudenberg: YES! 😻 

Should this be a `V1`, considering its the first version of this module? 

**EDIT**: I see this change got undone in a next commit. Out of curiosity, was that out of concerns for this PR to grow to big or something else?

src/Nri/Ui/Checkbox/V3.elm
```diff
@@ -443,8 +454,8 @@ type ColorTheme
 -- ICONS used instead of default browser implementations
 
 
-backgroundImage : Assets r -> CheckboxImage -> Css.Style
-backgroundImage assets checkboxImage =
+backgroundImageDeprecated : Assets r -> CheckboxImage -> Css.Style
+backgroundImageDeprecated assets checkboxImage =
```
jwoudenberg: ❓ Why is this function deprecated?

src/Nri/Ui/PremiumCheckbox/V1.elm
```diff
@@ -0,0 +1,84 @@
+module Nri.Ui.PremiumCheckbox.V1 exposing (PremiumConfig, premium)
```
jwoudenberg: 😻

src/Nri/Ui/Radio/V1.elm
tesk9: I was wrong about this being meant to represent a radio button, actually! The "Round" style was used to represent checkmarks on the topic sentences outlines in the Rate step, and that seemed like a really specialized use that was overcomplicating the checkmark. 

Also, in the monolith there's already a RadioButton component -- seems like that should be added to noredink-ui separately?

tesk9:
Thank you, Jasper!!! 🙏

src/Nri/Ui/Radio/V1.elm
jwoudenberg: Ah, that makes sense, thanks for explaining!

Concerning these styles representing one-off implementations for various parts of the codebase, like the round style, happy to see them removed, it simplified things a lot! Do you have some ideas how widgets currently using these one-of themes in the monolith should be implemented?

tesk9:
Updated look!

screen shot 2018-06-20 at 1 18 01 pm

@NoRedInk/design double-checking understanding: whenever the checkbox is disabled, we want the icon to be opaque but the text to not be opaque, right?

bendansby:
Yeah, that's probably best, for accessibility's sake.

tesk9:
@jwoudenberg do you think it's better to update the look of the checkboxes in one fell swoop as part of this PR, or would it be okay to stage that update in another release...? I'm happy either way, just aware this PR is big ⛰

Issue #77, opened 2018-06-19
Organize icons in styleguide example app


tesk9:
The monolith has nicely organized icons that are labelled with the icon name:

<img width="925" alt="screen shot 2018-06-19 at 4 40 13 pm" src="https://user-images.githubusercontent.com/8811312/41629636-759e6652-73df-11e8-99f5-762a90f085fa.png">

The styleguide example app should also (or instead?) be nice to work with.

PR #79, opened 2018-06-22
Ink  start buttons

tesk9:
# 🦊 
https://github.com/NoRedInk/noredink-ui/issues/77 and https://www.pivotaltracker.com/story/show/158473000

screen shot 2018-06-25 at 10 55 11 am

- matches the monolith sorting for Icons in the styleguide
- adds the start icons (I started out trying to add them according the svg sprite pattern, but it was tricky since the build script is in the monolith, and the icons didn't go through that clearly, but the only place I could see the icons rendered was over here 🤷‍♀️ seemed better to go one step at a time instead)


styleguide-app/Examples/Icon.elm
```diff
@@ -1,179 +1,201 @@
...
 import Nri.Ui.Colors.V1 exposing (..)
 import Nri.Ui.Icon.V2 as Icon
-import Nri.Ui.Styles.V1
 
 
 {-| -}
 example : ModuleExample msg
 example =
-    { filename = "Nri/Ui/Icon/V1.elm"
+    { filename = "ui/src/Nri/Icon.elm"
```
stoeffel: I think it would actually be nice to have some version info in the style guide. I wonder if we should create a new styleguide page whenever we bump the version. I think that's not part of this pr, but maybe worth creating an issue/discourse.

styleguide-app/Examples/Icon.elm
```diff
@@ -1,179 +1,201 @@
...
-        , Html.div [ styles.class [ Container ] ]
-            (List.map viewIcon imageIcons)
-        , Headings.h4 [ Html.text "Scalable Vector Graphics (SVGs)" ]
-        , Html.div [ styles.class [ Container ] ]
-            (List.map viewIcon svgIcons)
-        , Headings.h5 [ Html.text "Assignment Types" ]
-        , Html.div [ styles.class [ Container ] ]
-            (List.map viewIcon assignmentTypeSvgIcons)
+        [ Html.h3 [] [ Html.text "Icon" ]
+        , viewIconSection "Mastery Icons"
```
stoeffel: 😻 I like the sections! great idea!

src/Nri/Ui/Icon/V3.elm
```diff
@@ -0,0 +1,837 @@
+module Nri.Ui.Icon.V3
```
stoeffel: I assume there wasn't any change here yet, right?

styleguide-app/Examples/Icon.elm
```diff
@@ -195,7 +195,6 @@ viewIcon { alt, background, icon } =
             , title alt
             ]
             [ Icon.icon { alt = alt, icon = icon assets }
```
stoeffel: oh, there was a change her. I'll go back...


Meta:
There was a guide on how to/when to create a new version right? I think we should include a guideline in that to create a commit with an exact copy of the current version and make the changes on top of that commit. This would make reviews a lot simpler.

src/Nri/Ui/Icon/V3.elm
```diff
@@ -412,15 +412,15 @@ arrowRight assets =
 
 
 {-| -}
-assignmentStartButtonPrimary : { r | assignmentStartButtonPrimary_svg : String } -> IconType
+assignmentStartButtonPrimary : { r | assignmentStartButtonPrimary_svg : Asset } -> IconType
```
stoeffel: what was the problem? I think to remember that I had problems with icons that weren't clean. I had success by using https://jakearchibald.github.io/svgomg/. Maybe create an issue for that problem.

Edit: actually I don't know which script you ran. I had problems with the one in the monolith not here.

src/Nri/Ui/Icon/V3.elm
stoeffel: vimdiff'ed locally. Changes look good!

PR #81, opened 2018-06-26
Ink  checkbox opacity

tesk9:
Another small change for https://www.pivotaltracker.com/story/show/157794412.

We want to opacify the checkbox images but not the text. See https://github.com/NoRedInk/noredink-ui/pull/74#issuecomment-398883130 and https://github.com/NoRedInk/NoRedInk/pull/20360 for some context.

screen shot 2018-06-25 at 5 48 47 pm

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Checkbox/V3.elm
+++ b/src/Nri/Ui/Checkbox/V3.elm
@@ -61,7 +61,7 @@ type IsSelected
 {-| -}
 type Theme
     = Square
-    | LockOnInside
+    | Locked
 
 
 selectedToMaybe : IsSelected -> Maybe Bool
@@ -119,7 +119,7 @@ buildCheckbox assets modifierClasses model labelContent =
                 , labelContent = labelContent
                 }
 
-            LockOnInside ->
+            Locked ->
                 { containerClasses = toClassList (modifierClasses ++ [ "LockOnInsideClass" ])
                 , labelStyles = lockLabelStyles model assets.checkboxLockOnInside_svg
                 , labelClasses = labelClass model.selected
--- a/src/Nri/Ui/Checkbox/V3.elm
+++ b/src/Nri/Ui/Checkbox/V3.elm
@@ -168,7 +168,6 @@ positioning =
     batch
         [ display inlineBlock
         , padding4 (px 13) zero (px 13) (px 35)
-        , verticalAlign middle
         ]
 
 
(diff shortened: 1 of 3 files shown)


bendansby:
I was thinking about this a bit more, and having the text show some indication of disabledness would be good I think. What do you think about keeping the text opaque, but making it Gray45, which would still be accessible?

tesk9:
No strong opinion. At this point, it's pretty easy to change, so I'm very happy to play and get it right now if we can. 😸 

screen shot 2018-06-25 at 6 00 09 pm
It looks a bit more busy to me with the different fonts -- a bit harder for me to parse visually. But it's hard for me to picture what it would look like on a real page without putting it there.

Do it? 👍 👎 ? ... wait for an opinion from Stacey...? I'm happy to do anything here 😄

bendansby:
I say do it. @staceyadams ?

staceyadams:
Did the lock icons and the "~" in the indeterminate state used to be full opacity?

I think the "~" should definitely be opaque.
I have mixed feelings on the premium lock.

👍 on Gray45 *only* when disabled. So in the first screenshot, Identify Adjectives would not be gray because it's selected. In the second screenshot, since it's not selected (and presumably can't be selected), it'd be gray. (But also that state probably wouldn't exist because the only reason something wouldn't be selectable right now is because it's behind a Premium or Premium w/ Writing lock?)

tesk9:
> Did the lock icons and the "~" in the indeterminate state used to be full opacity?

Yup! They are still in the [styleguide](https://staging.noredink.com/style_guide#category/Inputs). 
Should the premium lock not show as opaque? It's an easy change.

>  Identify Adjectives would not be gray because it's selected

But don't we want this for the Likert items on the form? Disabled (so grayed out) but also showing as selected?

--

I kinda wonder if we should talk synchronously?

staceyadams:
> I kinda wonder if we should talk synchronously?

Yes!

tesk9:
checkbox

staceyadams:
**tl;dr:** no color changes. Color styles will be split off from the checkbox styles so whoever is working on them can change them as needed.

**Disabled**
The only disabled state we currently have is where an item is selected and cannot be unselected (classes/students when editing an assigment, likert items for Peer Review on the form). I think in this case, the text should look like the other items in the list because it's something the user has access to.

In the reverse, where an item cannot be selected and the user doesn't have the feature, that's where I think Gray45 would make sense.

**Premium and Premium w/ Writing**
These aren't "disabled" because clicking on them pops a modal. But from the user point of view, they still can't be selected. This is where I think Gray45 could be applied but I don't want to make that call without looking at it in context of the from and going through the different user types. A small concern I have is that changing it may also be one of those subtle things that causes a teacher to think that we've locked more things than before.

This is separate from the original task of this PR and should be explored separately if you want to take a stab at it @bendansby. Also happy to discuss more in person or in #design on Slack.

terezka:
Hi! Somehow missed this so sorry I didn't get to this earlier! Looking at it now.

src/Nri/Ui/Checkbox/V3.elm
```diff
@@ -183,7 +189,7 @@ addIcon : Asset -> Style
 addIcon icon =
     batch
         [ position relative
-        , before
+        , checkboxImageSelector
```
terezka: 🙏

tesk9:
Ahhhhhhhhhhhhh kitttttyyyyyyyyyyyyyyyyyyyyyyyyyyy 😍 😍 😍 😻

Issue #83, opened 2018-06-27
Checkbox classnames should be versioned to prevent global style collisions


tesk9:
<img width="1302" alt="screen shot 2018-06-27 at 10 44 07 am" src="https://user-images.githubusercontent.com/8811312/41990397-236c4da4-79f7-11e8-89cf-c9a640e0c7d5.png">
<img width="355" alt="screen shot 2018-06-27 at 10 43 29 am" src="https://user-images.githubusercontent.com/8811312/41990398-23868282-79f7-11e8-98e7-7a67e141f6ac.png">

The monolith styles make the checkbox look pretty terrible as-is, as it turns out!

I expect we should follow some kind of class name scheme. Probably `checkbox-v3__Checked` or something like that.

BrianHicks:
**interior, day, a classroom somewhere in America.**

teacher: Bobby, is your writing convincing?

student: well…

David Caruso (crashing through window): did you… CHECK?

**smash cut to CSI: Miami opening credits**

Issue #82, opened 2018-06-27
Expose the Premium Checkbox module


tesk9:
I forgot to add the Premium Checkbox module to the API. 🤦‍♀️ 

See https://github.com/NoRedInk/noredink-ui/pull/74/files#diff-5e3b384244cc449ce000177d9d73fc17

PR #84, opened 2018-06-27
Ink  fix checkbox

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/83 and https://github.com/NoRedInk/noredink-ui/issues/82.

There should be no visual changes to the styleguide as a result of this change. Should just expose the Premium checkbox api and change internal class names to prevent namespace collisions in the monolith.


elm-package.json
```diff
@@ -12,11 +12,11 @@
         "Nri.Ui.BannerAlert.V1",
         "Nri.Ui.Button.V1",
         "Nri.Ui.Button.V2",
-        "Nri.Ui.Colors.V1",
```
joneshf: 😻 😻 😻 😻 😻

PR #90, opened 2018-07-24
textarea uses defaultValue again

brookeangel:
TextArea.V3 was causing flakes in the monolith because it used `text` rather than `defaultValue` to set the value. 

Context: https://noredink.slack.com/archives/C0VVDLEES/p1532116237000092

This sets it back to using `defaultValue`

To QA, please test that the textarea in the styleguide app still works properly, and can autoresize, and all that good stuff.


brookeangel:
👍 Yeah - I always just write the QA instructions here for engineers to test it out, maybe there needs to be a better term for that though 😛

PR #92, opened 2018-07-27
Revert "Merge pull request #90 from NoRedInk/textarea-blast-from-the-…

brookeangel:
We thought the guided_writes_spec is flaking because Textarea.V3 used text instead of defaultValue, leading to missed keystrokes in some Capybara runs, but tried this in https://github.com/NoRedInk/NoRedInk/pull/20790 and still observed the flake.

- Slack conversation: https://noredink.slack.com/archives/C0VVDLEES/p1532116237000092
- Bug report: elm/virtual-dom#107

The actual fix can be found here: https://github.com/NoRedInk/NoRedInk/pull/20822

Reverting this because changing Textarea.V3 to use `defaultValue` does not actually fix the flake, and will result in us needing to update elm specs in the monolith that look for text. 

## Future fixes
This bug seems to only happen in the custom element resizing text area.... can we make that more performant?


brookeangel:
cc: @Arkham

Arkham:
Let's merge this in!

PR #93, opened 2018-07-31
Cache the prefix and styles on `Nri.Ui.styled`

rtfeldman:
This applies the same performance optimization as https://github.com/rtfeldman/elm-css/pull/456

It doesn't change the API or the behavior of the function. All it does is change it so when you call `Nri.Ui.styled div "funky" stylesGoHere` it *immediately* calculates the `Attribute` values for `data-nri-description` and `css`, including the hashing calculation for the generated classname.

Then it returns a `List (Attribute msg) -> List (Html msg) -> Html msg` function just like before, except now that function will run faster because the "styled div" function won't have to recalculate those values every time it gets called; it will already have them stored from before.


rtfeldman:
> I don't claim to get exactly how this works

It's a pretty unusual trick, so you're probably not alone on that! 😄

I've added a H&B demo to explain how it works!

PR #100, opened 2018-08-16
Add deploy steps to README and PR template

jamesgary:
Clarify the deploy process. Suggestions welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,3 +49,35 @@ make test
 
 Travis will run `make ci` to verify everything looks good.
 You can run this locally to catch errors before you push!
+
+## Deploying
+
+Once you are ready to publish `master` as a new version, run the following to bump the version in `elm-packages.json`:
+
+```
+elm-package bump
+```
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,3 +49,35 @@ make test
...
+```
+elm-package publish # requires a tag, see below
+```
+
+This will require a tag, so follow the instructions it provides. It should be something like this:
+
+```
+Verifying NoRedInk/noredink-ui 5.10.0 ...
+Version number 5.10.0 verified (MINOR change, 5.9.1 => 5.10.0)
+Error: Libraries must be tagged in git, but tag 5.10.0 was not found.
```
quelledanielle: Minor comment: could put the git tag instructions before `elm-package publish` so people don't need to see this error, but I think this is fine too

jamesgary: I left it in just because the error message generates the commands with the correct versions.

jamesgary: Thanks for the review!

PR #101, opened 2018-08-17
Upgrade Nri Button from V2 -> V3

brookeangel:
This upgrades Nri.Button V2 to Nri.Button V3 so we can make use of new elm-css! 

This needs thorough @NoRedInk/design review. Make sure the buttons look like they do on the current styleguide in all states. 

To build this: 
- Clone this repo
- `> make styleguide-app/elm.js`
- `> open styleguide-app/index.html`

Post merge TODO: 

- [ ] Run `elm-package bump` to make sure you're bumping to the correct version

### DIFF
--- a/src/Nri/Ui/Button/V3.elm
+++ b/src/Nri/Ui/Button/V3.elm
@@ -1,0 +1,908 @@
+module Nri.Ui.Button.V3
+    exposing
+        ( ButtonConfig
+        , ButtonContent
+        , ButtonSize(..)
+        , ButtonState(..)
+        , ButtonStyle(..)
+        , LinkConfig
+        , ToggleButtonConfig
+        , button
+        , copyToClipboard
+        , customButton
+        , delete
+        , link
+        , linkExternal
+        , linkExternalWithTracking
+        , linkSpa
+        , linkWithMethod
+        , linkWithTracking
+        , styles
+        , submit
+        , toggleButton
+        )
+
+{-| Changes from V2:
+
+  - Uses Html.Styled
+  - Removes buttonDeprecated
+  - Removes Tiny size
+
+Common NoRedInk buttons. For accessibility purposes, buttons that perform an
+action on the current page should be HTML `<button>` elements and are created here
+with `*Button` functions. Buttons that take the user to a new page should be
+HTML `<a>` elements and are created here with `*Link` functions. Both versions
+should be able to use the same CSS class in all cases.
+
+There will generally be a `*Button` and `*Link` version of each button style.
+(These will be created as they are needed.)
+
+
+## Required styles
+
+@docs styles
+
+
+## Common configs
+
+@docs ButtonSize, ButtonStyle, ButtonState, ButtonContent
+
+
+## `<button>` Buttons
+
+@docs ButtonConfig, button, customButton, delete, copyToClipboard, ToggleButtonConfig, toggleButton
+
+
+## `<a>` Buttons
+
+@docs LinkConfig, link, linkSpa, linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
+
+
+## `<input>` Buttons
+
+@docs submit
+
+-}
+
+import Accessibility.Styled exposing (..)
+import Accessibility.Styled.Role as Role
+import Accessibility.Styled.Widget as Widget
+import Css exposing (..)
+import EventExtras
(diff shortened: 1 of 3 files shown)


src/Nri/Ui/Button/V3.elm
```diff
@@ -0,0 +1,843 @@
...
+        , linkSpa
+        , linkWithMethod
+        , linkWithTracking
+        , toggleButton
+        )
+
+{-|
+
+
+# Changes from V2:
```
brookeangel: Note: other than the changes here, this file is more or less a copy of V2
The biggest change is updating all the CSS

src/Nri/Ui/Button/V3.elm
```diff
@@ -0,0 +1,843 @@
...
+                      , shadow = rgba 0 0 0 0
+                      }
+                    , [ Css.hover
+                            [ textDecoration underline
+                            , Css.disabled
+                                [ textDecoration none
+                                ]
+                            ]
+                      ]
+                      -- TODO: Weird case with borderless
```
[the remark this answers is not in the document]
brookeangel: We don't.

brookeangel:
> The only discrepancy I noticed is that the large size seems to have lost its min width

Huh, that's weird. Is this not correct ?

screen shot 2018-08-21 at 11 35 39 am

brookeangel:
> Also, links and buttons have slightly different vertical alignment of their labels, which is present in the current button too, but no reason they shouldn't be the same.

Ok if we get to this one in the next minor patch? Links and buttons currently behave differently - apparently you can't add a multi-line link, and line-height is set differently between the two of them. Maybe best if we just move them both to flexbox or something

bendansby:
>Huh, that's weird. Is this not correct ?

That's correct (didn't look that way for me locally)

bendansby:
>Ok if we get to this one in the next minor patch?

👍

brookeangel:
> That's correct (didn't look that way for me locally)

Did you uncheck the "width" option? (The width option hardcodes them to 120)

bendansby:
Maybe 🤷‍♂️  I trust your screenshot!

src/Nri/Ui/Button/V3.elm
```diff
@@ -0,0 +1,822 @@
...
+            , Widget.pressed <| Just False
+            , Role.button
+            , type_ "button"
+            ]
+        )
+        (viewLabel Nothing config.label)
+
+
+
+-- SUBMIT BUTTON
```
brookeangel: ❌ This section can be removed

src/Nri/Ui/Button/V3.elm
```diff
@@ -0,0 +1,822 @@
...
+
+
+{-| Describes the state of a button. Has consequences for appearance and disabled attribute.
+
+  - Enabled: An enabled button. Takes the appearance of ButtonStyle
+  - Unfulfilled: A button which appears with the InactiveColors palette but is not disabled.
+  - Disabled: A button which appears with the InactiveColors palette and is disabled.
+  - Error: A button which appears with the ErrorColors palette and is disabled.
+  - Loading: A button which appears with the LoadingColors palette and is disabled
+  - Success: A button which appears with the InactiveColors palette and is disabled
```
brookeangel: ❌ the comment for `Success` is wrong - it should say `SuccessColors` palette

src/Nri/Ui/Button/V3.elm
```diff
@@ -827,3 +830,8 @@ sizeStyle size width elementType =
             ]
         ]
     ]
+
+
+styledName : String -> String
```
brookeangel: thanks!

src/Nri/Ui/Button/V3.elm
```diff
@@ -216,21 +216,12 @@ customButton attributes config content =
...
          , Attributes.disabled disabled
          , Attributes.type_ "button"
-         , widthStyle config.width
          ]
             ++ attributes
         )
         (viewLabel content.icon content.label)
 
 
-widthStyle : Maybe Int -> Attribute msg
```
brookeangel: 🎉 Thanks, missed this

michaelglass:
will take a look. Thanks, Brooke!

todo: add backstopjs!

brookeangel:
Looks great! Let's merge!

PR #103, opened 2018-08-21
Update readme with phase out info

brookeangel:
This PR updates the Readme to give information about how we bring in widgets from the monolith, and phase out old versions of widgets


jwoudenberg:
This is awesome! 😻

PR #110, opened 2018-08-29
Upgrade Modal V1 -> Modal V2

brookeangel:
This upgrades Modal V1 -> Modal V2 so that we have access to modals with new elm css!!!

Weirdly, Nri.Ui.Modal.V1 is not used in the monolith yet.

It also uses the current styles in the monolith for V2, because Nri.Modal from the monolith also has newer styles than V1. 🤷‍♀️ 

## QA 
Build the styleguide-app and look at the modal example.

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -29,6 +29,7 @@
         "Nri.Ui.Icon.V3",
         "Nri.Ui.InputStyles",
         "Nri.Ui.Modal.V1",
+        "Nri.Ui.Modal.V2",
         "Nri.Ui.Outline.V1",
         "Nri.Ui.Page.V1",
         "Nri.Ui.Palette.V1",
--- a/src/Nri/Ui/Modal/V2.elm
+++ b/src/Nri/Ui/Modal/V2.elm
@@ -1,0 +1,213 @@
+module Nri.Ui.Modal.V2
+    exposing
+        ( Model
+        , info
+        , warning
+        )
+
+{-| Changes from V1:
+
+  - Use Styled Html
+
+@docs Model
+@docs info
+@docs warning
+
+-}
+
+import Accessibility.Styled as Html exposing (..)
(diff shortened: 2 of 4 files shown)


brookeangel:
Thanks @mavnn !!! Oh one final thing, I should probably ping @bendansby to take a looker at this one since we've historically had lots of modal style drama in the past. 

Note for Dansby: unfortunately this PR isn't meant to improve the modal styles to add back in shading - it should just keep the current status quo while transferring to new elm-css which should at least make it easier to change styles in the future.

bendansby:
Scrolled content seems to have regressed

napkin 16 08-31-18 8 38 33 am

brookeangel:
Hey @bendansby - how did you trigger that state? Is it when you make the window very short?

bendansby:
Yes. Prod handles it better

brookeangel:
screen shot 2018-08-31 at 12 13 18 pm

I think it may have been due to spare margins floating around - this is what it looks like now! 

Also... out of scope of this PR, but "style-guide app" here is deceiving because the layout doesn't set many defaults and it is not box-sizing border box. Should probably attempt to make styles here more similar to styles in the monolith to get a more accurate representation of what things will look like there. This is what they look like with the margin removed, and me sneakily adding box-sizing border box to the page.

bendansby:
🙏

Yes, it would be great if the style guide reflected our styles! 😅

brookeangel:
screen shot 2018-08-31 at 12 16 38 pm

heh one more shot at uploading a screenshot, fixed the footer margin too

brookeangel:
Ok... let me know if that looks ok... hopefully all the styles are identical now... some weird stylistic drift had occured between noredink-ui and the monolith 😛

bendansby:
Looks good!

Issue #85, opened 2018-07-05
Is it possible to use PremiumCheckbox?


SamGerber:
It seems like Nri.Ui.PremiumCheckbox.V1.PremiumConfig requires providing Nri.Ui.Data.PremiumLevel s?

https://github.com/NoRedInk/noredink-ui/blob/0a028997d13ce4652124845b25a5b794e4ec53ce/src/Nri/Ui/PremiumCheckbox/V1.elm#L30-L31

But the elm-package doesn't appear to expose PremiumLevel.
https://github.com/NoRedInk/noredink-ui/blob/0a028997d13ce4652124845b25a5b794e4ec53ce/elm-package.json#L9-L53

Is this a possible explanation for why I'm struggling to use PremiumCheckbox, or am I just doing things incorrectly? I appreciate any help on how to use this! Thanks <3

tesk9:
Oops! Sorry about that, Sam. Exposing should make it possible to use the premium checkbox, but might be kinda complicated on the monolith side depending on how many references to the Nri.Data.PremiumLevel there are/how intertwined they are.

I wonder if there's a way to decouple the Nri.Ui.Data.PremiumLevel from the Premium checkbox -- it seems kind of strange for a view component to depend on a specific NRI data type...? Maybe not.

tesk9:
Did anything end up changing here? Are premium checkboxes usable, or constrained to the monolith?

SamGerber:
I used the deprecated ones in the monolith for what I was working on.

jamesgary:
Just to follow up: https://github.com/NoRedInk/noredink-ui/pull/111 has been merged and published, so you ought to be able to use `Nri.Ui.PremiumCheckbox.V2` without `PremiumLevel`s.

PR #111, opened 2018-08-30
Don't require PremiumLevel; allow general premium levels

jamesgary:
See https://github.com/NoRedInk/noredink-ui/issues/85

This makes the `PremiumConfig` more general and not rely on a specific `PremiumLevel`. It also now requires `isFree : level -> Bool` and `allowedFor : level -> level -> Bool`.

The upside is that it's no longer coupled to any specific `PremiumLevel` and can we can change/add the levels in the monolith. The downside is that all premium checkboxes need the new `isFree` and `allowedFor` configs, but those should probably be defined functions in `Data.PremiumLevel` so there's not too much duplication of code.

Feedback welcome! /cc @tesk9 @SamGerber 

Side note: Not sure if this will need to get bumped from `Nri.Ui.PremiumCheckbox.V1` to `Nri.Ui.PremiumCheckbox.V2` since it's not getting used in the monolith. If it does need to get bumped, I'll keep `V1` as-is and move changes to `V2`.

### DIFF
--- a/src/Nri/Ui/PremiumCheckbox/V1.elm
+++ b/src/Nri/Ui/PremiumCheckbox/V1.elm
@@ -2,7 +2,7 @@ module Nri.Ui.PremiumCheckbox.V1 exposing (Pennant(..), PremiumConfig, premium)
 
 {-|
 
-@docs PremiumConfig, premium
+@docs PremiumConfig, premium, Pennant
 
 -}
 
@@ -34,6 +34,9 @@ type alias PremiumConfig msg =
     }
 
 
+{-| Premium is the yellow "P" pennant
+PremiumWithWriting is the yellow "P+" pennant
+-}
 type Pennant
     = Premium
     | PremiumWithWriting
--- a/src/Nri/Ui/PremiumCheckbox/V1.elm
+++ b/src/Nri/Ui/PremiumCheckbox/V1.elm
@@ -1,4 +1,4 @@
-module Nri.Ui.PremiumCheckbox.V1 exposing (PremiumConfig, premium)
+module Nri.Ui.PremiumCheckbox.V1 exposing (Pennant(..), PremiumConfig, premium)
 
 {-|
 
@@ -27,12 +27,16 @@ type alias PremiumConfig msg =
     , selected : Checkbox.IsSelected
     , disabled : Bool
     , isLocked : Bool
-    , isFree : Bool
-    , showFlagWhenLocked : Bool
+    , pennant : Maybe Pennant
     , onChange : Bool -> msg
     , onLockedClick : msg
     , noOpMsg : msg
     }
+
+
+type Pennant
+    = Premium
+    | PremiumWithWriting
 
 
 {-| A checkbox that should be used for premium content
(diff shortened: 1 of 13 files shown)


src/Nri/Ui/PremiumCheckbox/V1.elm
```diff
@@ -22,13 +21,15 @@ import Nri.Ui.Data.PremiumLevel as PremiumLevel exposing (PremiumLevel(..))
...
     { label : String
     , id : String
     , selected : Checkbox.IsSelected
     , disabled : Bool
-    , teacherPremiumLevel : PremiumLevel
-    , contentPremiumLevel : PremiumLevel
+    , teacherPremiumLevel : level
+    , contentPremiumLevel : level
+    , isFree : level -> Bool
+    , allowedFor : level -> level -> Bool
```
avh4: 💭 This is a case where I think it's very risky to not label the parameters, since flipping them will case runtime bugs.  Maybe something like `allowedFor: { userLevel : level, checkboxContentLevel : level } -> Bool` ?

src/Nri/Ui/PremiumCheckbox/V1.elm
```diff
@@ -22,13 +21,15 @@ import Nri.Ui.Data.PremiumLevel as PremiumLevel exposing (PremiumLevel(..))
...
-type alias PremiumConfig msg =
+type alias PremiumConfig level msg =
     { label : String
     , id : String
     , selected : Checkbox.IsSelected
     , disabled : Bool
-    , teacherPremiumLevel : PremiumLevel
-    , contentPremiumLevel : PremiumLevel
+    , teacherPremiumLevel : level
+    , contentPremiumLevel : level
```
avh4: 💭 hmm,  if we're passing the levels just so the checkbox can pass them to functions that we are also passing, then why not just pass the bools `isFree : Bool, isAllowed : Bool` instead of passing the levels and the functions?

jamesgary:
Simplified the config to just use two booleans: `isFree` and `isLocked`. I also noticed that these should never both be true; why would something free be locked? So I introduced a new type in a [different commit](https://github.com/NoRedInk/noredink-ui/commit/8d0ef9b26f9004c45a28f6e572c294023160a30a) outside of this PR: 

```
type Status
    = Free
    | PaidAndLocked
    | PaidAndUnlocked
```

But this didn't seem to improve things much. It made the module and client code more verbose, and offloaded the work to determine the `Status` to the client, which doesn't make it more convenient. I think I'm fine sticking with the two booleans approach.

jamesgary:
Ready for another review! 

- Added the ability to show a `P+` pennant
- Simplified the `PremiumConfig` a bit (no more `showFlagWhenLocked: Bool` since I didn't see that logic in prod anywhere, so I just have `pennant: Maybe Pennant`.
- Simplified example. The original controls didn't align with how the API is getting used, so I just made a handful of examples instead of dynamic controls. I can add them back in if we still think they're useful though.
- Added `.iconPremiumWithWritingFlag_svg` to `Assets`, which is the same name that's in prod

Side note (from above): Not sure if this will need to get bumped from Nri.Ui.PremiumCheckbox.V1 to Nri.Ui.PremiumCheckbox.V2 since it's not getting used in the monolith. If it does need to get bumped, I'll keep V1 as-is and move changes to V2.

src/Nri/Ui/PremiumCheckbox/V1.elm
```diff
@@ -27,29 +26,29 @@ type alias PremiumConfig msg =
...
+    , pennant : Maybe Pennant
     , onChange : Bool -> msg
     , onLockedClick : msg
     , noOpMsg : msg
     }
 
 
+{-| Premium is the yellow "P" pennant
+PremiumWithWriting is the yellow "P+" pennant
+-}
```
SamGerber: 😻

src/Nri/Ui/PremiumCheckbox/V1.elm
```diff
@@ -27,29 +26,29 @@ type alias PremiumConfig msg =
...
+type Pennant
+    = Premium
+    | PremiumWithWriting
+
+
 {-| A checkbox that should be used for premium content
 
 This checkbox is locked when the premium level of the content is greater than the premium level of the teacher
```
SamGerber: I can't decide if this comment is helpful here any longer.

PR #112, opened 2018-09-06
Add back close button to Modal.V2

brookeangel:
😳 Well, this is embarrassing. It turns out that neither modal V1 or V2 had a close button!

It's too late for V1 (which fortunately was never used in the monolith), but not too late for V2! As of this PR, V2 should have a close button again!

## QA 
- Build the styleguide app and check out the modal. Make sure the X appears

screen shot 2018-09-06 at 8 15 47 am

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -30,6 +30,7 @@
         "Nri.Ui.InputStyles",
         "Nri.Ui.Modal.V1",
         "Nri.Ui.Modal.V2",
+        "Nri.Ui.Modal.V3",
         "Nri.Ui.Outline.V1",
         "Nri.Ui.Page.V1",
         "Nri.Ui.Palette.V1",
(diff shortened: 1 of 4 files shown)


brookeangel:
CI is failing because major changes are not allowed - Modal V2 is not yet used in the monolith, so maybe it's ok?

BrianHicks:
I'm not sure what to do about the major version. It feels like we had better not change it, as we have no guarantees that someone isn't writing code against it at this very moment. If you feel strongly the other way, maybe we could ask in #eng-frontend or #eng-front-ops-changes? This does seem like a corner case of our current policy.

brookeangel:
Yeah, as far as I know, the only person writing code against it was me, and I discovered the bug because my integration tests in the monolith were failing! I'll post in eng-front-ops

brookeangel:
Oh, dismissable by clicking the Primary Button, or dismissable through other means?

If by clicking the primary button, that's by design (simulating the user having to do some action to make it go away)
If by other means... 🐛 !

brookeangel:
For context, asking the question here about version upgrades: https://noredink.slack.com/archives/C9A72T555/p1536264722000100

BrianHicks:
I think I misunderstood the intent of the non-dismissable modal. Is it that there is no X in the upper-right-hand corner? If that's the intent, it's working properly here after all. 😄

brookeangel:
🍐  @Arkham  - merging!

PR #113, opened 2018-09-07
Refactor and add thead/tbody to Nri.Table

terezka:
This PR adds accessibility tags to our table (`thead` / `tbody`) and moves a bit of code around. It should not result in any visual or functional changes.


src/Nri/Ui/Table/V2.elm
```diff
@@ -173,7 +176,10 @@ table styles =
 
 tableHeader : List (Column data msg) -> Html msg
 tableHeader columns =
-    tr [ css headersStyles ] (List.map tableRowHeader columns)
+    thead []
```
brookeangel: 👍

PR #115, opened 2018-09-10
SegmentedControl: "fit content" and "fill container" settings

juanedi:
This PR adds a new version of `SegmentedControl`, which uses auto-generated css classes and adds a `width` field to the the `Config`, which can control whether the element will expand to fill its container or fit its contents.

**This will probably be easier to review commit by commit**

For https://www.pivotaltracker.com/story/show/157378996

Some context:

Until V4 the control would always fit its contents. On V5 that behaviour was changed to always fill the container. This means that to limit the element's width one had to limit the value of its container (which might be hard to do when the labels of the tabs are unknown or might change).

We now support both settings, and the user has to decide what to do for each use case.

On the monolith, `Nri.SegmentedControl` is stuck with V4 for this reason. The plan is to add this new setting it this API's config too and make all pre-existing use cases use `FitContent`.

## 🎬 Preview

cc: @bendansby


brookeangel:
> Until V4 the control would always fit its contents. On V5 that behaviour was changed to always fill the container. This means that to limit the element's width one had to limit the value of its container (which might be hard to do when the labels of the tabs are unknown or might change).

😆 ahahhah goodness... SegmentedControl definitely wins award for highest version number

src/Nri/Ui/SegmentedControl/V6.elm
```diff
@@ -0,0 +1,144 @@
...
+import Accessibility.Styled exposing (..)
+import Accessibility.Styled.Role as Role
+import Css exposing (..)
+import Css.Foreign exposing (Snippet, adjacentSiblings, children, class, descendants, each, everything, media, selector, withClass)
+import Html.Styled as Html exposing (Html)
+import Html.Styled.Attributes as Attr exposing (css)
+import Html.Styled.Events as Events
+import Nri.Ui.Colors.Extra exposing (withAlpha)
+import Nri.Ui.Colors.V1 as Colors
+import Nri.Ui.CssFlexBoxWithVendorPrefix as FlexBox
```
brookeangel: ❓ Do we still need to vendor prefix? I never know

bendansby: No, not for flexbox

bendansby:
Works for me!

src/Nri/Ui/SegmentedControl/V6.elm
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+
+{-| -}
+view : Config a msg -> Html.Html msg
+view config =
+    config.options
+        |> List.map
+            (\option ->
+                Html.div
```
brookeangel: 🎨 Could we use an `Nri.Ui.styled div` here that points to "segmented-control-v6-tab" or something? Looks like we are pretty well covered with accessibility attributes, but I still find the `data` attrs helpful

styleguide-app/NriModules.elm
```diff
@@ -235,11 +235,10 @@ route location =
 styles : List Stylesheet
 styles =
     List.concat
-        [ -- NOTE: these will go away as the modules' styles are integrated with Nri.Css.Site.elm
+        [ -- NOTE: these will go away as the modules' styles are migrated to use autogenerated classes based on Html.Styled.Attributes.css
           [ ModuleExample.styles
           ]
         , (Examples.Page.styles |> .css) ()
-        , (Examples.SegmentedControl.styles |> .css) ()
```
brookeangel: 🎉

brookeangel:
screen shot 2018-09-11 at 9 36 35 am

Is this the right behavior for one long and one short segment @bendansby ? (I don't actually know)

bendansby:
>Is this the right behavior for one long and one short segment 

In all my designs, I have made each segment equally wide (or the equivalent of the new fill container argument). I'm happy to leave in the possibility of them having different widths in unusual cases like the above, though I don't imagine I'll use it very often.

juanedi:
> I'm happy to leave in the possibility of them having different widths in unusual cases like the above, though I don't imagine I'll use it very often.

I'm not sure I follow. Unless I'm missing some ad-hoc fixes, the current behavior in the monolith is to always fit the content, so we probably do have places where the tabs' sizes differ 😅 

Is that a bug? Should I add it to the style guide TODO list?

FWIW, I'm no CSS expert but making all tabs have the size of the one with the widest content doesn't seem to be an easy thing to do.

bendansby:
>so we probably do have places where the tabs' sizes differ

Oh, I know. Just that my *designs* aren't like that 😄 

>Is that a bug? Should I add it to the style guide TODO list?

I don't think we need to worry about it right now.

> making all tabs have the size of the one with the widest content doesn't seem to be an easy thing to do.

"Fill content" effectively does this, no?

juanedi:
> The only thing that needs changing is: don't forget to add V6 to the exposed modules in elm-package.json!

Hah! Thanks for catching that. Mind taking another look now @brookeangel? :-)

PR #119, opened 2018-09-21
Add select v4

tremlab:
using custom select arrows in V4.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,41 +1,43 @@
 {
   "name": "@noredink/ui",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "JSONStream": {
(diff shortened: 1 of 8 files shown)


tremlab:
image
@bendansby How does this look?  We have some questions:

- do we need error styles?
- do we need label option?

thanks!

bendansby:
Ideal would be to have the states and label option as seen in the style guide, but I leave it up to y'all as to whether to include them in this PR.

Style note: we don't want the inner shadow. But that made realize we do want a hard drop shadow to match our medium buttons! I've updated the Zeplin if you'd like to add that in this PR. If not, just removing the inner shadow would be fine.

Thanks for working on this!!

michaelglass:
updated elm-format. Can you use the new elm-format here before you merge? <3

stoeffel:
@tremlab is this still in use or can we close this?

PR #121, opened 2018-09-27
High-rise buttons

allenap:
This allows V2 and V3 buttons to grow vertically. Commit messages have more explanation.

I grafted this into the monolith and it works as expected:

image

I have visually inspected both buttons in the style guide (V3 in the noredink-ui style guide, V2 in the monolith style guide) and checked CSS attributes via Chrome's inspector.

> Hello there, wonderful author of widgets!
> This is `Nri.Ui.Friendly.AI.V1` speaking to you!
> Are you looking to get this awesome work reviewed as quickly as possible, so you can start putting it to use?
> Then feel free to grab a reviewer from the list below and assign them to the PR!
>
> https://paper.dropbox.com/doc/noredink-ui-widget-library-fUK6yq32g187lxw2A60a8#:uid=619243092748985044355046&h2=Reviewers-list
>
> It's best if we have separate pull requests for code changes and package version bumps. If you're just changing code, great! If you're just bumping the version, check out https://github.com/NoRedInk/noredink-ui#deploying.

PR #123, opened 2018-10-01
Add V4 button with unbounded height behaviour

allenap:
Well, #121 (which I released as 5.18.1 last week) [wasn't popular in the monolith](https://noredink.slack.com/archives/C0VVDLEES/p1538157412000100). This PR instead adds a V4 button. This is very similar to the V3 except that it changes behaviour to allow the button to grow vertically as big as the content. I've put it in the style guide app, so give that a go (`make styleguide-app/elm.js && open styleguide-app/index.html`).

I've also defined `width` using a custom type, with members `WidthExact Int` and `WidthUnbounded`, corresponding exactly to the behaviour of the `Just Int` and `Nothing` of its previous definition.

**Note** that height only applies to `button` and `copyToClipboard` (both of which are ultimately `<button>` elements; `link` and the others are `<a>` elements).

See a  [diff of changes from V3 to V4](https://gist.github.com/leapingfrogs/2800c73f30b2291ed0af4038fc508b83).

> Then feel free to grab a reviewer from the list below and assign them to the PR!
>
> https://paper.dropbox.com/doc/noredink-ui-widget-library-fUK6yq32g187lxw2A60a8#:uid=619243092748985044355046&h2=Reviewers-list
>

### DIFF
--- a/src/Nri/Ui/Button/V4.elm
+++ b/src/Nri/Ui/Button/V4.elm
@@ -1,5 +1,5 @@
 module Nri.Ui.Button.V4 exposing
-    ( ButtonSize(..), ButtonHeight(..), ButtonWidth(..), ButtonStyle(..), ButtonState(..), ButtonContent
+    ( ButtonSize(..), ButtonWidth(..), ButtonStyle(..), ButtonState(..), ButtonContent
     , ButtonConfig, button, customButton, delete, copyToClipboard, ToggleButtonConfig, toggleButton
     , LinkConfig, link, linkSpa, linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
     )
@@ -9,7 +9,6 @@ module Nri.Ui.Button.V4 exposing
 
 # Changes from V3:
 
-  - Adds `ButtonHeight`.
   - Adds `ButtonWidth`.
 
 
@@ -33,7 +32,7 @@ may be exceptions, for example if button content is supplied by an end-user.
 
 ## Common configs
 
-@docs ButtonSize, ButtonHeight, ButtonWidth, ButtonStyle, ButtonState, ButtonContent
+@docs ButtonSize, ButtonWidth, ButtonStyle, ButtonState, ButtonContent
 
 
 ## `<button>` Buttons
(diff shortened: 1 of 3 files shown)


allenap:
Status information for completed builds doesn't seem to be making it back from Travis to GitHub correctly. Click through and you'll see a 🍏 📗 💚 🥗 build!

allenap:
> Do we really need the sizing option? Shouldn't all V4 buttons just grow vertically? (My concern with the previous PR was that it retroactively changed V2 and V3, not that it prevented fixed-height buttons.)

I assume that the goal is, in time, to migrate all usage of V2/3 to V4, and that'll be blocked if V4 can't support the non-expanding use cases:

1. User-supplied button text shouldn't be allowed to cause the button to grow indefinitely.

2. This:
   image

I expect (2) could be fixed with some jiggling of CSS, but the first is a bit more of a problem. We could truncate user-supplied text before rendering it into the `<button>`, but that's a bigger migration exercise.

Even so, I did think overnight that maybe this is better served with a `height` config option instead of the `sizing` option I added, which might be something like `HeightFixed | HeightBounded Float | HeightUnbounded` where the `Float` would be a multiplier of the basic button size rather than a pixel value.

allenap:
Here is a [diff from `Button.V3` to `Button.V4`](https://gist.github.com/allenap/ca62f1e13d7374b0f1c4b4a7064bfb22).

allenap:
Spoke with @avh4. I'm going to:

- Remove the `height` field. The configuration seems to be getting overly complex. We do also want consistency on the site, and too many knobs to tweak is moving us away from that goal.
- The corollary is that all V4 buttons will be able to expand vertically.
- There will be a maximum height, but it will be large, only to act as a backstop. Maybe something like the equivalent of 5 lines of text.
- When we come to migrate from V2 or V3 and encounter layout issues we will have to address them from outside, e.g. tweak the container. If worse comes to worst there's always the possibility of a V5.
- I will ask @brookeangel again tomorrow for another review.

src/Nri/Ui/Button/V4.elm
```diff
@@ -0,0 +1,867 @@
...
+    , ButtonConfig, button, customButton, delete, copyToClipboard, ToggleButtonConfig, toggleButton
+    , LinkConfig, link, linkSpa, linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
+    )
+
+{-|
+
+
+# Changes from V3:
+
+  - Adds `ButtonHeight`.
```
brookeangel: This part of the doc can go away, I think!

brookeangel:
Hello, Gavin! The button is looking lovely! 

One question or maybe bug: Is this the correct behavior for a link with a bounded width?
screen shot 2018-10-03 at 7 47 56 am

allenap:
@brookeangel Woah, that's not right. Erm, I tried it here and got:

image

Can you check you've got the latest code? Thanks!

allenap:
I've also tried this in the monolith (using [these instructions](https://paper.dropbox.com/doc/Grafting-noredink-ui-into-the-monolith--AOaaSVg~wsl4AJpUnqSx_dZdAg-ff7N9zwMyRR2D5Umza5Ez)) and it looks good:

image

allenap:
@brookeangel I'm going to merge this to give me a chance of getting it into the monolith today. If there's anything else you find that needs to be fixed I'll do it in another PR. I hope that's okay.

PR #127, opened 2018-10-03
Quote PATH when setting env 

leapingfrogs:
Trying to build locally I hit issues as my path contained spaces (due to VMWare Fusion.app).

This quotes the PATH when setting up the SHELL variable.

> Then feel free to grab a reviewer from the list below and assign them to the PR!
>
> https://paper.dropbox.com/doc/noredink-ui-widget-library-fUK6yq32g187lxw2A60a8#:uid=619243092748985044355046&h2=Reviewers-list
>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 NPM_PREFIX=$(realpath .)/node_modules
 PATH:=${NPM_PREFIX}/.bin:${PATH}
-SHELL:=env PATH=${PATH} /bin/sh
+SHELL:=env PATH="${PATH}" /bin/sh
 
 .PHONY: test
 test: elm-stuff tests/elm-stuff node_modules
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 NPM_PREFIX=$(realpath .)/node_modules
 PATH:=${NPM_PREFIX}/.bin:${PATH}
```
avh4: Do we need to quote here too?

leapingfrogs: Somehow I missed that line when I submitted the quick fix yesterday.  

After testing it appears setting either, but not both of these fixes the problem I was seeing.  This is because the SHELL variable is set to the value set in the PATH defined above (i.e. it contains the NPM_PREFIX) as this shadows the previously defined PATH.  

Given that I've moved the quoting to the line defining the PATH so we can safely reuse PATH elsewhere in the makefile if necessary.

avh4:
👍

PR #129, opened 2018-10-04
Add Button.V5 to standardize link and button styles

brookeangel:
To do before merge: 
- [x] Test for compatibility in monolith (https://paper.dropbox.com/doc/Grafting-noredink-ui-into-the-monolith--AOae7awXY8Br7YEhbZunUEZdAg-ff7N9zwMyRR2D5Umza5Ez)
- [x] Get ok from @bendansby 
- [x] Get review from @allenap 

## What V5 does
- Button.V4 adds sweet high-rise buttons! 
- Button.V5 standardizes this behavior, so that links also rise high!

## Testing
There are lots of combos this should be tested on the styleguide app.
(`make styleguide-app/elm.js`)

Make vertical alignment and text looks great with: 
- [ ] Buttons (both width bo
  - [ ] Width unbounded
    - [ ] With a little text
    - [ ] With multiple lines of text
    - [ ] With long words that need to wrap lines
 - [ ] Width bounded
    - [ ] With a little text
    - [ ] With multiple lines of text
    - [ ] With long words that need to wrap lines
- [ ] Links
  - [ ] Width unbounded
    - [ ] With a little text
    - [ ] With multiple lines of text
    - [ ] With long words that need to wrap lines
 - [ ] Width bounded
    - [ ] With a little text
    - [ ] With multiple lines of text
    - [ ] With long words that need to wrap lines

## Screenshots
untitled

### DIFF
--- a/src/Nri/Ui/Button/V5.elm
+++ b/src/Nri/Ui/Button/V5.elm
@@ -1,0 +1,807 @@
+module Nri.Ui.Button.V5 exposing
+    ( ButtonSize(..), ButtonWidth(..), ButtonStyle(..), ButtonState(..), ButtonContent
+    , ButtonConfig, button, customButton, delete, copyToClipboard, ToggleButtonConfig, toggleButton
+    , LinkConfig, link, linkSpa, linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
+    )
+
+{-|
+
+
+# Changes from V4:
+
+  - Allows links with words exceeding button-width to wrap
+  - Standardizes vertical text alignment between buttons and links
+
+
+# About:
+
+Common NoRedInk buttons. For accessibility purposes, buttons that perform an
+action on the current page should be HTML `<button>` elements and are created here
+with `*Button` functions. Buttons that take the user to a new page should be
+HTML `<a>` elements and are created here with `*Link` functions. Both versions
+should be able to use the same CSS class in all cases.
+
+There will generally be a `*Button` and `*Link` version of each button style.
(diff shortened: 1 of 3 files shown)


allenap:
Here's [a diff from V4.elm to V5.elm](https://gist.github.com/allenap/a8c98b4d4eccd108edb93c1426af2fd8). There are more changes there than I expected.

I'm sorry for not doing this at the same time as I did the V4 button. I was so tunnelled in on getting buttons working that I was foolishly blind to links. Thank you for following up and doing this properly :-)

src/Nri/Ui/Button/V5.elm
```diff
@@ -0,0 +1,807 @@
...
+            , Css.bottom (Css.px 2)
+            , Css.verticalAlign Css.middle
+            ]
+        ]
+    ]
+
+
+styledName : String -> String
+styledName suffix =
+    "Nri-Ui-Button-V4-" ++ suffix
```
allenap: This ought to change to `V5`. Ah, maybe I'll do that!

allenap:
@brookeangel I rebased so pull first if you need make additional changes.

bendansby:
Beautiful!

src/Nri/Ui/Button/V5.elm
brookeangel: Whoopsie, many thanks for fixing!

brookeangel:
Thank you so much for making that diff Gavin! For whatever reason, never thought of doing that but it's very helpful.

PR #146, opened 2018-10-26
drop support for attributes and move class compatibility into factory

lukewestby:
This moves `class` compatibility into the CustomElements wrapper. We've been having issues with intermittent failures loading the shim and polyfill and this should alleviate those errors. It also removes the `onAttributeChange` function because properties are the best way to interop with Elm.


lib/CustomElement.js
```diff
@@ -130,44 +112,97 @@ exports.create = function create(config) {
...
+ * Attempt to make an ES6 class using the Function constructor rather than
+ * ordinary class syntax. The string we pass to the Function constructor is
+ * static so there is no script injection risk. It allows us to catch
+ * syntax errors at runtime for older browsers that don't support class and
+ * fall back to an ES5 constructor function.
+ */
+function makeMakeClass() {
+  try {
+    return new Function([
+      "return class extends HTMLElement {",
```
mthadley: That's a really interesting way to handle unsupported syntax. Nice!

mthadley:
Also, will these changes be ported over to the monolith as well?

lukewestby:
@mthadley yeah, we can import `noredink-ui/CustomElement` wherever we're using the monolith one

PR #151, opened 2018-10-29
Update icon dependencies

quelledanielle:
Updates latest modules to use
* `Icon.V4`
* `Button.V6`
* `SegmentedControl.V7`


src/Nri/Ui/Button/V6.elm
```diff
@@ -0,0 +1,813 @@
+module Nri.Ui.Button.V6 exposing
```
brookeangel: Thanks for updating button!

I don't want to force the scope of your PR to balloon in size, but I think I just realized that Button.V5.copyToClipboard may not work - can we make sure that the copyToClipboard button works ok in V6?

src/Nri/Ui/Button/V6.elm
```diff
@@ -0,0 +1,813 @@
...
+
+{-| Links are clickable things with a url.
+
+NOTE: Links do not support two-line labels.
+
+-}
+type alias LinkConfig =
+    { label : String
+    , icon : Maybe IconType
+    , url : String
```
brookeangel: The other thing I was wondering about was using `String` here for url - it seems like maybe we should pass in the wrapped Url type rather than a string. I know we do this in V5 as well, but I'm not quite sure why...

quelledanielle: Looks like it's always been `String`, but I agree that `Url` would be better!

brookeangel: Yeah! Weirdly Nri.Button wrapped over V2 to make it a URL... I'm not sure why we never did that in the noredink-ui package. 🤷‍ Anyways, either way works fine!

styleguide-app/Examples/Button.elm
```diff
@@ -11,8 +11,8 @@ import Html.Styled exposing (..)
 import Html.Styled.Attributes exposing (css)
 import ModuleExample as ModuleExample exposing (Category(..), ModuleExample, ModuleMessages)
 import Nri.Ui.AssetPath exposing (Asset)
-import Nri.Ui.Button.V5 as Button
-import Nri.Ui.Icon.V3 as Icon
+import Nri.Ui.Button.V6 as Button
+import Nri.Ui.Icon.V4 as Icon
```
brookeangel: screen shot 2018-10-29 at 4 15 55 pm

I think these little label-y thingies should be updated too!

PR #153, opened 2018-10-30
Add Select.V4 with optional id for select in config

lukewestby:
`Select.VX.customView` was removed to prevent manually styling selects with the extra attributes, but it was also used to apply IDs to selects for labels. This way we get the best of both worlds.

### DIFF
--- a/src/Nri/Ui/Select/V4.elm
+++ b/src/Nri/Ui/Select/V4.elm
@@ -1,0 +1,97 @@
+module Nri.Ui.Select.V4 exposing
+    ( Config
+    , view
+    )
+
+{-| Build a select input.
+
+
+## Upgrading to V3
+
+  - Remove the old styles. V3 uses compile-at-runtime elm-css!
+
+
+# Configure
+
+@docs Config
+
+
+# Render
+
+@docs view
+
+-}
+
+import Css
+import Dict
+import Html.Styled as Html exposing (..)
+import Html.Styled.Attributes as Attributes exposing (..)
+import Html.Styled.Events exposing (..)
+import Json.Decode exposing (Decoder, andThen, succeed)
+import Nri.Ui
+import Nri.Ui.Colors.V1
+import Nri.Ui.Util
+
+
+{-| Select-specific Choice.Config
+-}
+type alias Config a =
+    { choices : List { label : String, value : a }
+    , current : a
+    , id : Maybe String
+    }
+
+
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/Select/V4.elm
```diff
@@ -0,0 +1,97 @@
...
+    ( Config
+    , view
+    )
+
+{-| Build a select input.
+
+
+## Upgrading to V3
+
+  - Remove the old styles. V3 uses compile-at-runtime elm-css!
```
avh4: Should this section about upgrading to V3 be removed, or changed to refer to V4?

src/Nri/Ui/Select/V4.elm
```diff
@@ -0,0 +1,97 @@
...
+import Html.Styled as Html exposing (..)
+import Html.Styled.Attributes as Attributes exposing (..)
+import Html.Styled.Events exposing (..)
+import Json.Decode exposing (Decoder, andThen, succeed)
+import Nri.Ui
+import Nri.Ui.Colors.V1
+import Nri.Ui.Util
+
+
+{-| Select-specific Choice.Config
```
avh4: This comment is out of date.  `Choice` isn't a thing any more.

src/Nri/Ui/Select/V4.elm
```diff
@@ -0,0 +1,97 @@
...
+
+
+{-| TODO: Consider moving this to Nri.Ui.Util once the non-0.19-approved `toString` is removed
+-}
+niceId : String -> a -> String
+niceId prefix x =
+    prefix ++ "-" ++ Nri.Ui.Util.dashify (Nri.Ui.Util.removePunctuation (toString x))
+
+
+{-| A normal select dropdown
```
avh4: What does "normal" mean?  Can that qualification be dropped now that there are no other view functions in this module?

src/Nri/Ui/Select/V4.elm
```diff
@@ -0,0 +1,97 @@
...
+
+{-| Build a select input.
+
+
+## Upgrading to V3
+
+  - Remove the old styles. V3 uses compile-at-runtime elm-css!
+
+
+# Configure
```
avh4: 🎨 I think this "Configure" section can be combined with the "Render" section, since the `Config` type is only used for calling `view`.

lukewestby:
@avh4 i think a V5 for further changes would be alright. merging and publishing this will get 0.19 unblocked quickly.

PR #156, opened 2018-10-31
Remove elm-css-util

rtfeldman:
Removes `rtfeldman/elm-css-util` dependency in preparation for upgrading to Elm 0.19!

### DIFF
--- a/src/Nri/Ui/Text/V1.elm
+++ b/src/Nri/Ui/Text/V1.elm
@@ -34,7 +34,6 @@ module Nri.Ui.Text.V1 exposing
 
 import Css exposing (..)
 import Css.Foreign exposing (Snippet, children, descendants, everything, selector)
-import Css.Helpers exposing (identifierToString)
 import Html exposing (..)
 import Html.Styled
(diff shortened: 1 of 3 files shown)


src/Nri/Ui/Text/V1.elm
```diff
@@ -138,9 +137,45 @@ captionClassString =
...
+        SmallBodyGray ->
+            "SmallBodyGray"
+
+        Caption ->
+            "Caption"
+
+
+classToNamespacedString : CssClasses -> String
+classToNamespacedString classes =
+    classToString classes ++ namespace
```
terezka: 💭 Should this be `namespace ++ classToString classes`?

rtfeldman: Yeah that makes more sense!

The string should be unique either way, but this way would look weird in Inspect Element. 😄 

Changed in 2b11d2acb300e04d4fb6c775197fb6d7fa53ab94

terezka: Thanks!

PR #159, opened 2018-11-06
Deprecate Styles.V1

rtfeldman:
**DO NOT MERGE YET!**

We should merge this right before doing a `MAJOR` version bump.

It will require doing a find/replace across all dependent code bases to use `DEPRECATED.Nri.Ui.Styles.V1` instead of `Nri.Ui.Styles.V1`.

The public API of that module is the same as before, but marking it as deprecated will help make sure we aren't using it for new things.

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -7,6 +7,7 @@
         "src"
     ],
     "exposed-modules": [
+        "DEPRECATED.Nri.Ui.Styles.V1",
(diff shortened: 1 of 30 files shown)


rtfeldman:
Closing in favor of https://github.com/NoRedInk/noredink-ui/pull/163 - which will just remove this.

PR #160, opened 2018-11-06
Clean up unused imports

rtfeldman:
Remove some unused imports.


rtfeldman:
Merging without review since this is only deleting dead code!

PR #166, opened 2018-11-10
Puffins  final upgrades

lukewestby:
Final widget changes before 6.0.0 change


src/Nri/Ui/Tabs/V3.elm
```diff
@@ -0,0 +1,368 @@
...
+                            |> List.Zipper.previous
+                            |> Maybe.map (List.Zipper.current >> .id >> onSelect >> Json.Decode.succeed)
+                            |> Maybe.withDefault (Json.Decode.fail "No previous tab")
+
+                    else
+                        Json.Decode.fail "Wrong key code"
+                )
+                Events.keyCode
+        ]
+        [ Html.styled Html.button
```
lukewestby: this is one major change in this file

src/Nri/Ui/Tabs/V3.elm
```diff
@@ -0,0 +1,368 @@
...
+                    [ Css.color Nri.Ui.Colors.V1.navy
+                    , Css.display Css.inlineBlock
+                    , Css.padding4 (Css.px 14) (Css.px 20) (Css.px 12) (Css.px 20)
+                    , Css.textDecoration Css.none
+                    ]
+                    [ Attributes.href href ]
+                    [ Html.text tabLink.label ]
+
+            Nothing ->
+                Html.styled Html.button
```
lukewestby: this is the other. we just can't use anchor tags as buttons anymore. have to use the real thing!

terezka: Love how upgrading to 19 makes our code more correct <3

PR #168, opened 2018-11-13
Add Table.V4 with builtin keyframes

lukewestby:
So we don't have to include them in any legacy css


src-0.18/Nri/Ui/Table/V4.elm
```diff
@@ -0,0 +1,290 @@
...
+tableStyles =
+    [ borderCollapse collapse
+    , baseFont
+    , Css.width (Css.pct 100)
+    ]
+
+
+flash : Css.Animations.Keyframes {}
+flash =
+    Debug.log "" <|
```
terezka: 🔴Oops!

PR #180, opened 2018-11-26
🐧Fix width for select

terezka:
🦊 [Styling regression on teacher assignments index](https://www.pivotaltracker.com/story/show/161937786)

The `width: 100%` was not there in https://github.com/NoRedInk/noredink-ui/blob/6cbf5fece203cc2ef081be1b0584d8f143b236e8/src/Nri/Ui/Select/V1.elm and since the upgrade, it has caused problems. Removing it solves the trouble. @lukewestby Do you know if that style solved a particular problem?

### DIFF
--- a/src/Nri/Ui/Select/V5.elm
+++ b/src/Nri/Ui/Select/V5.elm
@@ -77,6 +77,5 @@ view config =
             , Css.cursor Css.pointer
             , Css.fontSize (Css.px 15)
             , Css.height (Css.px 45)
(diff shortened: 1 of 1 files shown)


lukewestby:
@terezka looks like this was intentional https://github.com/NoRedInk/noredink-ui/commit/1647b2f358c38ff485d52636528cd76443a697d3#diff-d7a588655ffda60d01928112bdda05ac

PR #186, opened 2018-12-11
Just use value instead of innerText

lukewestby:
This makes the textarea resettable from the model. We had to use `innerText` in Elm 0.18 because `value` had problems with synchronizing with the model correctly. Those problems are solved in 0.19.


Arkham:
Perfect!

PR #193, opened 2018-12-17
Only show scrollbar when needed

jamesgary:
Instead of this...

screen shot 2018-12-17 at 10 20 23 am

do this!

screen shot 2018-12-17 at 10 20 34 am


jamesgary:
@joneshf Thanks for the review! I should have linked the original request: https://github.com/NoRedInk/NoRedInk/pull/22327#issuecomment-447469940 But I think the protocol is if design requests it, we can merge right away.

PR #206, opened 2019-01-16
Fix autoresizing for TextArea

mthadley:
This fixes the broken autoresizing for `TextArea.V4`. We recently bumped it to a new version, but forgot to create a matching custom element for the new version.

`V4.js` is a shameless copy of `V3.js`, other than the tag name change to `nri-textarea-v4`.


mthadley:
@brookeangel Good idea! I threw in a quick comment about this whole situation.

PR #207, opened 2019-01-16
Bump version (6.6.1)

mthadley:
Bumps the version, which includes the fix for auto-resizing text areas.


mthadley:
@brookeangel It did not, but you're right, it will need to be bumped as well! Nice catch.

mthadley:
@brookeangel How does that look?

PR #210, opened 2019-01-17
Bump npm package

mthadley:
Bumping to release the export change from #209.


mthadley:
I'm actually going to resend this since I can actually batch this with some style fixes.

PR #211, opened 2019-01-18
Update TextArea font styles

mthadley:
This updates the font size and line height, as noted [here](https://www.pivotaltracker.com/story/show/162743330).

@NoRedInk/design How does this look?
image

### DIFF
--- a/src/Nri/Ui/InputStyles/V2.elm
+++ b/src/Nri/Ui/InputStyles/V2.elm
@@ -202,7 +202,7 @@ textAreaHeight =
 {-| -}
 writingLineHeight : Px
 writingLineHeight =
-    px 30
+    px 27
 
 
 {-| -}
--- a/src/Nri/Ui/InputStyles/V2.elm
+++ b/src/Nri/Ui/InputStyles/V2.elm
@@ -146,7 +146,7 @@ input theme isInError =
                     batch
                         [ sharedStyles
                         , Nri.Ui.Fonts.V1.quizFont
-                        , fontSize (px 20)
+                        , fontSize (px 18)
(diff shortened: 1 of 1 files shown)


bendansby:
The label doodad looks off. Did the text get bigger or did the container get smaller?

bendansby:
Here is staging 
image

mthadley:
Just compared and it looks like a difference of `line-height` in the label. It's `20px` on staging, and `15px` in the styleguide for this repo. Investigating why...

bendansby:
Looking at it now, I think I was mistaken in specifying a 30px line height. Could we do 27px instead? Apologies for the change!

mthadley:
How's this?
screen shot 2019-01-17 at 5 47 55 pm

bendansby:
Byootyful (on the line height)!

mthadley:
@bendansby Okay, looks like we have a global line-height of `20px` for all labels declared in one of the monolith stylesheets. These screenshots are from the styleguide app in this repo which doesn't contain that stylesheet, so instead we are getting `15px` from `reset.css`. I think when this reaches staging, it should look fine.

**TLDR:** Staging should be fine, it's just my screenshots from the nri-ui repo that look weird.

bendansby:
Oh our styling code base...

Thanks!

bendansby:
It's used in guided writes and peer reviews. And we want to apply these new styles to all instances.

mthadley:
@avh4 @bendansby I just double checked, and I only see `TextArea.writing` called in Peer Reviews and Guided Drafts. I think we should be good!

Thank you both for reviewing. 🙏

PR #221, opened 2019-01-28
add new icons

BrianHicks:
adds three new icons!

I also alphabetized the existing icons in the source.


BrianHicks:
no, nothing in particular. There just seem to be a lot of PNG icons vs SVG icons and stuff, and I don't do it enough to remember the 100% correct way. This is why we do PR review, so thank you 😄

PR #224, opened 2019-02-11
tweak little things for self-reviews

BrianHicks:
- change the cyan highlight color to be slightly lighter
- center the checkboxes when the label text spans multiple lines

image


BrianHicks:
@bendansby do these look about how you'd expect?

bendansby:
Oui!

BrianHicks:
what a gif 😆

PR #225, opened 2019-02-12
bump from 6.8.0 to 6.8.1

BrianHicks:
Minor change! Here's the whole diff (hint, it's only #224) https://github.com/NoRedInk/noredink-ui/compare/6.8.0...8649344


BrianHicks:
🤔 In this case, it's very minor changes. Color change is kinda feature level (I guess?) but the CSS change is a bugfix so I didn't mess with the versioning.

PR #226, opened 2019-02-12
Add node to nix

BrianHicks:
add some stuff I was missing from nix!

### DIFF
--- a/.envrc.nix
+++ b/.envrc.nix
@@ -1,1 +1,2 @@
 use nix
+export PATH="./node_modules/.bin:$PATH"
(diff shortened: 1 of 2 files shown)


omnibs:
Came here to test it on AV's request, seems to work:

```
$ which node
/nix/store/b7bv1hbj8c51a98mcvjk6m53bg6xyc2p-nodejs-8.11.4/bin/node
$ node
> console.log('hi!')
hi!
```

.envrc.nix
```diff
@@ -1 +1,2 @@
 use nix
+export PATH="./node_modules/.bin:$PATH"
```
avh4: is this preferred over just using `npx <thing>` to run thing?

allenap: ☁️ Make it an absolute path?

BrianHicks: @allenap ahh, good catch!

@avh4 I prefer it so that I can just run things "normally"

PR #238, opened 2019-03-28
Tessa/detangle icons

tesk9:
Hello! 

I'm trying to decouple the linking between the lists of assets and the views that sometimes include icons.

Also -- did folks know that the icon sprite sheet isn't being loaded properly into the styleguide app?  I'm addressing in another pr.


src/Nri/Ui/Button/V8.elm
```diff
@@ -0,0 +1,783 @@
...
+    , delete
+    , toggleButton
+    , link, linkSpa
+    , linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
+    )
+
+{-|
+
+
+# Changes from V7:
```
BrianHicks: I wanna put :tada: on all four bullet points below but that would be too much, so instead you get:

- :tada:
- :tada:
- :tada:
- :tada:

tesk9:
Thanks for reviewing!!

allymcknight:
. I approved this but I didn't do it in time but I still want to give you this gif I chose. Also Html Never 👏

allymcknight:
Apparently the gif doesn't work and this whole review thing is just a mess.

tesk9:
Haha link me link me I want to see this gif!

PR #240, opened 2019-03-28
Add make rule and index.html for gh-pages

tesk9:
WIP: Not branched off of master.

I think we can get the stylesheet conveniently available on gh pages, and then there'll be less reason to keep the monolith version up and running.

It's also hard to tell whether this is working or not given https://github.com/elm/url/issues/10. Also, I don't know how makefiles are supposed to work.

Before this actually works, we'll need to enable https://github.com/NoRedInk/noredink-ui/settings to use the docs/ folder on the master branch for github pages. But then we should be able to go to `noredink.github.io/noredink-ui` to see all the styleguide stuff.

I wonder if we should add a check for making sure that the gh pages version isn't out of date to the build? But also -- maybe get it working at all first :P

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,13 @@ styleguide-app/bundle.js: lib/index.js node_modules
 styleguide-app/elm.js: styleguide-app/bundle.js $(shell find src styleguide-app -type f -name '*.elm')
 	cd styleguide-app; npx elm make Main.elm --output=$(@F)
 
+
+.PHONY: gh-pages
+	# This is used by github pages to render the styleguide.
+    # You will not be able to test it locally due to https://github.com/elm/url/issues/10
+gh-pages:
+	cd styleguide-app; npx elm make Main.elm --output=./../docs/index.html; cd ..
+
 # plumbing
 
 node_modules: package.json
--- a/Makefile
+++ b/Makefile
@@ -34,12 +34,14 @@ styleguide-app/bundle.js: lib/index.js node_modules
 styleguide-app/elm.js: styleguide-app/bundle.js $(shell find src styleguide-app -type f -name '*.elm')
 	cd styleguide-app; npx elm make Main.elm --output=$(@F)
 
+docs: docs/index.html styleguide-app/elm.js docs/bundle.js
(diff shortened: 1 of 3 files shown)


BrianHicks:
> It's also hard to tell whether this is working or not given elm/url#10.

you can test by doing `cd docs; python -m http.server` (or if you have python 2.x installed, `python -m SimpleHTTPServer`)

> Also, I don't know how makefiles are supposed to work.

This would probably work OK and not require using phony targets:

```make
docs: docs/index.html
    touch -m $@

docs/index.html:
    cd styleguide-app; npx elm make Main.elm --output=./../docs/index.html
```

`docs/index.html` should probably list dependencies so it doesn't falsely think it's up-to-date. That's probably like `$(shell find styleguide-app src -name '*.elm')` or something.

tesk9:
Thanks! That definitely did it. Do you have a recommendation for dealing with assets? I'm thinking about whether the image tags are worth trying to get to show up, in particular

BrianHicks:
Hmmmmm. Maybe something like:

```make
SRC_ASSETS=$(wildcard styleguide-app/assets/* styleguide-app/assets/**/*)
DOCS_ASSETS=$(SRC_ASSETS:styleguide-app/%=docs/%)

docs/assets/%: styleguide-app/assets/%
    mkdir -p $(@D)
    cp $< $@

docs: $(DOCS_ASSETS)
```

That would make it so that `make public` would copy all the assets over from `styleguide-app`, and only the necessary ones, whenever it changed.

It's wildly magical though… I'd be happy to step through it or pair sometime, if that'd be helpful.

rtfeldman:
Just gonna throw this out there - I've used both GH Pages and Netlify in the past, and I found Netlify (also costs $0) much more enjoyable to use. They still have the "push your branch and it instantly deploys" workflow, including running whatever build scripts you ask it to (`make` or whatever), but they have various other niceties:

https://www.netlify.com/github-pages-vs-netlify/

It also only takes a few minutes to set up; you basically authorize Netlify to access your GH account and that gets you almost all the way there.

PR #242, opened 2019-03-28
Tessa/detangle icons banner alert

tesk9:
Uses bulb, checkmark, and exclamation mark elm/svgs in new module Nri.Ui.Alert.V3 and existing (but unpublished) Nri.Ui.BannerAlert.V3.

Screen Shot 2019-03-28 at 2 33 41 PM
Screen Shot 2019-03-28 at 2 33 33 PM

I merged `tessa/fix-stylesheet-build` (https://github.com/NoRedInk/noredink-ui/pull/239) in so that I could see what I was doing.

@NoRedInk/design


bendansby:
Bulb looks a bit big but otherwise 👍

tesk9:
Wait, which bulb? Both?

bendansby:
White bulb!

tesk9:
Ahh now I see it!

How's this:

Screen Shot 2019-03-29 at 2 37 19 PM

Also, I noticed that I've changed the navy. So the icon that we used to have, the navy was "#004889". Here, I'm setting the background color Nri.Ui.Colors.V1.navy, which is "#004e95". Is that okay, or do you want me to switch to the old image background color?

bendansby:
Looks good! Definitely want to use standard Navy 👍

src/Nri/Ui/BannerAlert/V3.elm
```diff
@@ -10,53 +10,53 @@ import Accessibility.Styled as Html exposing (Html)
...
+import Nri.Ui.SpriteSheet exposing (bulb, checkmark, exclamationMark)
+import Nri.Ui.Svg.V1 as NriSvg exposing (Svg)
 
 
 {-| A banner to show error alerts
 -}
-error : { a | exclamationPoint_svg : Asset } -> String -> Html msg
-error assets =
+error : String -> Html msg
+error =
```
jwoudenberg: 😻

src/Nri/Ui/SpriteSheet.elm
```diff
@@ -0,0 +1,91 @@
+module Nri.Ui.SpriteSheet exposing (checkmark, exclamationMark, bulb)
```
jwoudenberg: ❔ Do we want to expose this module so we can use these icons in monolith and monorepo directly?

tesk9: Maybe? Maybe we could wait until we need them though -- easier to expose new stuff than remove exposed stuff

jwoudenberg: Good point!

PR #243, opened 2019-04-01
make previewable styleguide

BrianHicks:
this adds a rule to make a public styleguide, which we can push with netlify. I'm gonna try and set that up next!


BrianHicks:
I can't figure out how to get it to make me a preview environment! Very strange. Must be because this isn't branched off of master.

styleguide-app/View.elm
```diff
@@ -29,7 +29,6 @@ view model =
             ]
             []
         , RootHtml.div [ Html.Attributes.id "svg-target" ] []
-        , RootHtml.node "SCRIPT" [ Html.Attributes.src "bundle.js" ] []
```
tesk9: There's also the fonts linked above on line 26

BrianHicks:
I've address the rest but…

> But I think that without the file watching, this won't work to develop against.

agreed! Do you think it's OK to ask people to use different versions of things than we would in other projects? For example, `parcel` is very nice and can do the building, watching, and publishing tasks.

tesk9:
I don't feel strongly about versions of things -- I think mostly, as a user of this repo, I want to not need to do lots of setup or worrying about versions. I don't have an opinion on parcel either. Maybe other folks would care to weigh in?

BrianHicks:
I'm'a just try it out in this branch and we can get some proof from the pudding.

BrianHicks:
update: it did not work.

- in serve/watch mode, it uses `--debug`. That has the `Map.!` issue on this repo.
- in build mode, it uses `--optimize`, which disallows our use of `Debug.toString` here.

overall, I'm sad that it didn't work out since I've used it successfully in other projects! Oh well.

Does the rest address your concerns @tesk9?

Makefile
```diff
@@ -18,12 +18,12 @@ format: node_modules
...
 .PHONY: clean
 clean:
-	rm -rf node_modules styleguide-app/elm.js styleguide-app/bundle.js $(shell find . -type d -name 'elm-stuff')
+	rm -rf node_modules styleguide-app/elm.js styleguide-app/bundle.js $(shell find . -type d -name 'elm-stuff') public
 
 .PHONY: styleguide-app
 styleguide-app: styleguide-app/elm.js
-	@echo "Visit http://localhost:8000/index.html to see the styleguide app in your browser"
-	cd styleguide-app && npx elm reactor
+	@echo "the styleguide-app command is deprecated; use serve-public."
```
tesk9: 💭 Why not remove it?

README.md
```diff
@@ -56,6 +56,14 @@ make styleguide-app
 
 If you'd like to test your widget in the monolith before publishing, check out the helpful tips [here](https://paper.dropbox.com/doc/Grafting-noredink-ui-into-the-monolith--AOae7awXY8Br7YEhbZunUEZdAg-ff7N9zwMyRR2D5Umza5Ez).
 
+### Public Styleguide
+
+We publish a public styleguide. To see a preview locally, run:
```
tesk9: There are references to `make styleguide-app` a few lines above this in the README. I think we should remove them, and make it clear that this is what to use when developing.

BrianHicks:
No, watching does not work with this. Alas! But I'm not sure if it worked with the elm reactor thing either? Did it for you? I could definitely just have been doing something wrong.

Makefile
BrianHicks: Because I bet a few people say `make styleguide-app` and it would be confusing if that was just not a thing anymore without any reason. This way seems more discoverable in that case.

tesk9: I wonder if the new thing could be called styleguide-app, in that case?

tesk9:
@BrianHicks yeah `make styleguide-app` definitely had watching for me 🤔

BrianHicks:
🤔 can we pair on this? I feel like there are a few moving pieces which would be good to resolve sync.

tesk9:
Yeah, for sure!

BrianHicks:
@tesk9 ready for review!

BrianHicks:
Whichever. The changes looked good so master seems fine. I’ll make it clean build in the morning

PR #244, opened 2019-04-03
fix key name in netlify.toml

BrianHicks:
netlify's key is `publish`, not `public`. Oops!


BrianHicks:
ah ha! It worked: https://deploy-preview-244--noredink-ui.netlify.com/

PR #245, opened 2019-04-04
Tessa/icons away

tesk9:
Adds a new version of the checkbox that inlines the checkbox icons.

Screen Shot 2019-04-04 at 12 16 12 PM

Also, removes the `noOp` msg from the checkbox configuration.


tesk9:
> How do we (do we?) reuse the same icon in an other component? It seems like we might have a need for a nice blue squiggle somewhere else.

I guess I kind of think of this the same way as code -- when we have a need, we can make it reusable. But it doesn't need to be reusable to start. Right now, to my knowledge, these icons are used anywhere else. I bet @NoRedInk/design has a better sense than me though! Design folks, do you foresee us needing to reuse the checkbox icons?

> How easy will it be to update these icons? It looks like it could be a bit awkward.

It's not too bad -- you need to convert the svg to elm/svg, but there are lots of tools for that. The tools definitely aren't perfect (the one I use doesn't include xlinkhrefs, probably because they're deprecated, but our svgs use them so 🤷‍♀️). But they're easy enough. The designers are already exporting the svgs from other programs -- converting to elm/svg is just another step.

We may run into trouble if we try to optimize the svgs though -- we won't have access to the general svg optimization tools of the web, unless we go back to the original svg. I think though that the design team isn't relying on us to manage their copies of the icons, so it doesn't seem that risky to me

bendansby:
We certainly use checkmarks and locks in other places throughout the site.

tesk9:
The backgrounds are actually part of these svgs -- do we use the lock with the background and the checkmark with the background elsewhere?
Or, should this module only store the background and use the checkmark and such from elsewhere? 🤔

bendansby:
No, we don't use those anywhere else (besides other versions of the checkbox floating about, like in checkbox.sass)

PR #250, opened 2019-04-09
Lab/active progress safari

tesk9:
Fixes https://www.pivotaltracker.com/story/show/165216075

## Chrome
chrome

## Safari
safari

## Firefox
firefox

cc @NoRedInk/design


bendansby:
Way to put those dots in their place!

PR #252, opened 2019-04-09
Lab/change clickable text alignemnt

tesk9:
Part of https://www.pivotaltracker.com/story/show/165216675

### Lock
Screen Shot 2019-04-09 at 4 50 22 PM

### Performance
Screen Shot 2019-04-09 at 4 50 12 PM

### Help
Screen Shot 2019-04-09 at 4 50 00 PM

I kinda feel like the lock and the performance icon look less good? But the help icon _definitely_ looks better and that's the one we use all over the place.

cc @NoRedInk/design


bendansby:
The lock and performance are visually bottom heavy, so that makes sense. Looks good!

bendansby:
Actually, I think there could stand to be a pixel or two more space between the icon and label. Let me see...

bendansby:
Yeah, 1 more px on medium and 2 on large. How's that for picky?

tesk9:
How's this?

Screen Shot 2019-04-10 at 9 20 56 AM

bendansby:
Beautimous!

PR #254, opened 2019-04-12
SlideModal expects the height to be specified in pixels instead of vh

juanedi:
It's a better fit for the only use case we have right now :-D


juanedi:
@tesk9 mind taking another look? I think I fixed the scrolling issue :-D

PR #257, opened 2019-04-17
Tessa/alert v4

tesk9:
Part of https://www.pivotaltracker.com/story/show/165389442

- Changes the error font color from `purpleDark` to `purple`
- ~~Adds `complexError` and `complexWarning` errors that support Html arguments~~
- Adds `somethingWentWrong` error for displaying stacktraces

Screen Shot 2019-04-17 at 2 10 05 PM

Screen Shot 2019-04-17 at 2 10 15 PM

cc @NoRedInk/design 

(Note -- for the somethingWentWrong error, it seemed like some of the colors were one-offs. I guestimatched the colors. Lmk if you want something different! (IMO, not red might be good? Red is scary and stacktraces are scary))

### DIFF
--- a/src/Nri/Ui/Alert/V4.elm
+++ b/src/Nri/Ui/Alert/V4.elm
@@ -33,7 +33,7 @@ error : String -> Html msg
 error content =
     alert
         [ exclamation Colors.purple
-        , viewAlertContent Colors.purpleDark content
+        , viewAlertContent Colors.purple content
         ]
 
 
--- a/src/Nri/Ui/Alert/V4.elm
+++ b/src/Nri/Ui/Alert/V4.elm
@@ -1,0 +1,136 @@
+module Nri.Ui.Alert.V4 exposing
+    ( error
+    , success
+    , tip
+    , warning
+    )
+
+{-| UI components that highlight information to the user.
+
+@docs error
+@docs success
+@docs tip
+@docs warning
+
+-}
+
+import Accessibility.Styled as Html exposing (Html)
+import Css
+import Css.Global
+import Html.Styled exposing (fromUnstyled)
+import Html.Styled.Attributes exposing (css)
+import Markdown
+import Nri.Ui
+import Nri.Ui.Colors.V1 as Colors
+import Nri.Ui.Fonts.V1 as Fonts
+import Nri.Ui.Icon.V3 as Icon
+import Nri.Ui.SpriteSheet exposing (bulb, checkmark, exclamationMark)
+import Nri.Ui.Svg.V1 as NriSvg exposing (Svg)
+
+
+{-| -}
+error : String -> Html msg
+error content =
+    alert
+        [ exclamation Colors.purple
+        , viewAlertContent Colors.purpleDark content
+        ]
+
+
+{-| -}
+success : String -> Html msg
+success content =
+    alert
+        [ iconContainer
+            [ Css.color Colors.white
+            , Css.backgroundColor Colors.green
+            ]
+            (Html.div
+                [ css
+                    [ Css.width (Css.px 12)
+                    , Css.height (Css.px 12)
+                    , Css.margin Css.auto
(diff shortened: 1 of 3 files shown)


src/Nri/Ui/Alert/V4.elm
```diff
@@ -0,0 +1,226 @@
...
+complexError : Html msg -> Html msg
+complexError content =
+    alert
+        [ exclamation Colors.purple
+        , alertHtml Colors.purple [ content ]
+        ]
+
+
+{-| -}
+complexWarning : Html msg -> Html msg
```
avh4: 💭 should we just make `warning` and `error` take Html instead of String, since it could still be used with `warning (text "...")` ?  I guess I like having the one that directly takes a string, but is it worth having a much bigger API?

avh4: I guess alternatively, why do we need alerts to be able to contain custom HTML?  Maybe an Alert is incorrectly being used in the places that's needed?  (Maybe could the styleguide example be updated to show an example of this that is an actual example of where custom HTML in the alert is needed?)

tesk9: Partly, there are a LOT of alert uses in the monolith already -- the API is already broad, so this work will streamline where the implementations live/forces to get the colors consistent (which they haven't) and get styles to match up and all.

I'll see if I can find some uses in the repo of the actual html alerts and use them in the exampels. I'm also not sure why we support markdown (do we actually use this?) but didn't want to check literally 80-plus pages 😅.

tesk9: Oh, I also deleted complexTip and complexSuccess from the monolith cause we weren't using them. I suspect that in the long run (maybe V5?) we'll delete the html support

bendansby: We had at least one alert that contained a link/javascript action (when editing a quick write's due date), but I think that should no longer be relevant (might not be actually be dead because feature flag). There may be others, not sure

bendansby:
Napkin 19 04-17-19, 3 39 34 PM

src/Nri/Ui/Alert/V4.elm
avh4: > Partly, there are a LOT of alert uses in the monolith already

Doesn't the module versioning we use solve this?  (Is there a reason you want all the callers to upgrade to V4 asap?)

tesk9: @avh4 So I converted a bunch of complex html alerts to markdown-based alerts in the monolith.
There's one html-based alert left:

Screen Shot 2019-04-17 at 4 11 55 PM

I don't see a way around that one being html-based.

Buuut I think that you're right that the API doesn't need to support html. So I'll drop the complex views from this PR.

tesk9: oops, missed your comments.

So the reason that I want to upgrade everything is that prior to today, we weren't actually on the Nri.Ui.Alert version consistently -- we were using `Nri.Alert` and `Nri.AlertDEPRECATED` and `Nri.Ui.Alert.V2`. Also, the color of the alert changed between some of the versions, and I like for colors to match up 😅

tesk9:
I think the caret glyph is going to take a littttle more work than I have bandwidth for right now. (If anyone is interested, you can read about how we would use an icon here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Customizing_the_disclosure_widget .. but I'm not sure how we'd do the directions for it.)

Also, at first I misunderstood what was supposed to be gray45 and it did not look good and I thought it was funny so here's a screenshot:
Screen Shot 2019-04-17 at 4 25 39 PM

And here's the actual view:
Screen Shot 2019-04-17 at 4 35 21 PM

bendansby:
Hard to choose which one is my favorite! 😛

PR #258, opened 2019-04-17
Update highlightCyan

tesk9:
https://www.pivotaltracker.com/story/show/165294495

Updates the highlightCyan color. Updating in-place because I don't think a small color shift is worth the complexity of a new version.

Screen Shot 2019-04-17 at 3 27 43 PM
cc @NoRedInk/design


bendansby:
That's the one!

PR #259, opened 2019-04-18
Add further overrides

tesk9:
Relevant to https://www.pivotaltracker.com/story/show/165389442 -- adds overrides to deal with bad sass in https://github.com/NoRedInk/NoRedInk/blob/master/app/assets/stylesheets/text.sass#L37

@brookeangel does this seem like it would solve the problem?

### DIFF
--- a/src/Nri/Ui/Alert/V3.elm
+++ b/src/Nri/Ui/Alert/V3.elm
@@ -130,7 +130,21 @@ viewAlertContent color content =
         , Css.fontSize (Css.px 13)
         , Css.lineHeight (Css.num 1.2)
         , Css.listStyleType Css.none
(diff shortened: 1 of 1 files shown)


brookeangel:
🙌 giving it a test in the monolith! _loading_

tesk9:
Thanks, Brooke! (Re font color -- at some point we switched to purpleDark, and that's what V3 uses. V4 switches back to purple. (I'll be sure the overwrite is either unnecessary or included before using V4))

Also -- I do not remember test-package existing! awesome! that's super handy.

brookeangel:
Read me update: https://github.com/NoRedInk/noredink-ui/pull/260

PR #263, opened 2019-04-22
Set the lineheight

tesk9:
Relevant to https://github.com/NoRedInk/NoRedInk/pull/23657#issuecomment-485046540

## Before:
Screen Shot 2019-04-22 at 9 10 23 AM

## After

Screen Shot 2019-04-22 at 9 09 54 AM

### DIFF
--- a/src/Nri/Ui/Alert/V3.elm
+++ b/src/Nri/Ui/Alert/V3.elm
@@ -108,6 +108,7 @@ iconContainer styles icon =
         (styles
             ++ [ -- Content positioning
                  Css.marginRight (Css.px 5)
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/Alert/V3.elm
```diff
@@ -108,6 +108,7 @@ iconContainer styles icon =
         (styles
             ++ [ -- Content positioning
                  Css.marginRight (Css.px 5)
+               , Css.lineHeight (Css.num 1.2)
```
brookeangel: 🤷‍♀️ Not sure what makes this fall at magic number 1.2, but looks good to me 🤷‍♀️

PR #264, opened 2019-04-23
Set the icon height agaainn

tesk9:
### Student sign up
Screen Shot 2019-04-23 at 12 13 16 PM

### Monolith styleguide
Screen Shot 2019-04-23 at 12 12 42 PM

### Noredink-ui styleguide (switching to V3 from V4)
Screen Shot 2019-04-23 at 12 09 41 PM

### DIFF
--- a/src/Nri/Ui/Alert/V3.elm
+++ b/src/Nri/Ui/Alert/V3.elm
@@ -108,14 +108,14 @@ iconContainer styles icon =
         (styles
             ++ [ -- Content positioning
                  Css.marginRight (Css.px 5)
-               , Css.lineHeight (Css.num 1.2)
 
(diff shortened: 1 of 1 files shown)


tesk9:
Heyyyyyyyyyyyyyy @brookeangel :P

I have an idea. Maybe I should request a QA of just the NRi.Alert stuff to avoid a re-re-re-vert or wherever we are now. 😭 but also how does this look?

BrianHicks:
is this for https://www.pivotaltracker.com/story/show/165537637?

brookeangel:
Hehe I agree that QA should maybe live check this with you before merge 😛 From a code perspective, it looks good

PR #274, opened 2019-05-14
Zamboni/customize page button label

glitteringkatie:
Unblocks https://www.pivotaltracker.com/story/show/165920196

Adds a type wrapper for the recovery button label so we can keep the existing `"return to " ++ blah` phrasing, add in a default `Reload the page` text (think for when the page breaks in an unrecoverable way), and adds in a Custom type that lets you set the whole of the button text.

image
image
image

cc @bendansby mostly for the okay on "Reload the page" phrasing since `ReturnTo` is exactly as it was and `Custom` is ... custom.

### DIFF
--- a/src/Nri/Ui/Page/V3.elm
+++ b/src/Nri/Ui/Page/V3.elm
@@ -1,0 +1,223 @@
+module Nri.Ui.Page.V3 exposing
+    ( DefaultPage, broken, blocked, notFound, noPermission
+    , RecoveryText(..)
+    )
+
+{-| A styled NRI issue page!
+
+@docs DefaultPage, broken, blocked, notFound, noPermission
+
+-}
+
+import Css exposing (..)
+import Html.Styled as Html exposing (Html)
+import Html.Styled.Attributes as Attributes
+import Nri.Ui.Button.V5 as Button
+import Nri.Ui.Text.V2 as Text
+
+
+{-| The default page information is for the button
+which will direct the user back to the main page of
+the SPA. Specify it's name and the message which will
+navigate to the page.
+-}
+type alias DefaultPage msg =
+    { link : msg
+    , recoveryText : RecoveryText String
+    }
+
+
+{-| ReturnTo just needs the name of the page the user
+will be returned to. Reload displays default text to reload
+the current page. Custom is to add flexibility to the button.
+-}
+type RecoveryText a
+    = ReturnTo a
+    | Reload
+    | Custom a
+
+
+{-| For the not found page.
+-}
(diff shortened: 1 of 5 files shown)


bendansby:
How about “Try again”

src/Nri/Ui/Page/V3.elm
```diff
@@ -0,0 +1,223 @@
...
+
+
+{-| ReturnTo just needs the name of the page the user
+will be returned to. Reload displays default text to reload
+the current page. Custom is to add flexibility to the button.
+-}
+type RecoveryText a
+    = ReturnTo a
+    | Reload
+    | Custom a
```
allenap: ☁️ Remove the type variable? It's always going to be `String`. If we change the definition of `recoveryText` above it would be a breaking change, and we'd need to create a `V4` module anyway.

allenap:
Fwiw, here's the diff between `Page/V2.elm` and `Page/V3.elm`:

```diff
1c1,4
< module Nri.Ui.Page.V2 exposing (DefaultPage, broken, blocked, notFound, noPermission)
---
> module Nri.Ui.Page.V3 exposing
>     ( DefaultPage, broken, blocked, notFound, noPermission
>     , RecoveryText(..)
>     )
5c8
< @docs DefaultPage, broken, blocked, notFound, noPermission
---
> @docs DefaultPage, broken, blocked, notFound, noPermission, RecoveryText(..)
23c26
<     , name : String
---
>     , recoveryText : RecoveryText String
26a30,39
> {-| ReturnTo just needs the name of the page the user
> will be returned to. Reload displays default text to reload
> the current page. Custom is to add flexibility to the button.
> -}
> type RecoveryText a
>     = ReturnTo a
>     | Reload
>     | Custom a
> 
> 
129c142,151
<                 { label = "Return to " ++ defaultPage.name
---
>                 { label =
>                     case defaultPage.recoveryText of
>                         ReturnTo name ->
>                             "Return to " ++ name
> 
>                         Reload ->
>                             "Reload the page"
> 
>                         Custom text ->
>                             text
```

src/Nri/Ui/Page/V3.elm
allenap: My other thought about this was to smush `DefaultPage` and `RecoveryText` into a single custom type:

```elm
type Config msg
    = ReturnTo String msg
    | Reload String msg
    | Custom String msg

-- ... and, later in the file, rename the
-- non-public Config to ViewConfig, say.
```

Just a thought, not a request.

glitteringkatie:
>How about “Try again”

✅ 
>☁️ Remove the type variable?

✅

PR #276, opened 2019-05-16
Add ability to provide a default selection to Select

glitteringkatie:
### Change
Allow an initial "nothing is selected" state.

### Use case
On the interests page we have form inputs for friends & pets which takes gender. Originally we used Dropdown for this reason but turns out when we visit the form and want previously saved info to be loaded, it doesn't respect the user telling it what option to have selected! So we chose to augment the more standard Select

### How it looks
default-select

### Other details
Came out of [this slack conversation](https://noredink.slack.com/archives/C0VVDLEES/p1558043241053400).

Possible weird edge cases that are being tested
* Something as default + Nothing currently selected -> the default is the currently selected value
* Something as default + Something currently selected -> The selected choice is selected
* Nothing as default + Something currently selected -> The selected choice is selected
* Nothing as default + Nothing currently selected -> The first choice is selected
(where "choice" is an item in the choice list)

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -38,6 +38,7 @@
         "Nri.Ui.PremiumCheckbox.V2",
         "Nri.Ui.SegmentedControl.V6",
         "Nri.Ui.Select.V5",
+        "Nri.Ui.Select.V6",
         "Nri.Ui.Table.V3",
         "Nri.Ui.Table.V4",
         "Nri.Ui.Tabs.V3",
--- a/elm.json
+++ b/elm.json
@@ -48,6 +48,7 @@
         "Nri.Ui.PremiumCheckbox.V4",
         "Nri.Ui.SegmentedControl.V6",
         "Nri.Ui.Select.V5",
+        "Nri.Ui.Select.V6",
         "Nri.Ui.Svg.V1",
         "Nri.Ui.Slide.V1",
         "Nri.Ui.SlideModal.V1",
--- a/src/Nri/Ui/Select/V6.elm
+++ b/src/Nri/Ui/Select/V6.elm
@@ -77,9 +77,18 @@ view config =
             config.id
                 |> Maybe.map (\id -> [ Attributes.id id ])
                 |> Maybe.withDefault []
+
+        currentVal =
+            if config.current == Nothing && config.defaultDisplayText == Nothing then
+                config.choices
+                    |> List.head
+                    |> Maybe.map .value
+
+            else
+                config.current
     in
     config.choices
-        |> List.map (viewChoice config.current config.valueToString)
+        |> List.map (viewChoice currentVal config.valueToString)
         |> (++) defaultOption
         |> Nri.Ui.styled Html.select
             "nri-select-menu"
--- a/src/Nri/Ui/Select/V6.elm
+++ b/src/Nri/Ui/Select/V6.elm
@@ -1,8 +1,8 @@
-module Nri.Ui.Select.V6 exposing (Config, view)
+module Nri.Ui.Select.V6 exposing (Config, Choice, view)
 
 {-| Build a select input.
 
-@docs Config, view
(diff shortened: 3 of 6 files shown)


glitteringkatie:
Here's the diff between the two versions! @allenap, how did you get the nicely colored version?
```
✨ diff src/Nri/Ui/Select/V6.elm src/Nri/Ui/Select/V5.elm
1c1
< module Nri.Ui.Select.V6 exposing (Config, view)
---
> module Nri.Ui.Select.V5 exposing (Config, view)
23,24c23,24
<     { choices : List (Choice a)
<     , current : Maybe a
---
>     { choices : List { label : String, value : a }
>     , current : a
27d26
<     , defaultDisplayText : Maybe String
31,34d29
< type alias Choice a =
<     { label : String, value : a }
<
<
61,74c56,62
<         defaultOption =
<             config.defaultDisplayText
<                 |> Maybe.map
<                     (\displayText ->
<                         [ Html.option
<                             [ Attributes.id (niceId "nri-select" "default")
<                             , Attributes.value (niceId "nri-select" "default")
<                             , Attributes.selected (config.current == Nothing)
<                             , disabled True
<                             ]
<                             [ text displayText ]
<                         ]
<                     )
<                 |> Maybe.withDefault []
---
>         viewChoice choice =
>             Html.option
>                 [ Attributes.id (niceId "nri-select" (config.valueToString choice.value))
>                 , Attributes.value (niceId "nri-select" (config.valueToString choice.value))
>                 , Attributes.selected (choice.value == config.current)
>                 ]
>                 [ Html.text choice.label ]
80,88d67
<
<         currentVal =
<             if config.current == Nothing && config.defaultDisplayText == Nothing then
<                 config.choices
<                     |> List.head
<                     |> Maybe.map .value
<
<             else
<                 config.current
91,92c70
<         |> List.map (viewChoice currentVal config.valueToString)
<         |> (++) defaultOption
---
>         |> List.map viewChoice
105,120d82
<
<
< viewChoice : Maybe a -> (a -> String) -> Choice a -> Html a
< viewChoice current toString choice =
<     let
<         isSelected =
<             current
<                 |> Maybe.map ((==) choice.value)
<                 |> Maybe.withDefault False
<     in
<     Html.option
<         [ Attributes.id (niceId "nri-select" (toString choice.value))
<         , Attributes.value (niceId "nri-select" (toString choice.value))
<         , Attributes.selected isSelected
<         ]
<         [ Html.text choice.label ]
```

allenap:
> @allenap, how did you get the nicely colored version?

Give a `diff` hint at the opening of the backticks:

image

which results in:

```diff
diff src/Nri/Ui/Select/V6.elm src/Nri/Ui/Select/V5.elm
1c1
< module Nri.Ui.Select.V6 exposing (Config, view)
---
> module Nri.Ui.Select.V5 exposing (Config, view)
...
```

I usually also generate "unified" diffs (with `diff -u`) which is probably what you're used to seeing more often, but I think GitHub will highlight any style.

allenap:
@glitteringkatie I pushed a few commits. It's tidying really, of stuff from V5, but 7931adff73598ca0cefab893386bd0b6a2626813 is potentially interesting. I don't think we need to set `id` or `value` on the default, disabled, choice, but maybe there's a good reason for it? I was also concerned that the choice (`nri-select-default`) could legitimately coincide with a real choice. Not sure that actually matters. It's a bit confusing.

glitteringkatie:
> 7931adf is potentially interesting. I don't think we need to set id or value on the default, disabled, choice, but maybe there's a good reason for it? I was also concerned that the choice (nri-select-default) could legitimately coincide with a real choice. Not sure that actually matters. It's a bit confusing.

Yeah, I was cargo culting the "id" and "value" so I'm glad we could delete! Thanks for the clean up, it looks great.

PR #278, opened 2019-05-17
:skull: Remove old table and deprecated sytles

tesk9:
Removes `Nri.Ui.Table.V3` (which is not used in the monolith) and a deprecated module aimed at helping ease the transition to elm-css 0.15+.

Should we do this? I'm not sure. I dislike having stuff called DEPRECATED floating around. And I don't think a major version bump is thaatt big of a deal. 

But maybe we should get a bucket of stuff together (like killing old Icon module versions) and do a single major version bump?


tesk9:
Failure is:

```
if (npx elm diff | tee /dev/stderr | grep -q MAJOR); then echo "MAJOR changes are not allowed!"; exit 1; fi
This is a MAJOR change.
---- REMOVED MODULES - MAJOR ----
    Nri.Ui.Table.V3
MAJOR changes are not allowed!
```

So judgy!

But I assume we'll want a major version change eventually, right?

quelledanielle:
🤷‍♂ `Nri.Ui.Table.V2` was deleted so it seems nice to me that `V3` can be deleted as well

brookeangel:
@tesk9 - resolved a merge conflict to prepare for merge

PR #279, opened 2019-05-18
Always use a class to create a custom element

lukewestby:
Just cleaning some stuff up

**⚠️DO NOT MERGE ⚠️**
This won't be usable until the next release of Edge https://caniuse.com/#feat=custom-elementsv1


stoeffel:
@lukewestby can we close this or is this still in use?

PR #285, opened 2019-05-29
Fix padding and add dismiss label to banner alert

tremlab:
Explicitly set padding for the banner alert "dismiss" button ('x').  (This was displaying strangely on the monolith due to its css resets.)

Including "Dismiss banner" aria-label

### DIFF
--- a/src/Nri/Ui/BannerAlert/V4.elm
+++ b/src/Nri/Ui/BannerAlert/V4.elm
@@ -7,6 +7,7 @@ module Nri.Ui.BannerAlert.V4 exposing (alert, error, neutral, success)
 -}
 
 import Accessibility.Styled as Html exposing (Html)
+import Accessibility.Styled.Widget as Widget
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/BannerAlert/V4.elm
```diff
@@ -144,16 +145,17 @@ dismissButton msg =
         []
         [ Html.button
             [ Html.Styled.Events.onClick msg
+            , Widget.label "Dismiss banner"
```
tesk9: 🙌

tremlab:
> Looks good! Although I didn't verify in the monolith. 😅
> 
> One thing -- I expect if you ask @NoRedInk/design, they'll ask that you change the cursor behavior on the button to `pointer`.

you are 100% correct!   😆 Thanks for pointing this out.  Fixed in a second commit.

PR #286, opened 2019-05-29
Lab/remove unused icons

juanedi:
This PR removes some icons that are now inlined in the monolith.

Replaced use cases can be found here: https://github.com/NoRedInk/NoRedInk/pull/23994
Part of https://www.pivotaltracker.com/story/show/166191815

### DIFF
--- a/src/Nri/Ui/Button/V6.elm
+++ b/src/Nri/Ui/Button/V6.elm
@@ -63,7 +63,7 @@ import Nri.Ui.AssetPath as AssetPath exposing (Asset)
 import Nri.Ui.Colors.Extra as ColorsExtra
 import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Fonts.V1
-import Nri.Ui.Icon.V4 as Icon exposing (IconType)
+import Nri.Ui.Icon.V5 as Icon exposing (IconType)
 
 
 {-| Sizes for buttons and links that have button classes
--- a/src/Nri/Ui/Button/V6.elm
+++ b/src/Nri/Ui/Button/V6.elm
@@ -63,7 +63,7 @@ import Nri.Ui.AssetPath as AssetPath exposing (Asset)
 import Nri.Ui.Colors.Extra as ColorsExtra
 import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Fonts.V1
-import Nri.Ui.Icon.V5 as Icon exposing (IconType)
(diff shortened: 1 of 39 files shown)


src/Nri/Ui/Button/V6.elm
```diff
@@ -63,7 +63,7 @@ import Nri.Ui.AssetPath as AssetPath exposing (Asset)
 import Nri.Ui.Colors.Extra as ColorsExtra
 import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Fonts.V1
-import Nri.Ui.Icon.V4 as Icon exposing (IconType)
+import Nri.Ui.Icon.V5 as Icon exposing (IconType)
```
tesk9: Ooo wait we can't do this, can we? Won't this change will break all our existing uses of Nri.Ui.Button.V6?

I ran `elm diff`, and it says this is fine, which seems suspicious to me. I'm going to investigate a bit more.

tesk9: Yeah it looks like it's this: https://github.com/elm/compiler/issues/1777

juanedi: Ouch, good catch! I'll change that back to how it was before.

tesk9:
@juanedi I deleted a bunch more stuff! how does this look to you?

src/Nri/Ui/Icon/V5.elm
```diff
@@ -1,173 +1,109 @@
...
-@docs assignmentTypeQuickWrite
-@docs assignmentTypeQuiz
-@docs assignmentTypeSelfReview
-@docs assignmentTypeWritingCycle
-@docs attention
-@docs bang
-@docs bulb
-@docs calendar
-@docs caret
+
```
juanedi: 😻

src/Nri/Ui/Button/V6.elm
```diff
@@ -63,7 +63,7 @@ import Nri.Ui.AssetPath as AssetPath exposing (Asset)
 import Nri.Ui.Colors.Extra as ColorsExtra
 import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Fonts.V1
-import Nri.Ui.Icon.V5 as Icon exposing (IconType)
```
juanedi: Thank you for fixing this!

juanedi:
This looks great to me! 🎉
Should we merge?

PR #293, opened 2019-06-04
Accessible Modal

tesk9:
Adds a new version of the Modal module. 

This changes the API pretty dramatically. 

Now, users of the module must tag the first or last focusable element in the modal. This enables us to keep the user's focus within the modal -- preventing the user from interacting with elements behind the overlay.

We can also now exit the modal with the escape key (which means that users of this module need to wire in subscriptions).

I'm not sure that the API here is final -- it doesn't seem obviously usable to me yet. I'm looking for suggestions on how to improve it. Thanks!

## Views & gifs
accessible modal

Screen Shot 2019-06-11 at 12 12 37 PM
Screen Shot 2019-06-11 at 12 12 29 PM


tesk9:
Okay, so for accessibility, we'll need a key trap to keep the user's focus in the modal, and we'll need escape to work properly.

```
module Main exposing (main)

import Accessibility.Key as Key
import Browser
import Browser.Dom exposing (focus)
import Browser.Events
import Html exposing (Html, button, div, text)
import Html.Attributes exposing (id, style)
import Html.Events exposing (onClick)
import Task


type alias Model =
    { modal : Modal
    }


initialModel : Model
initialModel =
    { modal = Closed
    }


type Modal
    = Opened
    | Closed


type Msg
    = OpenModal
    | CloseModal
    | Focus String
    | Focused (Result Browser.Dom.Error ())


update : Msg -> Model -> ( Model, Cmd Msg )
update msg model =
    case msg of
        OpenModal ->
            ( { model | modal = Opened }
            , Cmd.none
            )

        CloseModal ->
            ( { model | modal = Closed }
            , Cmd.none
            )

        Focus id ->
            ( model, Task.attempt Focused (focus id) )

        Focused _ ->
            ( model, Cmd.none )


view : Model -> Html Msg
view model =
    div []
        [ button [] [ text "Supports keyboard focus" ]
        , case model.modal of
            Opened ->
                div
                    [ style "border" "2px solid blue"
                    , id "modal"
                    ]
                    [ text "Modal!"
                    , div []
                        [ button
                            [ onClick CloseModal
                            , Key.onKeyDown [ Key.tabBack (Focus "last-button") ]
                            , id "first-button"
                            ]
                            [ text "Close Modal" ]
                        , button
                            [ Key.onKeyDown [ Key.tab (Focus "first-button") ]
                            , id "last-button"
                            ]
                            [ text "other action" ]
                        ]
                    ]

            Closed ->
                button [ onClick OpenModal ] [ text "Open Modal" ]
        , button [] [ text "Supports focus" ]
        ]


main : Program () Model Msg
main =
    Browser.element
        { init = \flags -> ( initialModel, Cmd.none )
        , subscriptions =
            \model ->
                case model.modal of
                    Opened ->
                        Browser.Events.onKeyDown (Key.escape CloseModal)

                    Closed ->
                        Sub.none
        , view = view
        , update = update
        }
```

avh4:
Comments on tesk9/modal:

```
type alias Model =
    { state : State
    , dismissOnEscAndOverlayClick : Bool
    }
init : { dismissOnEscAndOverlayClick : Bool } -> Model
```

I think following Evan's sortable-table example, it's better to not tie the user-provided config and the module-managed state together:  I'd suggest instead:

```
type alias Config = { dismissOnEscAndOverlayClick : Bool }
init : State
update : Config -> State -> (State, Cmd Msg)
view : Config -> State -> Html msg
```

---

```
view :
    { overlayColor : String
    , wrapMsg : Msg -> msg
    , modalAttributes : List (Attribute Never)
    , title : ( String, List (Attribute Never) )
    , content : Html msg
    }
    -> Model
    -> Html msg
firstFocusableElement : List (Attribute Msg)
```

I think this will be a bit circuitous to work with, since `firstFocusableElement` gives `Attribute Msg`, but then you'll have to apply your `wrapMsg` to it (which you already have to provide in the config record anyway).

I wonder if something like this could work and keep things simpler for the caller:

```
view :
    { overlayColor : String
    , wrapMsg : Msg -> msg
    , modalAttributes : List (Attribute Never)
    , title : ( String, List (Attribute Never) )
    , content : 
        { firstFocusableElement : List (Attribute msg)
        , ...
        }
           -> Html msg
    }
    -> Model
    -> Html msg
```

avh4:
💭 for tesk9/modal -- maybe the module in that package should be called `Accessibility.Modal` or something like that to avoid naming conflicts with applications that define a module called "Modal" ?

tesk9:
` it's better to not tie the user-provided config and the module-managed state together` <- this point and your suggestion are really interesting!! It definitely made things feel messy to have that setting in the model, and I only did it so the data would be available to subscriptions too. I'm excited to try that out and see how it goes.


---

For this suggestion, I don't quite see how things would fit together. We wouldn't know how or where or what the firstFocusableElement was supposed to be, would we?

```
view :
    { overlayColor : String
    , wrapMsg : Msg -> msg
    , modalAttributes : List (Attribute Never)
    , title : ( String, List (Attribute Never) )
    , content : 
        { firstFocusableElement : List (Attribute msg)
        , ...
        }
           -> Html msg
    }
    -> Model
    -> Html msg
```

tesk9:
Oooohhh wait I'm following now!

tesk9:
Okay, I thinkkkk that's nicer now? what do you think?

tesk9:
@avh4 is this okay to merge?

styleguide-app/Examples/Modal.elm
```diff
@@ -52,50 +53,60 @@ example parentMessage state =
...
     }
 
 
-viewContent state wrapMsg primaryButton { firstFocusableElement, lastFocusableElement } =
+viewContent :
+    State
+    -> (Modal.Msg -> Msg)
+    -> (Msg -> String -> List (Root.Attribute Msg) -> Html Msg)
+    ->
+        { a
```
avh4: 🎨 I think it would be ideal to have the Modal package define a type alias for this and just use that rather than record extension

styleguide-app/Examples/Modal.elm
```diff
@@ -52,50 +53,60 @@ example parentMessage state =
...
         ]
             |> List.map (Html.map parentMessage)
     }
 
 
-viewContent state wrapMsg primaryButton { firstFocusableElement, lastFocusableElement } =
+viewContent :
+    State
+    -> (Modal.Msg -> Msg)
+    -> (Msg -> String -> List (Root.Attribute Msg) -> Html Msg)
```
avh4: 🎨  could Modal define a union type of `Modal.Style = Danger | Primary | Secondary`  or even reuse Button.V8.ButtonStyle to avoid the need to pass a function as a configuration argument?

src/Nri/Ui/Modal/V5.elm
```diff
@@ -1,17 +1,59 @@
...
     , primaryButton, secondaryButton, dangerButton
     )
 
 {-| Changes from V4:
 
-@docs info, warning
+  - Remove dependence on Assets
+  - Adds keyboard support (escape key to exit, tabs contained within the modal)
+
+These changes have required major API changes. Be sure to wire up subscriptions!
```
avh4: Hmm, is there some we we could check whether subscriptions were wired up?   I can't think of an easy way to do that.

src/Nri/Ui/Modal/V5.elm
```diff
@@ -1,17 +1,59 @@
...
+                    div []
+                        [ Modal.viewContent [ text "Content goes here!" ]
+                        , Modal.viewFooter
+                            [ Modal.primaryButton DoSomething "Continue" onlyFocusableElement
+                            , text "`onlyFocusableElement` will trap the focus on the 'Continue' button."
+                            ]
+                        ]
+            }
+            state
+
```
avh4: 💭 could you add writing up subscriptions to the example?

src/Nri/Ui/Modal/V5.elm
tesk9: Yeah, it seems really easy to miss! I'm not sure

tesk9:
> Thanks for trying out the Accessibility.Modal suggestions! Does this seem better?

It definitely does! In particular, these https://github.com/NoRedInk/noredink-ui/pull/293#issuecomment-500999190 made it feel much nicer

Issue #299, opened 2019-06-21
Modal.V5 needs


tesk9:
1. Manual close/open helpers
2. More flexible button helpers (especially disabled)
3. Launch button helper missing NRI styles
3. API requires that the attributes passed to `primaryButton`, etc. be `Html.Attribute` rather than `Html.Styled.Attribute`. This is inconvenient/has bad error messages.
4. Helper for scrolling modal contents
5. `{title: {title:"", visibleTitle:""}, ..}` naming is weird

PR #301, opened 2019-06-21
Tessa/accessible modal cont

tesk9:
In the course of working on my hack day project, I used the newest modal version and found some areas to be improved. Listed here: https://github.com/NoRedInk/noredink-ui/issues/299

This PR makes some of the improvements (directly on the old V5 modal version -- if this is the approach we want to take, I'll make V6 and preserve V5 as-is) but I wanted to talk about implications.

So the new Modal keeps track of focus, and requires that the user label the focusable elements in the modal as such via Html Attributes. Ultimately, this approach means that all of our focusable widgets would need to accept an arbitrary list of attributes as an argument so that we could properly label them as focusable elements in the modal. Please see Button.V8 in this PR as an example.

Is this fine? Is this good? Are there other options we should consider?


bendansby:
I don't know if this is relevant, but I'd love for our modals to follow a common structure, so that where available, Escape activates the Cancel and X buttons, and Return/Enter activates the "primary" action (usually denoted with a Primary button).

tesk9:
@bendansby in V5, the escape key does exit the modal! (if allowed) And if the continue button is the only button or the first button on the modal, it'll get the focus when the modal opens, so return/space should accept/continue/do the primary action

src/Nri/Ui/Modal/V5.elm
```diff
@@ -114,49 +115,77 @@ update config msg model =
...
+    , firstFocusableElement : List (Attribute msg)
+    , lastFocusableElement : List (Attribute msg)
     }
 
 
 {-| -}
 info :
-    { title : { visibleTitle : Bool, title : String }
+    { visibleTitle : Bool
+    , title : String
```
avh4: 👍

avh4: Just to confirm that I understand:  this is a Bool * String instead of a Maybe String because we always use the title in the acciessibility labels even if it's not shown?

src/Nri/Ui/Modal/V5.elm
```diff
@@ -114,49 +115,77 @@ update config msg model =
...
-        , content = config.content >> toUnstyled
+        , title = viewTitle titleColor { title = config.title, visibleTitle = config.visibleTitle }
+        , content =
+            \{ onlyFocusableElement, firstFocusableElement, lastFocusableElement } ->
+                { onlyFocusableElement = List.map Html.Styled.Attributes.fromUnstyled onlyFocusableElement
+                , firstFocusableElement = List.map Html.Styled.Attributes.fromUnstyled firstFocusableElement
+                , lastFocusableElement = List.map Html.Styled.Attributes.fromUnstyled lastFocusableElement
+                }
+                    |> config.content
+                    |> toUnstyled
```
avh4: 👍 yeah, kinda unfortunately that it's so verbose to convert between Html.Styled vs Html, but thanks for doing this!

src/Nri/Ui/Modal/V5.elm
```diff
@@ -238,18 +267,28 @@ viewFooter =
...
 {-| -}
-launchButton : (Msg -> msg) -> List Css.Style -> String -> Html msg
-launchButton wrapMsg styles label =
-    button
-        (css styles
-            :: List.map Html.Styled.Attributes.fromUnstyled
-                (Modal.openOnClick wrapMsg (String.replace " " "-" label))
-        )
-        [ text label ]
+launchButton : (Msg -> msg) -> String -> Html msg
```
avh4: ❓ What is a launch button?  Is this for opening the modal?  Wouldn't different events cause different modals to open?  I'm confused about what this is for... if you know, could you explain it in the doc comments for this function?

src/Nri/Ui/Button/V8.elm
```diff
@@ -154,6 +154,24 @@ button :
         }
     -> Html msg
 button config content =
+    customizableButton config content []
+
+
+{-| -}
+customizableButton :
```
avh4: 💭 rather than adding this as a new extra function, I wonder if now is the time to change the Button API to the pattern of taking attributes?  Roughly like...

```elm
module ...Button.V9

type Attribute
    = Size ButtonSize
    | Style ButtonStyle
    | ExtraAttribute Sytyled.Attribute msg
    | Width ButtonWidth
    ...


button : msg -> List Attribute -> Html msg
link : Url -> List Attribute -> Html msg
```

Then all the attributes would be optional, and new flexibility like this customizable button would fit into the existing API rather than requiring a new function to be added.

src/Nri/Ui/Modal/V5.elm
tesk9: Yes, exactly!

src/Nri/Ui/Button/V8.elm
tesk9: Oooooo yes!!! Love this idea

PR #307, opened 2019-07-08
Icon in Alert module doesn't shrink

juanedi:
Add `flex-shrink 0` so that the icon doesn't shrink when using `Alert.tip` and similar views.

I need it to turn this...

Screen Shot 2019-07-08 at 11 48 42

into this...

Screen Shot 2019-07-08 at 11 48 54

I didn't find any other imports of `Nri.Ui.Alert.V4` in the monolith yet, so changing the style should be pretty safe I guess.

### DIFF
--- a/src/Nri/Ui/Alert/V4.elm
+++ b/src/Nri/Ui/Alert/V4.elm
@@ -167,6 +167,7 @@ iconContainer styles icon =
                , Css.justifyContent Css.center
                , Css.marginRight (Css.px 5)
                , Css.lineHeight (Css.px 13)
+               , Css.flexShrink Css.zero
 
                -- Size
                , Css.borderRadius (Css.px 13)


juanedi:
While we are at it... according to [the design spec](https://app.zeplin.io/project/5cdde2fe4c7fea1d6280c5be/screen/5ce854fb617ba078b57b4df8), the tip should have `font-size: 15px`.

@tesk9 do you think it would make sense to make the font size configurable? Should I override it via CSS on the call site?

staceyadams:
For this case, I chose to override the normal alert style because:
* Color change - to help it stand out so that it's not blue sandwiched between more blue on a blue background
* Font size change - because the normal alert font size is pretty small and this tip should draw more attention than the "I want to continue without..." link

If it makes any difference, the font change maps to small-body (as far as size and color; I'm not sure what line spacing changes might have been tweaked).

I'm also not sure how reusable this will be in the future. For the most part, I expect that we would stick to the normal alert style.

tesk9:
@juanedi based on what Stacey said, I think overriding at the call site is probably the easiest thing. Alternatively, maybe this is a place to introduce the builder pattern? that would mean a new Alert version...

juanedi:
Yup! I was also leaning towards overriding since this seems like a one-off.
I'll make a PR with the version bump to get this into the monolith 🤘

PR #310, opened 2019-07-11
Tessa/adds sortable table

tesk9:
Adds SortableTable to the styleguide.

Changes from the monolith version:
- Uses Svg directly for the sort arrows
- Uses Styled approach rather than a separate stylesheet

Screen Shot 2019-07-11 at 1 54 33 PM

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/SortableTable/V1.elm
+++ b/src/Nri/Ui/SortableTable/V1.elm
@@ -23,8 +23,8 @@ import Nri.Ui.Colors.V1
 import Nri.Ui.CssVendorPrefix.V1 as CssVendorPrefix
 import Nri.Ui.Icon.V5 as Icon
 import Nri.Ui.Table.V4
-import Svg exposing (Svg)
-import Svg.Attributes
+import Svg.Styled as Svg exposing (Svg)
+import Svg.Styled.Attributes as SvgAttributes
 
 
 type SortDirection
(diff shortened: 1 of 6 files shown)


src/Nri/Ui/SortableTable/V1.elm
```diff
@@ -0,0 +1,362 @@
+module Nri.Ui.SortableTable.V1 exposing
+    ( Column, Config, Sorter, State
+    , init, initDescending
+    , custom, string, view, viewLoading
+    , invariantSort, simpleSort, combineSorters
+    )
```
brookeangel: Out of scope side note... but any idea what https://package.elm-lang.org/packages/NoRedInk/elm-sortable-table/latest/Table is?

src/Nri/Ui/SortableTable/V1.elm
```diff
@@ -0,0 +1,362 @@
...
+            folder =
+                \sorter acc ->
+                    case acc of
+                        EQ ->
+                            sorter sortDirection elem1 elem2
+
+                        _ ->
+                            acc
+        in
+        List.foldl folder EQ sorters
```
brookeangel: Totes out of scope for this... but didn't we have some other implementation of Sort in the monolith? Maybe it was https://package.elm-lang.org/packages/rtfeldman/elm-sorter-experiment/latest/Sort ? (Just a side note, not suggesting that anything is consolidated here)

src/Nri/Ui/SortableTable/V1.elm
tesk9: Uhhhhhhh I don't remember that! It looks like it's a fork of evancz/elm-sortable-table -- I wonder why we forked it..

PR #312, opened 2019-07-11
Tessa/more flexible button

tesk9:
Experimental re-write of Button.

Goals:
- support custom attributes
- flexibility
- ease of setup of typical buttons

See note here: https://github.com/NoRedInk/noredink-ui/pull/301/files#r296427302


tesk9:
cc @avh4  -- this is the work I've been doing since we talked about more flexible buttons on the modal PR. It ended up being a big project!

styleguide-app/Examples/Button.elm
```diff
@@ -139,79 +147,73 @@ viewButtonExamples messages (State control) =
...
-        , icon = Nothing
-        , url = "#"
-        }
+    , Button.build
+        |> Button.withLabel "linkExternalWithTracking"
+        |> Button.unboundedWidth
+        |> Button.secondary
+        |> Button.onClick (messages.showItWorked "linkExternalWithTracking clicked")
+        |> Button.linkExternalWithTracking
+        |> Button.renderLink
```
avh4: ❓ I wonder what the tradeoffs are between a pipeline API like this and an "attributes" API like

```elm
Button.button
    [ Button.label "linkExternalWithTracking"
    , Button.unboundedWidth
    , Button.secondary
    , Button.onClick msg
    ]
```

I wonder if @stoeffel has thoughts about this?

avh4: Ah, here's the thing from Stöffel that I was thinking of: https://github.com/NoRedInk/engblog/pull/68

stoeffel: In general, I came to think that a function with many optional fields in its configuration is an indication that it might do too many things and splitting it up into more specific functions might be better.

In this case though, I think having an html-like api or pipeline api will be an improvement.
I wonder which api gives the better error messages. 
Two things I like about the html-like api:
* no need to render at the end of the pipeline
* it feels like html :-)

tesk9: I went with the builder-render pattern mainly because @BrianHicks has been enthusiastic about it recently. I definitely like stuff to feel like html tho!

Maybe I'll try out an alternate attributes approach and see what that feels like 🤔

tesk9:
I think the attributes-based approach is panning out really nicely! closing this PR in favor of https://github.com/NoRedInk/noredink-ui/pull/319

PR #313, opened 2019-07-12
Remove P+ in Checkbox V3

bendansby:
Designed to make these checkboxes in Peer Review on the assignment form stop saying P+

Screen Shot 2019-07-12 at 9 13 18 AM

### DIFF
--- a/src/Nri/Ui/PremiumCheckbox/V3.elm
+++ b/src/Nri/Ui/PremiumCheckbox/V3.elm
@@ -87,7 +87,7 @@ premium assets config =
                                     assets.iconPremiumFlag_svg
 
                                 PremiumWithWriting ->
-                                    assets.iconPremiumWithWritingFlag_svg
+                                    assets.iconPremiumFlag_svg
                             )
                         , backgroundRepeat noRepeat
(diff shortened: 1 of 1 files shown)


brookeangel:
@bendansby - looks like a spec failed, lmk if you need assistance fixing!
```
 Spec.Nri.Ui.PremiumCheckbox.V3
↓ Nri.Ui.PremiumCheckbox.V2
↓ premium
✗ appears with P+ flag when Premium pennant is passed in
```

brookeangel:
Screen Shot 2019-07-12 at 1 54 40 PM

😬 it looks like the other versions still have access to PremiumWithWriting, including V5 on the styleguide-app from netlify (code in this repo) and V2 used on Page.Teach.Assignment.Form.PeerReview.View.Criteria in the monolith.

I think the best thing to do with our versioning policy is to make a V6 as a copy of V5, and just remove the PremiumWithWriting flag from that, then upgrade the references in the monolith to V6 (looks like that's just in the form)

Does that seem like an alright thing to do? I'd be happy to pair

bendansby:
OK! I will set up some time.

PR #319, opened 2019-07-19
Tessa/more flexible button v2

tesk9:
Alternate version of https://github.com/NoRedInk/noredink-ui/pull/312.

This version of the Button module follows an attribute-setting pattern.

@avh4 @stoeffel @BrianHicks -- thoughts?


src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+success (ButtonOrLink config) =
+    ButtonOrLink { config | state = Success }
+
+
+
+-- INTERNALS
+
+
+{-| -}
+type alias Attribute msg =
```
avh4: 🎨 I think this should be a custom type instead of an alias, to keep this detail private.  Could be `type Attribute msg = Attribute (ButtonOrLink msg -> ButtonOrLink msg)` for now

src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+-- COLOR SCHEMES
+
+
+type ButtonStyle
+    = Primary
+    | Secondary
+    | Danger
+    | Premium
+      -- The following styles cannot be set directly.
+      -- Use a helper like `unfulfilled` to change the state instead.
```
avh4: 🎨 I think it would be clearer to have two distinct type, one for the 4 "styles", and then a separate one with all 8 "color schemes" if a type for the scheme is really even necessary

src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+        , url = "#"
+        , linkType = Default
+        , size = Medium
+        , style = Primary
+        , width = WidthUnbounded
+        , label = ""
+        , state = Enabled
+        , icon = Nothing
+        , customAttributes = []
+        }
```
avh4: 💭 I think the documentation somewhere should list what the defaults are (medium, primary, widthUnbounded, enabled)

src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+        { onClick = config.onClick
+        , url = url
+        , linkType = config.linkType
+        , size = config.size
+        , style = config.style
+        , width = config.width
+        , label = config.label
+        , state = config.state
+        , icon = config.icon
+        , customAttributes = config.customAttributes
```
avh4: ❓ 🎨  was there a reason not to use record update syntax here and in `onClick`?

src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+    ButtonOrLink { config | linkType = External }
+
+
+{-| Wrap some text so it looks like a button, but actually is wrapped in an anchor to some url and have it open to an external site
+
+This should only take in messages that result in tracking events. For buttons that trigger other effects on the page, please use Nri.Ui.Button.V2.button instead
+
+-}
+linkExternalWithTracking : Attribute msg
+linkExternalWithTracking (ButtonOrLink config) =
```
avh4: 🎨 for some of these extra link attributes, I'm not sure I understand what all are used for, but I think some for instance will only be used when there's also an href and/or a msg -- in cases like those, I think it'd be nice to have the attribute function handle multiple private options at once, like `linkExternalWithTracking : Url -> msg -> Attribute msg` (which sets onClick, href, and linkType all at once)

src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+
+
+{-| -}
+icon : Svg -> Attribute msg
+icon icon_ (ButtonOrLink config) =
+    ButtonOrLink { config | icon = Just icon_ }
+
+
+{-| -}
+custom : Html.Attribute msg -> Attribute msg
```
avh4: 🎨 I wonder if `custom` should just take a list of `Html.Attribute msg`, just to make the calling code a little less redundant in the case that there are multiple custom attributes?

src/Nri/Ui/Button/V9.elm
```diff
@@ -0,0 +1,887 @@
...
+
+
+{-| -}
+label : String -> Attribute msg
+label label_ (ButtonOrLink config) =
+    ButtonOrLink { config | label = label_ }
+
+
+{-| -}
+icon : Svg -> Attribute msg
```
avh4: 💭 I think we'd want to enforce that icon buttons always also have a label to use for alt text, so maybe icon should be `icon : String -> Svg -> Attribute msg` and set both the label and the icon at once?

src/Nri/Ui/Button/V9.elm
avh4: (or maybe the view code could just destructure the Style and the State at the same time and then having the extra "schemes" as custom type variants wouldn't even be necessary?)

src/Nri/Ui/Button/V9.elm
tesk9: uhhhhhhhhhhhhhhhh nope 😅

src/Nri/Ui/Button/V9.elm
tesk9: Yeah I thought about that and waffled! I can do multiple 👍

src/Nri/Ui/Button/V9.elm
tesk9: I think actually we don't want to label icon -- the icon's purpose is descriptive, and when that's the case the alt text should be explicitly set to `""`. 

That said, I don't have any idea about how to enforce that  here 🤔

src/Nri/Ui/Button/V9.elm
tesk9: I thought this was a great suggestion and I started it, but I ran into this:

```
linkWithMethod : String -> String -> Attribute msg
linkWithMethod method url (ButtonOrLink config) =
    ButtonOrLink { config | linkType = WithMethod method, url = url }
```

We don't have the Url type defined in noredink-ui. I wouldn't want the Url data type to be versioned/seems like a pain to add it.

I guess I could make this a record? What do you think?

src/Nri/Ui/Button/V9.elm
avh4: Oh, right, there's still text when there's an icon.  That makes sense!

Hmm, maybe label actually shouldn't be an optional Attribute, then, and should just be a required string parameter to the button functions?

src/Nri/Ui/Button/V9.elm
avh4: Hmm, yeah, I guess a record would be fine then.  Also I think just having two strings and documenting the order of them in the doc comment would also probably be fine 🤷‍♀

src/Nri/Ui/Button/V9.elm
tesk9: An idea that occurred to me is to let the button take html children, like a normal button:

```
Button.button 
    [  Button.primary
    ,  Button.onClick DoAThing
    ]
    [  Html.text "My label"
    ]
```

Or maybe:

```
Button.button 
    [  Button.primary
    ,  Button.onClick DoAThing
    ]
    [  Button.icon myIcon
    ,  Button.label "HELLO!!!"
    ]
```

tesk9: Yeah I think required string sounds good..

src/Nri/Ui/Button/V9.elm
tesk9: All done!

tesk9:
Alright! I think I've addressed all the feedback. I'll plan to merge this early next week, unless folks need or want more time to review. Thank you!!! 💚

src/Nri/Ui/Button/V9.elm
stoeffel: I agree with AV. I think this should be an opaque type.

stoeffel:
Looks great! Thanks for taking the time to make this so much nicer!

I agree with AV, that `Attribute` should be private.

src/Nri/Ui/Button/V9.elm
avh4: I'd be hesitant about having it take a list of children because I suspect we didn't really think through making the CSS handling arbitrary children -- things like wrapping, vertical alignment, containing inline vs inline-block vs block elements, dealing with things of unexpected height, etc would all need to be handled if we allowed more flexibility in specifying the children.

PR #320, opened 2019-07-23
add Nri.Ui.Text.V4 without headings

BrianHicks:
this removes headings from Nri.Ui.Text. No need, they're duplicate!

PR #321, opened 2019-07-23
Simplify headings API

BrianHicks:
## 🚧 Warning! 🚧

This is branched off of #320 instead of `master` and should change before merge.

## back to your regularly scheduled PR…

I liked the idea behind `Nri.Ui.Heading.V1`, but I think it may be a bit heavy for this use case. It's like 4 lines of code for each heading, and a bit repetitive/bothersome to type every time. I want to simplify the API here:

```elm
type Attribute msg

type Style
style : Style -> Attribute msg

css : List Css.Style -> Attribute msg

customAttr : Html.Styled.Attribute msg -> Attribute msg

h1 : List (Attribute msg) -> List (Html msg) -> Html msg
h2 : List (Attribute msg) -> List (Html msg) -> Html msg
h3 : List (Attribute msg) -> List (Html msg) -> Html msg
h4 : List (Attribute msg) -> List (Html msg) -> Html msg
h5 : List (Attribute msg) -> List (Html msg) -> Html msg
```

That means the calls go from:

```elm
heading [ Html.text "hey there" ]
  |> withVisualLevel Top
  |> withDocumentLevel H1
  |> view
```

to

```elm
Heading.h1 [] [ Html.text "hey there" ]
```

and if you need to customize them, they become

```elm
Heading.h3 [ style Subhead ] [ Html.text "subhead that uses an h3 tag" ]
```

way more compact. I even added customization options which we need for monolith code—this will allow us to get rid of some `Css.Global` calls! 😁


avh4:
I think it would be nice to add something in the doc comments in Text.V4 about what the Text.V3 heading functions should map to by default so it's easy to upgrade from Text.V3 to (Text.V4 + Heading.V2) -- maybe something like “Text.smallHeading "text" --> Heading.V2.xxx [] [ text "text" ]” or whatever the mapping actually is

PR #322, opened 2019-07-23
add Axe testing

BrianHicks:
This adds automatic accessibility testing to the styleguide. Right now we have 155 failures! So I've made it so that the test will pass if we have exactly that many, and fail if we go any higher or lower (so that we do not forget to lower the budget.)

### DIFF
--- a/script/format-axe-report.sh
+++ b/script/format-axe-report.sh
@@ -20,4 +20,12 @@ if test "$NUM_ERRORS" -ne "$TARGET_ERRORS"; then
     echo
     echo 'If it went down, hooray!'
     echo "Check out ${0:-} and change the count to the reported value above."
+    echo
+    echo "Did it go up? That's OK too."
+    echo "Since there are so many errors right now, a decent debugging strategy is:"
+    echo
+    echo " 1. save tests/axe-report.log somewhere ('mv tests/axe-report.log tests/axe-report.log.failing' is one way)"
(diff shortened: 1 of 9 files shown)


script/format-axe-report.sh
```diff
@@ -0,0 +1,31 @@
...
+# failures to the correct value.
+NUM_ERRORS="$(jq '.violations | map(.nodes | length) | add' "$JSON_FILE")"
+TARGET_ERRORS=155
+if test "$NUM_ERRORS" -ne "$TARGET_ERRORS"; then
+    echo "got $NUM_ERRORS errors, but expected $TARGET_ERRORS."
+    echo
+    echo 'If it went down, hooray!'
+    echo "Check out ${0:-} and change the count to the reported value above."
+    echo
+    echo "Did it go up? That's OK too."
```
tesk9: I read this as "feel free to increase the violations count." Is that what you mean?

tesk9:
(Also, I am EXTREMELY enthused about this! THANK YOU!)

BrianHicks:
oh! I didn't add it to the test target. That's what's going on. I tested by removing some low-hanging fruit (duplicate IDs in embedded SVGs) and it gave an error.

BrianHicks:
oh sorry, that is to say: give it another go, `make test` should do the right thing now.

script/format-axe-report.sh
BrianHicks: nope! reworded.

PR #332, opened 2019-07-30
Use avh4/elm-debug-controls

tesk9:
Use `avh4/elm-debug-controls` rather than `noredink/elm-debug-controls-without-datepicker`


tesk9:
(cc @rtfeldman because I know you're clearing out old NRI repos)

PR #336, opened 2019-08-05
Upgraded table versions to allow for additional cell styles

ckoster22:
To test, run the style guide locally and validate the new Table example. You should expect to see cells that turn red when the number is less than 5.

image


src/Nri/Ui/Table/V5.elm
```diff
@@ -0,0 +1,285 @@
+module Nri.Ui.Table.V5 exposing
+    ( Column, custom, string
+    , view, viewWithoutHeader
+    , viewLoading, viewLoadingWithoutHeader
+    )
+
+{-| Upgrading from V1:
```
brookeangel: Should this have instructions for upgrading from V4?

styleguide-app/Examples/Table.elm
```diff
@@ -33,12 +34,28 @@ example parentMessage state =
                 [ Table.string
                     { header = "First Name"
                     , value = .firstName
-                    , width = calc (pct 50) minus (px 125)
+                    , width = calc (pct 50) minus (px 250)
```
brookeangel: Don't forget to update the name of this module in the config above!

styleguide-app/Examples/Table.elm
```diff
@@ -33,12 +34,28 @@ example parentMessage state =
...
                     , value = .lastName
-                    , width = calc (pct 50) minus (px 125)
+                    , width = calc (pct 50) minus (px 250)
+                    , cellStyles = always []
+                    }
+                , Table.string
+                    { header = "# Submitted"
+                    , value = .submitted >> String.fromInt
+                    , width = px 125
+                    , cellStyles =
```
brookeangel: 😻 yay thanks for showing this in the styleguide!

src/Nri/Ui/Table/V5.elm
ckoster22: Nice catch. Added

Issue #343, opened 2019-08-12
Heading.V2 doesn't expose Style(..)


brookeangel:
Noticed while adding a heading that Style(..) wasn't exposed, so I couldn't actually customize the styles!

PR #344, opened 2019-08-12
expose heading styles

brookeangel:
For https://github.com/NoRedInk/noredink-ui/issues/343


BrianHicks:
thanks, I had this on my list but you beat me to it! 😁

brookeangel:
aw man, this fails CI because apparently exposing constructors constitutes a major version bump 😢 - do you think I should just make a v3 to avoid the major version bump conundrum?

BrianHicks:
I think it'd be fine to a major bump here. I don't see how this would cause breaking behavior in any application—it's not like we could have constructed or matched on a `Style` value before!

BrianHicks:
let's ask in Slack though.

brookeangel:
Thanks for asking!

PR #347, opened 2019-08-16
Tessa/modal scrolling

tesk9:
modal

### Top shadow
Screen Shot 2019-08-16 at 3 59 00 PM

### Bottom shadow
Screen Shot 2019-08-16 at 3 59 11 PM

cc @NoRedInk/design 

Adds modal shadows & adds support for autofocusing on any focusable element when the modal launches.

NOTE: this also removes ids from some svgs! This decreases the axe errors (and reduces the chances that we have multiple identical ids on the same page) but does mean there's some regression risk.

### DIFF
--- a/styleguide-app/Examples/Modal.elm
+++ b/styleguide-app/Examples/Modal.elm
@@ -15,7 +15,7 @@ import ModuleExample exposing (Category(..), ModuleExample)
 import Nri.Ui.Button.V9 as Button
 import Nri.Ui.Checkbox.V5 as Checkbox
 import Nri.Ui.Colors.V1 as Colors
-import Nri.Ui.Modal.V6 as Modal
+import Nri.Ui.Modal.V7 as Modal
 
 
 {-| -}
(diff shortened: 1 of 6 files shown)


bendansby:
Hero! One unrelated request: could we add baseFont to the modal title?

tesk9:
Screen Shot 2019-08-16 at 4 07 51 PM

styleguide-app/Examples/Modal.elm
```diff
@@ -252,6 +331,7 @@ type Msg
     | SetShowContinue Bool
     | SetShowSecondary Bool
     | SetDismissOnEscAndOverlayClick Bool
+    | SetLongContent Bool
```
allenap: Without a visible title the scroll hints conflict with the X:

image

I'm not sure that should block this PR though.

allenap: It can also overlap with the focus hint for the X when there is a title:

image

Again, I don't think this is a blocker.

tesk9:
Thanks for the review!!! I realized that I messed up the checkboxes a bit, so I'm going to separate out the removal of the ids from the modal scrolling work.

tesk9:
@BrianHicks can you help me with the netlify deploy failures? I tried to re-run the deploy but it doesn't seem to work

BrianHicks:
looks like it's an internal failure at Netlify. I'll retry it from the admin thingy.

BrianHicks:
it succeeded now!

tesk9:
Thank you!

PR #351, opened 2019-08-20
add Nri.Ui.Callout.V1

BrianHicks:
this adds Nri.Ui.Callout.V1

image

### DIFF
--- a/src/Nri/Ui/Callout/V1.elm
+++ b/src/Nri/Ui/Callout/V1.elm
@@ -25,6 +25,9 @@ import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Fonts.V1 as Fonts
 
 
+{-| Attributes of callouts. Use functions like [`sideText`](#sideText) and
+[`containerCss`](#containerCss) to construct these.
+-}
 type Attribute msg
     = SideText (Maybe (Html msg))
     | ContentCss (List Css.Style)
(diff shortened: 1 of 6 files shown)


BrianHicks:
@bendansby does this look OK? (tagging you since Stacey is out)

design is at https://app.zeplin.io/project/5cdde2fe4c7fea1d6280c5be/screen/5d4d98396986389b1e385236

one potential problem: the contrast ratio for the link is 4.4. ARIA guidelines say we need 4.5 at minimum. The current colors are azure:sunshine. Should we change that at all?

bendansby:
>one potential problem: the contrast ratio for the link is 4.4. ARIA guidelines say we need 4.5 at minimum. The current colors are azure:sunshine. Should we change that at all?

Yeah, we ran into that on Ink with Azure on Gray96. If we change Azure to #0A64FF, it should solve the problem.

bendansby:
Can you invite me to that Zeplin project?

BrianHicks:
you're added! I'll try the alternate color now.

bendansby:
What do the CSS customizations represent?

BrianHicks:
I've just had to override global styles too many times in Elm using `Css.Global` and the like. It always breaks unexpectedly because it's not right in the call site, so I've just started building in style customizations when I make new doodads.

In this particular case, this doodad ships with no external margins. They'll change across invocations, so they just get set with `containerCss`.

bendansby:
Looks good! If you're up for it you could go ahead and update the definition of Azure rather than use a one-off so we don't run into it again in the future.

BrianHicks:
if that's fine with you, I'll go ahead and do that. I'd rather go for a global solution, yknow?

bendansby:
Yup

src/Nri/Ui/Callout/V1.elm
```diff
@@ -0,0 +1,196 @@
...
+import Html.Styled.Attributes exposing (css)
+import Nri.Ui.Colors.V1 as Colors
+import Nri.Ui.Fonts.V1 as Fonts
+
+
+{-| Attributes of callouts. Use functions like [`label`](#label) and
+[`containerCss`](#containerCss) to construct these.
+-}
+type Attribute msg
+    = Label (Maybe (Html msg))
```
allymcknight: 🤔 ❤️

PR #360, opened 2019-09-05
constructor now reverses before

michaelglass:
missed a functionality change from list-zipper 3.1.1 to 4.0.0 (#359).  This includes it.


michaelglass:
nevermind. Because it's reversed in both places (in `from` and in `before`) this isn't a bug

PR #363, opened 2019-09-16
SegementedControl as toggle

glitteringkatie:
### Background
The district report needs a segmented control but in the mock (https://at0hwb.axshare.com/#id=sbgn8r&p=district_report click the `skills tab`) it is used as an input element as opposed to a page control so this adds a version of the segmented control view called `viewToggle` that takes all the same stuff except `content`.

Allows for progress on https://www.pivotaltracker.com/story/show/167167857

### 🖼 Style Guide screenshot
image
cc @bendansby making this change for the pathways toggle

### DIFF
--- a/src/Nri/Ui/SegmentedControl/V7.elm
+++ b/src/Nri/Ui/SegmentedControl/V7.elm
@@ -100,7 +100,17 @@ viewToggle config =
             , cursor pointer
             ]
         ]
-        (List.map (viewTab Nothing (Widget.selected True) config) config.options)
+        (List.map
+            (viewTab
+                { onClick = config.onClick
+                , selected = config.selected
+                , width = config.width
+                , selectedAttribute = Widget.selected True
+                , maybeToUrl = Nothing
+                }
+            )
+            config.options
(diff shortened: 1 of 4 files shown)


src/Nri/Ui/SegmentedControl/V7.elm
```diff
@@ -142,17 +175,23 @@ viewTab maybeToUrl config option =
...
+                Aria.currentPage
+
+            else
+                Widget.selected True
     in
     element
         (List.concat
             [ [ Attr.id idValue
               , Role.tab
-              , Aria.controls (panelIdFor option)
```
glitteringkatie: Turns out Aria.controls doesn't do much and we get errors when we have controls for an id that doesn't exist on the page (https://github.com/reach/reach-ui/issues/4). Additionally, we only show the id of the selected content (https://github.com/NoRedInk/noredink-ui/pull/363/files#diff-ae19bbc48c636566ee966b8343810172L100) so the other tab options throw axe errors. It was best to delete

Issue #367, opened 2019-09-23
Heading/Text view-only-this and view docs links are broken


tesk9:
See https://noredink-ui.netlify.com/#doodad/Nri.Ui.Text.V4%20(with%20headers%20from%20Nri.Ui.Heading.V2)

The title is used in the view-only-this feature and to construct the docs link. Including extra information in the title breaks these features.

PR #372, opened 2019-10-01
Fix broken links

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/367

Screen Shot 2019-09-30 at 6 15 30 PM


tesk9:
I really don't get what's going on with this branch and CI. I re-pushed as tessa/broken

PR #296, opened 2019-06-17
Highlighted style on Checkbox

tremlab:
Adding a highlighted style to the Checkbox module.  

Checkbox.V6
Premium. Checkbox.V6

image

(in reference to:
https://www.pivotaltracker.com/story/show/165900400)

for reviewer:  commit history may not be helpful.  I had made various changes to padding, but design preferred to revert those. (plus a few merge-to-master thrown in).

PR #379, opened 2019-10-09
Adds a new "Highlighted" checkbox in V6.

mthadley:
This adds a new `Highlighted` constructor to the `Theme` type in `Checkbox`. This new theme will be used for some new features on the Curriculum Library.

Here's what it looks like:

highlighted_checkbox

Since `PremiumCheckbox` actually depends on `Checkbox`, I decided to also create a new version of `PremiumCheckbox` that would be compatible with the newest version of `Checkbox`. I'm not sure if that was the right move, so please take a look at that change.

Thanks!


src/Nri/Ui/Checkbox/V6.elm
```diff
@@ -0,0 +1,709 @@
...
+    , disabled : Bool
+    , theme : Theme
+    }
+
+
+{-|
+
+    = Selected --  Checked (rendered with a checkmark)
+    | NotSelected -- Not Checked (rendered blank)
+    | PartiallySelected -- Indeterminate (rendered dash)
```
avh4: 🎨 4-space indent will render as a code block.

suggested:
```
- `Selected`: Checked (rendered with a checkmark)
- `NotSelected`: Not Checked (rendered blank)
- `PartiallySelected`: Indeterminate (rendered dash)
```

src/Nri/Ui/Checkbox/V6.elm
```diff
@@ -369,6 +373,100 @@ checkboxUnchecked =
         |> Icon
 
 
+checkboxCheckedHighlighted : Icon
+checkboxCheckedHighlighted =
```
avh4: ❓ why is the check itself a complete different icon?

src/Nri/Ui/Checkbox/V6.elm
mthadley: This was copied from the previous module version. I think they may have actually intended to for this to be rendered as a code block? It's like, sort of Elm code, after all. Either way, I think your suggestion is nicer to read.

src/Nri/Ui/Checkbox/V6.elm
mthadley: This was the asset I received from design. We could also style the existing checked icon based on the current theme.

mthadley:
> ❓ This new "theme" has completed different spacing, padding, and layout properties -- this seems like it's meant to be used in a completely different way than the normal checkbox; is that the intent from the design team? Also why does this highlighted one have a different icon from normal checkboxes -- should the normal checkboxes use the new icon?

Oh, I actually forgot to update the new checked icon to match the dimensions of the other icons. It doesn't look as out of place with that change.

As far as the intention of the new icon, we can ping @NoRedInk/design about it. Was it intended that the checkbox use the old "unchecked" style, and the have this new style when "checked"?

bendansby:
My intention was for us to just swap in a new graphic for the checked state so it has a blue border

src/Nri/Ui/Checkbox/V6.elm
mthadley: Also, I wonder if the only SVG assets we really need in this module are the lock and the checkmark. The rest of the checkbox could be CSS.

bendansby:
I'm wondering if this needs to be part of checkbox? It seems like the highlighting is more specific to the CL. Reminds me of this situation https://github.com/NoRedInk/noredink-ui/pull/296

mthadley:
> I'm wondering if this needs to be part of checkbox? It seems like the highlighting is more specific to the CL. Reminds me of this situation #296

Hah, yes! Very similar.

If we don't plan on having more checkboxes use this "style" in the future, than we should just implement this only on the CL.

I also originally thought I needed to make changes this module to actually get the styles to match, but it sounds like I can just add that blue border to the SVG and call it a day.

Unless there are any objecting thoughts, I can close this and go that route instead.

PR #384, opened 2019-10-11
bump to 7.5.0

brookeangel:
Releases Nri.Ui.Tooltip.V1


brookeangel:
cc: @allenap - not actually gonna ping you for a review because it's a minor version bump but fyi

PR #387, opened 2019-10-11
use onClickStopPropagation in tooltip.V1

brookeangel:
Fixes a bug where clicking a tooltip was triggering too many effects in https://github.com/NoRedInk/NoRedInk/pull/25320

The deprecated version of tooltip in the monolith also stops propagation.


bendansby:
We would only need to allow for click if we are not able to allow the user to mouse off the question mark onto the tooltip to click a link without the tooltip disappearing. If we can achieve that, hover only is fine.

brookeangel:
ah ok forgot we had some links in "?" tooltips - we aren't quite there in this version (this just added some accessibility stuff)

brookeangel:
@bendansby - was doing some research into "links on hover tooltips" and it's do-able but seems to not be accessibility best practice (it requires having the <button> that surrounds the trigger content also contain the content of the tooltip bubble, which would be confusing to screen readers).

Also, this module isn't in use anywhere yet so it can change, but I initially made all the "?"-triggered tooltips click-only based on https://inclusive-components.design/tooltips-toggletips/

Any ways you can think of around the accessibility snafu?

brookeangel:
Gonna merge this in on CI 🍏 and add hover in the next edit, since this just fixes a bug

PR #391, opened 2019-10-24
Use min and max widths

tesk9:
Screen Shot 2019-10-23 at 5 54 16 PM

Fixes https://noredink.slack.com/archives/C0VVDLEES/p1571876729020400

cc @NoRedInk/design


michaelglass:
after we merge this, can we get rid of min-width everywhere else?

tesk9:
Yeah for sure!!

PR #392, opened 2019-10-24
Tessa/accordion

tesk9:
Moves the accordion into noredink-ui.

I kept the API identical, and mostly this accordion is identical to the monolith version. 1 major difference is that the entry header is now for-real a button (before, it was a clickable div, but not a keyboardable div).

Screen Shot 2019-10-24 at 10 41 58 AM

cc @NoRedInk/design (No visual changes expected. Mostly pinging as a heads up that the accordion is moving to the https://noredink-ui.netlify.com)

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -6,6 +6,7 @@
     "version": "7.8.0",
     "exposed-modules": [
         "Nri.Ui",
+        "Nri.Ui.Accordion.V1",
         "Nri.Ui.Alert.V2",
         "Nri.Ui.Alert.V3",
         "Nri.Ui.Alert.V4",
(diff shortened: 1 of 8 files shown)


Arkham:
Hey @tesk9, the code looks good! The only thing I've noticed that is different from the version we have in the monolith is the outline. It looks particularly weird in the "peer review" version:

Screenshot 2019-10-28 at 16 15 09

If that's cool, then we can merge!

tesk9:
Thanks Ju! I think the focus outline should be fine, since it's more accessible and because of the work that Hadley did on focus rings in the monolith (https://github.com/NoRedInk/NoRedInk/pull/25233).

PR #393, opened 2019-10-24
Tessa/remove assets

tesk9:
While working on something else, I noticed that we were still passing an assets record to a few styleguide examples. This PR aims to fix that, so that we're only passing assets where it's unavoidable not to.

This required a new version of segmented control. As far as I can tell, SegmentedControl only uses Icons on the rate results page, so I focused on getting the segmented control example to look nice with Svg icons that correspond to the rate results view. 

Screen Shot 2019-10-24 at 12 29 25 PM

<details>
<summary>Rate results view, for reference
</summary>

Screen Shot 2019-10-24 at 11 03 11 AM

</details>

<details>
<summary>Styleguide showing icons prior to this change
</summary>

Screen Shot 2019-10-24 at 11 03 06 AM

</details>

I also had to convert some old icons to elm-svg to get this to work. However, it kinda seems like the svg code I exported from zeplin wasn't quite right. I had to futz with the feColorMatrix for the "star" icon to get it to look darker on the edges, and then the empty star seemed at first like the viewbox was wrong, but I think actually the path is getting cut off. @NoRedInk/design are there other icons I should use? 

Screen Shot 2019-10-24 at 12 42 20 PM

### DIFF
--- a/src/Nri/Ui/Svg/V1.elm
+++ b/src/Nri/Ui/Svg/V1.elm
@@ -1,22 +1,31 @@
 module Nri.Ui.Svg.V1 exposing
     ( Svg
+    , withColor
     , fromHtml, toHtml
     )
 
 {-|
 
 @docs Svg
+@docs withColor
 @docs fromHtml, toHtml
 
 -}
 
+import Css exposing (Color)
 import Html.Styled as Html exposing (Html)
-import Svg
+import Html.Styled.Attributes as Attributes
 
 
 {-| -}
 type Svg
     = Svg (Html Never)
+
+
+{-| -}
+withColor : Color -> Svg -> Svg
+withColor color (Svg svg) =
+    Svg (Html.span [ Attributes.css [ Css.color color ] ] [ svg ])
 
 
 {-| Tag html as being an svg.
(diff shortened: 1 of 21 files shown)


bendansby:
Yay for removing PNGs! I'll defer to @staceyadams on the details

src/Nri/Ui/Svg/V1.elm
```diff
@@ -1,24 +1,33 @@
 module Nri.Ui.Svg.V1 exposing
     ( Svg
+    , withColor
```
michaelglass: 😍

styleguide-app/Examples/Button.elm
```diff
@@ -108,26 +107,13 @@ init assets =
...
-        , ( "Just Lock"
-          , Icon.lock assets
-                |> Icon.decorativeIcon
-                |> NriSvg.fromHtml
-                |> Just
-                |> Control.value
-          )
+        [ ( "Nothing", Control.value Nothing )
+        , ( "Just Performance", Control.value (Just UiIcon.performance) )
+        , ( "Just Share", Control.value (Just UiIcon.share) )
```
michaelglass: 🤘

styleguide-app/Examples/UiIcon.elm
```diff
@@ -58,5 +58,10 @@ example =
             , ( "attention", UiIcon.attention )
             , ( "exclamation", UiIcon.exclamation )
             ]
+        , (IconExamples.view "Notifs" << List.map (Tuple.mapSecond Svg.toHtml))
```
michaelglass: 👍

PR #394, opened 2019-10-24
Add Nri.Ui.Pennant.V1

brookeangel:
Part of https://www.pivotaltracker.com/story/show/169182354

Integration plan:
- [ ] Use in https://github.com/NoRedInk/NoRedInk/pull/25423 (fixes pennant sizing so it is the same as with Premium Checkbox)
- [ ] Deprecate Nri.Pennant in monolith

Screen Shot 2019-10-24 at 2 29 41 PM

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -58,6 +58,7 @@
         "Nri.Ui.Page.V2",
         "Nri.Ui.Page.V3",
         "Nri.Ui.Palette.V1",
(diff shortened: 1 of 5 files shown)


brookeangel:
The failure says I removed Modal.V8? Is there some weird stuff happening on Master right now?

PR #398, opened 2019-10-28
update nixpkgs and switch to niv

BrianHicks:
This PR switches us to `niv` to manage nix dependencies. It also adds docs!


BrianHicks:
@Arkham hey Ju, please LMK if you don't use Nix for this and I will find someone else to review. Teambot just suggested you 😄

Arkham:
I'll try to set it up, can't think of a better test :)

BrianHicks:
good idea ❤️

BrianHicks:
It’s because nix is still not the only way to do it. Thanks for reviewing!

PR #408, opened 2019-11-15
bump to 7.12.1

brookeangel:
releases a patch on Tooltip.V1 - stops click propagation


brookeangel:
Looks like one of the percy failures is a flake, & the other percy failure was because it flaked the previous time, so gonna merge this!

PR #409, opened 2019-11-15
Fix a11y low hanging fruit

BrianHicks:
this gets rid of ~26~ 31 easy-ish AXE violations. There are a few left but clearing these out revealed which are hard and which are easy. (This was 50% of our Axe violations BTW!)

remaining issues after this:

- controls from avh4/elm-debug-controls are not labeled. I submitted a PR but goofed and made invalid HTML… probably a way to get around this but it turned into non-low-hanging-fruit!
- some SVG elements have duplicate IDs for filters. Unfortunately I don't see a way around this if we're going to embed those directly instead of referencing them ☹️
- ~some elements do not have sufficient color contrast. I'm gonna get together a list and ask Design about them.~ We're darkening the default red to meet AA levels (4.5:1) 💪 
- the style guide doesn't have skip links.

This also makes the axe report much easier to scan by relying less on 2-space indentation for hierarchy.


BrianHicks:
@mncharlton it looks like this is failing to click a modal link. Any ideas what could be happening?

BrianHicks:
Darkened default red to #E70D4F per @bendansby. Thanks for the flexibility. ❤️

BrianHicks:
that GIF 😻

PR #411, opened 2019-11-18
add an input type that handles floats

glitteringkatie:
Adds the ability to handle text-input floats so we can input averages in the admin page https://www.pivotaltracker.com/story/show/169453952

To review
- [x] checkout this branch locally
- [x] run `script/develop.sh`
- [x] hit http://localhost:8000/#doodad/Nri.Ui.TextInput.V5
- [x] Play around with the various inputs on the page and confirm that the "Points" field only takes integers and the "Points (decimal)" field accepts floats

### DIFF
--- a/src/Nri/Ui/TextInput/V5.elm
+++ b/src/Nri/Ui/TextInput/V5.elm
@@ -2,7 +2,7 @@ module Nri.Ui.TextInput.V5 exposing
     ( Model
     , view, writing
     , generateId
-    , number, text, password
+    , number, float, text, password
     )
 
 {-|
@@ -19,7 +19,7 @@ module Nri.Ui.TextInput.V5 exposing
 
 ## Input types
 
-@docs number, text, password
+@docs number, float, text, password
 
 -}
 
@@ -68,13 +68,24 @@ text =
         }
 
 
-{-| An input that allows number entry
+{-| An input that allows integer entry
 -}
 number : InputType (Maybe Int)
 number =
     InputType
         { toString = Maybe.map String.fromInt >> Maybe.withDefault ""
         , fromString = String.toInt
+        , fieldType = "number"
+        }
+
+
+{-| An input that allows float entry
+-}
+float : InputType (Maybe Float)
+float =
+    InputType
+        { toString = Maybe.map String.fromFloat >> Maybe.withDefault ""
+        , fromString = String.toFloat
         , fieldType = "number"
         }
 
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/TextInput/V5.elm
```diff
@@ -68,7 +68,7 @@ text =
         }
 
 
-{-| An input that allows number entry
+{-| An input that allows integer entry
 -}
 number : InputType (Maybe Int)
```
glitteringkatie: Right now this PR is not a breaking change since it only adds functionality. We could make a breaking change if we feel weird about there being a "number" input separate from a "float" input. @ckoster22 and I talked about this and saw two solutions--either change the name "number" to "integer" or create a custom Number type that encompasses both Int and Float and does some logic as far as Stringify functions go. Does anyone feel strongly about doing this a certain way?

brookeangel: seems fine to me ~ maybe in v6 we should make this take a `number` as it's parameter and reconsolidate? Isn't `number` an encapsulating type in elm already?

src/Nri/Ui/TextInput/V5.elm
```diff
@@ -110,39 +121,48 @@ view_ theme model =
 
         (InputType inputType) =
             model.type_
+
+        maybeStep =
+            if inputType.fieldType == "number" then
+                [ step "any" ]
```
BrianHicks: "any" in the spinner control looks like it might default to "1", same as not setting it. Why do we need to? Alternatively, why aren't we allowing control of the value here?

glitteringkatie: What's the spinner control? The presence of the "step" property is what allows for input of a float value without the text input becoming invalid.  "any" seemed to not mess up ints, not error on floats, and be good enough. We could allow control of the value! But didn't have a current need--not sure how we like to balance future proofing vs doing the least invasive change.

BrianHicks: spinner control is the arrow keys working to increase/decrease the value and the arrows here:

image

if that's not the goal here this seems fine!

glitteringkatie: Oh cool! New terminology for me 👏

brookeangel: TIL also!

PR #412, opened 2019-11-19
Fix percy flakes

mncharlton:
Percy's been flaking a bit since it was introduced, most likely it's that it's firing too early before pages have loaded (it appeared to be capturing the same page twice or not capturing modals), this adds in a load of waits so that the right pages is definitely there before Percy tries to capture it (🤞). It does the same for the modals, making sure they're open before capturing.

For the main pages, I added a Class - ID would have been better but my Elm knowledge ran out at trying to remove the spaces in the string I was using and IDs aren't supposed to have spaces in (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id), using a Class means I can just use the first word as the thing to check - feel free to change if there is a better way of doing this.

For the modals I used the `role="dialog"` attribute they have.

This seems to now show the correct things in Percy - the three diffs are ones where it had flaked in the past and this branch is showing the correct snapshot.

### DIFF
--- a/script/percy-tests.js
+++ b/script/percy-tests.js
@@ -1,36 +1,32 @@
 const PercyScript = require('@percy/script')
 
 PercyScript.run(async (page, percySnapshot) => {
-  await page.goto('http://localhost:8000/')
-  // ensure the page has loaded before capturing a snapshot
-  await page.waitFor('#categories')
-
   await page.goto('http://localhost:8000/#category/Animations')
-  await page.waitFor('.Animations')
+  await page.waitFor('#animations')
   await percySnapshot('Animations')
 
   await page.goto('http://localhost:8000/#category/Buttons')
-  await page.waitFor('.Buttons')
+  await page.waitFor('#buttons-and-links')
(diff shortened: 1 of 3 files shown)


BrianHicks:
🤔 we shouldn't have spaces in our IDs. Which one was giving you trouble?

mncharlton:
@BrianHicks, it was when I tried to add an ID, because I used the category name some of them have spaces (eg 'Buttons and Links') and I didn't know how to get rid of them in Elm. On line 61 of view.elm I'd initially made that an ID, but swapped to a class because of the spaces.

As an aside we do have spaces in IDs in other bits of the code, for example the tabs on /admin/mock_data and the date picker on the assignment form step 3), but that's totally unrelated to this! 😆

BrianHicks:
could we pair on this sometime tomorrow? I'd rather we do this with IDs since it'll be less flakey long-term. I'll send you a calendar invite, but feel free to decline 😄

BrianHicks:
rebased, will merge on green.

PR #415, opened 2019-11-22
Set `rel` on external links

mthadley:
Adds `rel="noopener noreferrer` on buttons that use `Button.linkExternal` and `Button.linkExternalWithTracking`.

This is a reasonable default to set for any link using `target="_blank"`. It prevent security exploits like ["tabnabbing"](https://en.wikipedia.org/wiki/Tabnabbing), and also can improve performance. For more info, check out these links:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Security_and_privacy_concerns
* https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

Its possible that this could break fancy cross-window communication that we had in one of our apps, but I'm pretty sure we don't do anything like that. I checked both the rails monolith and Haskell monorepo for references to `window.opener` or `window.postMessage`, but didn't find anything.

### DIFF
--- a/src/AttributeExtras.elm
+++ b/src/AttributeExtras.elm
@@ -1,0 +1,20 @@
+module AttributeExtras exposing (targetBlank)
+
+import Html.Styled exposing (Attribute)
+import Html.Styled.Attributes as Attributes
+
+
+{-| Use this list of attributes instead of applying `Attributes.target "_blank"`
+directly. This prevents an exploits like "tabnabbing", among other things.
+
+See these resources:
+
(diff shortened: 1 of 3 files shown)


avh4:
How can this "improve performance" ?

mthadley:
> Linking to another page with target="_blank" will run the new page in the same process as your page. If the new page executes JavaScript, your page's performance may suffer. Avoid this with rel="noopener". (Can be combined with the above as rel="noopener noreferrer".)

avh4:
Ah, cool

mthadley:
No urgency on this, so I'll just let it get bundled with the next release.

PR #417, opened 2019-11-25
Decrease heading size

tesk9:
Often I accidentally click the wrong link (between the docs and the just-this links), and the headers take up a lot of vertical real estate. Trying out making them slimmer! feel free to tell me this is worse I won't be crushed

Screen Shot 2019-11-25 at 2 51 53 PM
Screen Shot 2019-11-25 at 2 51 41 PM

cc @NoRedInk/design


staceyadams:
I like the slimness!

PR #419, opened 2019-11-26
Tessa/add width label helpers

tesk9:
Add helpers to Svg for convenient label-adding and width-setting.
Add a view to the styleguide for seeing how to work with Nri.Ui.Svg "svgs"

Screen Shot 2019-11-25 at 6 19 08 PM


BrianHicks:
Looks like we're failing due an an a11y violation. The additional failure is:

```
scrollable-region-focusable: moderate violation with 1 instances.

Ensure that scrollable region has keyboard access (https://dequeuniversity.com/rules/axe/3.3/scrollable-region-focusable?application=axeAPI)

  at pre:

    Fix any of the following:
      Element should have focusable content
      Element should be focusable
```

src/Nri/Ui/Svg/V1.elm
```diff
@@ -1,42 +1,109 @@
...
+{-| -}
+withWidth : Css.Px -> Svg -> Svg
+withWidth width (Svg record) =
+    Svg { record | width = Just width }
+
+
+{-| -}
+withHeight : Css.Px -> Svg -> Svg
+withHeight height (Svg record) =
+    Svg { record | height = Just height }
```
BrianHicks: it looks like these set max width and max height according to the aspect ratio of the icon itself—that is, they will not necessarily set the exact width/height. Maybe `setWidthBound` or `setMaxWidth` or something else that indicates exactly what's going on?

Tempting to do…

```elm
type Bounds
    = MaxWidth Int
    | MaxHeight Int
    | MaxRect { width : Int, height : Int }

withBounds : Bounds -> Svg -> Svg
```

src/Nri/Ui/Svg/V1.elm
```diff
@@ -1,42 +1,109 @@
...
+        ( [], Just label ) ->
+            Html.div
+                [ Widget.label label
+                , Attributes.css [ Css.display Css.inlineBlock ]
+                ]
+                [ Html.map never record.icon
+                ]
+
+        ( [], Nothing ) ->
+            Html.map never record.icon
```
BrianHicks: I'm skeptical of the repetition here. Would it be possible to repeat less? Or share values?

src/Nri/Ui/Svg/V1.elm
tesk9: Our SVGs should always have `Attributes.width "100%", Attributes.height "100%"` as properties, so yeah, this property won't usually change the aspect ratio. I didn't want to do `maxHeight` or `maxWidth` because it might get confusing if you were actually trying to set the max-height or max-width CSS properties.

tesk9: Although, there's nothing enforcing the `Attributes.width "100%", Attributes.height "100%"` properties apart from convention -- we could end up with an Svg element that behaved different

tesk9:
@BrianHicks what does "jaggy" mean?

BrianHicks:
It did not update smoothly for me. Like a low frame rate for animation

tesk9:
Hmm, weird!

BrianHicks:
Possibly just Safari. If it’s fine for you not a blocker here

tesk9:
If we ignore the slight jitter on safari, do you think this is mergeable?

BrianHicks:
Yeah, I think it is!

Sorry for the delay; your response got buried under Black Friday emails 🙄

PR #423, opened 2019-11-27
Make modal ignore horizontal overflow

Arkham:
## What's up

We need to ignore horizontal overflow in modals :)

This will fix https://www.pivotaltracker.com/story/show/169919487


Arkham:
@tesk9 do you think this is fine or should we make a new modals version?

BrianHicks:
I think all our modal content should only scroll vertically so this seems fine. Gonna merge it.

PR #427, opened 2019-12-04
add ratchet script

BrianHicks:
this adds a ratcheting script similar to what we have on the monolith.

The basic idea: we want to use fewer old versions of our widgets over time, not more. This script will enforce that and give friendly error messages when things go wrong!


BrianHicks:
@tesk9 what additional modules do you think we need to deprecate here? HTML again?

tesk9:
I think everything in `tesk9/accessible-html` too -- Accessibility, Accessibility.Role, etc.  We don't have that many dependencies, at this point, luckily

BrianHicks:
will do in a second PR!

PR #429, opened 2019-12-09
fix text alignment and propagation when links are clicked

brookeangel:
Fixes a few bugs discovered in https://github.com/NoRedInk/NoRedInk/pull/25670
- text is now left aligned in tooltips
- clicking links in hover toggle tips will not propagate events


brookeangel:
oops 😅 i guess that is a slightly confusing link - out of laziness not gonna change it right now but i'll keep that in mind next time

PR #438, opened 2020-01-07
Tessa/add svg button

tesk9:
Adds Nri.Ui.ClickableSvg.V1, which helps us conveniently create a clickable (button or anchor tag) svg.

Screen Shot 2020-03-09 at 11 09 35 AM

cc @NoRedInk/design Note: on hover, the color remains azure. Is that desired for icon buttons?

Reviewer: I recommend looking at the files changed rather than going commit by commit. The API changed mid-PR.

<details>
<summary> API discussion starting point -- out of date at this point </summary>

Hello!

@tremlab and I were pairing the other day and needed to add a button that was just an icon. It was inconvenient! Thought it made sense to start talking about how to create this kind of thing.

Questions:
- should Nri.Menu icons be generalized instead?
- should this live in Nri.Ui.Button or Nri.UI.ClickableText instead?
- should interact-able icons like these always have tooltips?
- do we need to set the line-height? (wondering if this will be hard to position...)
- are there other oversets that we need to do?
- should the color change on hover..?

Please review in-context by running `make styleguide-app` and navigating to http://localhost:8000/#doodad/Nri.Ui.Svg.V1.

</summary>


mthadley:
Saw your message in slack, and had a few thoughts about  some of your questions:

> - should this live in Nri.Ui.Button or Nri.UI.ClickableText instead?

That was my first thought when I saw this. I think it might be easier if we have fewer modules to check when trying to figure out how to create a button.

> - should interact-able icons like these always have tooltips?

Yes! Even apart from the normal accessiblity concerns, I've experienced many user interfaces where "icon buttons" don't have some kind of `title=` or tooltip, and it's not obvious what it does. I just end up pressing it to find out, with varying outcomes of my being happy about what it _actually_ did.

tesk9:
Thanks for sharing your thoughts, Michael!

On your first answer, living in Nri.Ui.Button or Nri.Ui.ClickableText... Do you have API suggestions? I'm not sure how to integrate this nicely. In both modules, there's already something called `icon`... Maybe there should be another attribute that hides the button text behind a tooltip?

```
import Nri.Ui.ClickableText.V1 as ClickableText
import Nri.Ui.UiIcon.V1 as UiIcon

...
    ClickableText.button "Edit"
       [ ClickableText.small
       , ClickableText.iconAndTooltip UiIcon.edit
       ]
```

Something like this maybe...?

mthadley:
I'm not sure I would expect it to be in `Ui.ClickableText`. Would it actually render any clickable text?

That leaves `Ui.Button`, and I see what you mean about difficulty in integrating it into that existing API. Maybe `Ui.Svg` is actually the best place right now, then?

tesk9:
I agree with you though that Nri.Ui.Svg wouldn't be where I would expect to see a clickable icon! It does seem like.. separate.

How would you feel about adding `Nri.Ui.ClickableIcon.V1`? or something like that?

mthadley:
> How would you feel about adding `Nri.Ui.ClickableIcon.V1`? or something like that?

I think that makes sense. It follows the pattern of `ClickableX` where `X` is something that we don't always expect to be clickable, like a button. Sometimes icons are used to only convey information, like text.

stoeffel:
updated against master. Is this still in use or can we close this? @tesk9

jwoudenberg:

tesk9:
Still in use, thanks!

staceyadams:
Sorry, I totally missed this from January 😱 

`azureDark` for hover state if it's not too late.

tesk9:
I don't think I pinged you until yesterday, Stacey! I started this a long time ago, but wasn't finished with it yet :) I'll add azureDark for hover -- thanks!

tesk9:
with hover

PR #439, opened 2020-01-09
Tessa/add highlight icons

tesk9:
Playing with this idea https://noredink.slack.com/archives/C7CQZM7CL/p1578603481030600?thread_ts=1578597202.025300&cid=C7CQZM7CL for making highlighting interface clearer.

adds highlight and eraser icons

Initial impressions: 
- for a given icon, we'd want to be really careful about the "point" of the icon (like, this highlight icon the bit that you'd aim the cursor with is the tip of the pen, but when you use the eraser, you'd aim it with the bottom edge.) if we want to pursue this idea, we'd need to adjust the positioning a tad
- I think these icons may have too much whitespace?
- they look less crisp than I was expecting

cc @NoRedInk/design 

Incidentally, this PR also adds the highlight and eraser icons to noredink-ui.


staceyadams:
Sorry I haven't responded to this yet! I'll try to look in more detail before the end of the week.

* I think the eraser icon lines are definitely too thin/too much whitespace
* Also the idea of erasing a highlighter doesn't make sense. I think this is another one of those things I put into a design as a placeholder and it stuck 😅

staceyadams:
Compared to how simple and crisp browser cursors are, I feel like the line at the bottom of the highlighter icon (the highlight) gets in the way. And the icon itself doesn't read that well at the cursor size.

Would the highlighter cursor color change with the highlight color or would the cursor color always be the same?

tesk9:
Yeah they're not crisp! I'm not super happy with them. But they could change color depending on the highlight

stoeffel:
Updated against master.

Is this still in use?

jwoudenberg:

PR #442, opened 2020-01-21
No button margin

bendansby:
Zeroes out all margin on buttons to avoid misalignment problems like this on Safari.

Napkin 2 01-21-20, 11 51 03 AM

### DIFF
--- a/src/Nri/Ui/Button/V9.elm
+++ b/src/Nri/Ui/Button/V9.elm
@@ -720,7 +720,7 @@ buttonStyle =
         , Css.property "transition" "background-color 0.2s, color 0.2s, box-shadow 0.2s, border 0.2s, border-width 0s"
         , Css.boxShadow Css.none
         , Css.border Css.zero
-        , Css.marginBottom Css.zero
+        , Css.margin Css.zero
         , Css.hover [ Css.textDecoration Css.none ]
         , Css.disabled [ Css.cursor Css.notAllowed ]
         , Css.display Css.inlineFlex


bendansby:
Hmm, we definitely want to preserve any intentional margin, no matter whether it's top, bottom, left, or right. It's odd that the "base" styling would override the additional styling applied to the button instance used in the modal example. Just a matter of specificity, I guess. I suppose the best course would be a new version?

tesk9:
It seems like it works fine if you set `marginTop`, `marginRight` to zero one by one. Which doesn't make sense to me. I'm looking into it to see what's going on

Screen Shot 2020-01-22 at 10 45 19 AM

tesk9:
Seems like:
1. Class names are determined based on the styles applied (Richard says "The generated class name is a pure function of the Style values involved, and the output is determined by iterating through the virtual dom nodes, concatenating all their styles, and emitting that to a <style> element.")
2. Classes are added in a consistent order, _not_ in the order that you add `css` attributes
Initial investigation with margins: https://ellie-app.com/7QfYbCgVhfWa1
Clearer example with background colors: https://ellie-app.com/7Qg3R8Q74rja1 (the background ends up being green, even though you would expect for it to be blue or red)

Related to https://github.com/rtfeldman/elm-css/issues/488 and https://github.com/rtfeldman/elm-css/issues/415, I expect

All of which is to say... for components that accept custom attributes, I think we _must_ create a new version of the component whenever we change styles in this library :/. Or commit to double-checking that there aren't regressions, but I don't think that's feasible for Button cause it's used everywhere.

Alternately: we could avoid using the `css` helper, and add our own `style` nodes.

Or we could talk to Richard about it...

bendansby:
>Or we could talk to Richard about it...

I think I might say something I'd regret later 😆 

I'll make a new version. Thanks for such a thorough investigation!

PR #443, opened 2020-01-21
remove x bottom padding

bendansby:
The close x on the modal is clipped in Safari. This fixes it.
Untitled 2020-01-21 14_53_06


bendansby:
The focus looks good in Safari but not Chrome! Given the choice between crappy x in Safari or crappy focus in Chrome, I'll go with the latter 😆

PR #441, opened 2020-01-16
Lab/customizable select

tesk9:
For https://www.pivotaltracker.com/story/show/170709954

Adds Nri.Ui.Select.V7, which customizes the arrows used to indicate that this element is a dropdown. Replaces https://github.com/NoRedInk/noredink-ui/pull/119.

This PR also adjusts the border (adds error color, focus color, and increases bottom width) per "Inputs" guidelines in the zeplin styleguide.

Screen Shot 2020-01-16 at 1 16 06 PM

<details>
<summary>Kinda hard to see given focus outline, but the border is blue on focus.
</summary>

Screen Shot 2020-01-16 at 1 16 14 PM

</details>

cc @NoRedInk/design

PR #445, opened 2020-01-22
bump to 7.18.0

BrianHicks:
bumps to 7.18.0, which adds Nri.Ui.Select.V7 from #441


BrianHicks:
oh look I'm leaving myself a gif what a maverick

PR #446, opened 2020-01-27
Fix overflow in selectors

BrianHicks:
Selectors currently look a little funny if the text would be long enough to go under the arrows.

This fixes selectors to truncate instead:

image

### DIFF
--- a/src/Nri/Ui/Select/V7.elm
+++ b/src/Nri/Ui/Select/V7.elm
@@ -103,8 +103,9 @@ view config =
 
             -- Size and spacing
             , Css.height (Css.px 45)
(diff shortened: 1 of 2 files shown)


BrianHicks:
@NoRedInk/design what do you think?

BrianHicks:
note: Chrome does not respect `text-overflow: ellipsis` in select boxes but Safari and Firefox both do. It's a huge pain to get Chrome to respect the setting, requiring some dubious DOM changes to get working. I don't think it warrants the effort in this case.

bendansby:
Looks good! Thanks for the fix.

PR #448, opened 2020-01-29
Dansby/button v10

bendansby:
Adds v10 of the button, which zeroes out margin.


tesk9:
I'm wondering if we should change the API a bit here in order to try to prevent style collisions that don't work out predictably. Maybe we should add a `css` helper and try to only add elm-css styles once.. I suppose we could also have `custom` only take Html.Attributes instead of Html.Styled.Attributes to keep users from doing `Button.custom [ css [] ]`, since apparently it's not safe to do. That latter suggestion sounds inconvenient... but it would be safer.

bendansby:
I'm totally out of my depth when it comes to those ideas, so I fully defer to you!

bendansby:
Any conclusions @tesk9 ?

tesk9:
I made some API changes -- added `css` to a new version of ClickableText and to Button.V10. Thinking more on it, we could also change ClickableText.V3 in place... I'm going to request a review from another developer and see where we end up!

terezka:
Oh, you already saw that!

tesk9:
@terezka thanks for reviewing! I'll change it to be a minor update -- that'll be easier to deal with in the monolith for sure

PR #450, opened 2020-02-03
New checkbox API

terezka:
Additionally to revamping the api, this adds opacity to disabled label, as the designers style guide [prescribes](https://app.zeplin.io/project/5973fb495395bdc871ebb055/screen/597e878eac1fe12b4829df30
). 

>
>


src/Nri/Ui/PremiumCheckbox/V6.elm
```diff
@@ -32,7 +32,7 @@ import Nri.Ui.Pennant.V1 exposing (premiumFlag)
 view :
     { label : String
     , id : String
-    , selected : Checkbox.IsSelected
```
terezka: FYI: I checked in our code (monolith and monorepo) and we never use this name directly, so it will not require an update.

src/Nri/Ui/Checkbox/V6.elm
```diff
@@ -0,0 +1,640 @@
...
+    - `onChange` is the handler called when the state changes.
+
+-}
+checkbox :
+    { identifier : String
+    , label : String
+    , anonymous : Bool
+    , selected : Selected
+    , onChange : Bool -> msg
+    }
```
avh4: Should we switch to the new style of API of having `checkbox : List Checkbox.Attribute -> Html` ?   I guess label, selected, onChange should always be required,   but I think identifier could be option (generate from the label if not given),  anonymous (hidden label) is only needed occasionally,  and disabled/locked could make sense as an attribute instead of as a separate function.   Switching to attributes would also make it nicer for selected, since we could have both `selected : Bool -> Attribute` and `maybeSelected : Selected -> Attribute`

terezka: I have a few concerns with that approach:
- You loose the ability to prevent faulty configurations.
- If one makes a faulty configuration, it won't necessarily be obvious what the result is.
- You can set the same attribute twice, overriding the first without knowing.
- More lookups in documentation to see what attribute to use.

Personally, I prefer to just be explicit about the API, but if NRI already took a stand on the issue, I'm happy to oblige.

jwoudenberg:

PR #451, opened 2020-02-04
Icon center

bendansby:
A tweak to outfox https://www.pivotaltracker.com/story/show/171053637


tesk9:
@bendansby I'm a bit nervous about changing Nri.Ui.Icon.V5.. Also, since it's a deprecated component, I'd prefer to fix issues by updating uses to the UiIcon and Svg. I'm making a separate PR doing so

bendansby:
OK! Thanks for assistance

PR #452, opened 2020-02-12
Clickable text underline

bendansby:
Changes the clickable text hover style to be a bottom border rather than text-decoration

image


bendansby:
>Why do we want to use a border instead of text decoration

Because it looks better! (to me :-)

bendansby:
>Could you please add this change to the "post-release patch changes" section at the top of the doc?

Done!

bendansby:
>I'm going to see if I can get a PR through to the monolith that'll protect us before this merges

Thank you!

tesk9:
Made a PR here: https://github.com/NoRedInk/NoRedInk/pull/26334

PR #456, opened 2020-02-14
Adds google classroom icon

tesk9:
Screen Shot 2020-02-14 at 2 45 40 PM

cc @NoRedInk/design Is this still what we want the google classroom icon to look like? (I wonder if there's a single-color alternative?)

### DIFF
--- a/src/Nri/Ui/Logo/V1.elm
+++ b/src/Nri/Ui/Logo/V1.elm
@@ -1,12 +1,12 @@
 module Nri.Ui.Logo.V1 exposing
     ( noredink
-    , facebook, twitter, clever
+    , facebook, twitter, clever, googleClassroom
     )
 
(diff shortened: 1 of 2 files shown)


bendansby:
Looks like they offer a single color version here https://developers.google.com/classroom/brand

tesk9:
Hmm it also says that they care about all the colors in the black and white and gray logo 🤔now i'm not sure what to do 😅

Arkham:
Oh, I merged this because I thought it looked great, only realized now that we were still discussing if we wanted the colorized version or the greyscale version. My bad! If you are planning to follow up and create a new pr for the greyscale version, assign me to it!

tesk9:
No thanks for merging! I'm happy with this as our next step!

PR #458, opened 2020-02-17
Text links

bendansby:
Improves the styling of links in text by adding an underline and making them always blue.

image

### DIFF
--- a/src/Nri/Ui/Text/V4.elm
+++ b/src/Nri/Ui/Text/V4.elm
@@ -54,6 +54,7 @@ API. See the Nri.Ui.Heading.V2 docs for details.
 -}
 
 import Css exposing (..)
+import Css.Global exposing (a, descendants)
 import Html.Styled exposing (..)
 import Html.Styled.Attributes exposing (css)
 import Nri.Ui.Colors.V1 exposing (..)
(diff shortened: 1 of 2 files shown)


tesk9:
Screen Shot 2020-02-18 at 10 51 15 AM

I moved the examples around a bit so that links would always show for non-user generated text. (I don't think there's a way for users to generate an actual link, unless we're parsing markdown somewhere? Which might make sense for non rich text prompts 🤔...)

bendansby:
Thank you!! That strikes this Hack Day item off the list: 
image


Re: ug, maybe relevant? https://noredink.slack.com/archives/CT3E6DFS5/p1581611224016500

tesk9:
Screen Shot 2020-02-18 at 11 18 03 AM

I looked into it. I believe that all the writing assignment prompts are rich text, which means that they're not rendered with the text component.  The rich text read only styles claim to be duplicated from UG text (https://github.com/NoRedInk/NoRedInk/blob/master/monolith/ui/src/Nri/Writing/RichText/Styles.elm#L72). I'm not sure to what degree it's worth improving RichText, since we're moving towards the Quill implementation (although I doubt we'll have time to do anything with prompts for a long time).

I see 1 use of ugMediumBody in the monolith (in Page/Learn/GuidedDrafts/ExpandingCard.elm). 

We seem to use ugSmallBody often, but only for overall comments. Theoretically we should also be planning to transition these comments to support rich formatting (imo). We don't seem to be interpreting overall  comments as markdown at this point:

Screen Shot 2020-02-18 at 11 24 02 AM

Screen Shot 2020-02-18 at 11 25 22 AM



All of which is to say: (1) I bet the next version of Text doesn't need ug* helpers at all (2) we should decide how we want to handle user-authored links and be consistent with them.

PR #466, opened 2020-03-06
Move the source back from lib/noredink-ui to the root

allenap:
Regarding the noredink-ui migration to the monorepo: there's a lot of work left to do; people have questions and reservations about doing this; having the noredink-ui source in a subfolder is preventing people from publishing new versions. For the last point alone we have to go forwards or go backwards, we can't stay still. Given that going backwards – moving the code back to the top of the noredink-ui repo – is simple and will take the pressure off, Puffins propose doing that for now.


allenap:
The failing security/snyk test is because it's expecting to find `package.json` at the repo root in **master**. It's not checking this branch. Merging this PR ought to fix it.

allenap:
> The failing security/snyk test is because it's expecting to find `package.json` at the repo root in **master**. It's not checking this branch. Merging this PR ought to fix it.

Fwiw, the Snyk test has gone green again since this was merged to master.

PR #473, opened 2020-03-17
Tessa/add banner alert custom

tesk9:
Screen Shot 2020-03-17 at 1 37 19 PM

cc @NoRedInk/design  NOTE: in context, the background and text color will be correct. Pinging for spacing. I kept the same margin as was used in the existing banner alerts, but I think it looks a bit different since the premium flag's right edge is `<  ` rather than `)  `. Also pinging to let you know that after this code is released, we'll be able to support arbitrary icons in the banner alerts.

Adds `custom` helper to BannerAlert. Part of https://www.pivotaltracker.com/story/show/171840495.

I think there's a better API possible here, but not without adding a new version of BannerAlert. I think this is an okay short-term step.


staceyadams:
> Also pinging to let you know that after this code is released, we'll be able to support arbitrary icons in the banner alerts.

🎉

Arkham:
I took a look at the Percy diffs and noticed the premium image got slightly smaller, was this part of the plan?

staceyadams:
Not part of the plan but I think it's ok

PR #483, opened 2020-03-27
Prevent text area with auto resize from jumping scroll position

mthadley:
Here's the summary from the first commit message

> The old logic looked like this:
> 
> ```
> if (this._textarea.scrollHeight > minHeight) {
>   this._textarea.style.height = minHeight + 'px'
>   this._textarea.style.height = this._textarea.scrollHeight + 'px'
> }
> ```
> 
> Imagine a case where the `minHeight` is actually less than the current
> height of the text area due to the current content. The height of the
> element would be set to a smaller value, and then later reset to the
> desired height. This causes the overall scroll height of the page to
> shrink (because the text-area did), and then expand again. Now the
> scroll position for the user on the page has changed, and this happens
> on every keystroke.
> 
> This change just skips the first assignment to `.height`. I'm not sure why it
> was initially there, but I can't see a reason for keeping it (especially
> since it seems to be the cause of this bug).

Here's the bug in action, where the scroll position on the page seemingly jumps every time I input a character:

before

Here's after the change. Notice no jumping as I type:

after

You can play around with the behavior using the same test content below used in the gifs in the current netlify-deployed styleguide, and the one listen in the CI checks (which has the fix).

<details>
  <summary>Lorem Ipsum</summary>
   Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus quis finibus lorem. Donec in ante est. Pellentesque velit ligula, ornare a justo sed, consequat molestie lorem. Vivamus congue, risus eu feugiat mattis, ante eros efficitur arcu, maximus imperdiet leo odio eu risus. Curabitur pellentesque justo sed pulvinar semper. Nunc mi eros, tincidunt in fermentum a, pulvinar ut libero. In dapibus justo lectus, quis consequat tellus tincidunt vel. Vivamus convallis augue eget diam suscipit malesuada. Duis convallis ante est, sit amet tempus nisi venenatis et. Nunc eleifend pretium purus non egestas. In hac habitasse platea dictumst. Morbi leo justo, pharetra sit amet arcu non, ullamcorper rutrum urna. In venenatis, magna sit amet convallis lacinia, urna justo blandit purus, ac interdum diam diam id libero. Praesent ac massa at mauris scelerisque placerat et sit amet massa.

[...]

PR #485, opened 2020-03-31
Bump npm package to 1.2.2

mthadley:
Releases the fix in the TextArea custom element from #483 

Also, I didn't actually need to bump the elm package in #484 since it didn't actually change any Elm code! Woops! I kinda went on autopilot there, since I rarely make changes to our js code here.


mthadley:
Force pushing a new commit since I just remembered that `npm version patch` exists, which makes the exact change I want, and adds the right tag for release.

PR #486, opened 2020-03-31
Spring cleaning

BrianHicks:
I looked at the `Nri.Ui` imports we use internally and removed any that have reached zero uses from this package. This lightens things up quite a bit! I also fixed/removed some broken tests I noticed while doing this 😄


BrianHicks:
the major version bump test failed. If this PR looks good otherwise we will make it pass by bumping the version in here, but I don't want to do it before making sure that everything in here looks good.

README.md
```diff
@@ -20,6 +20,8 @@ Suppose you just released version `5.0.1`, a small styling fix in the checkbox w
 
 To prevent these big Yaks from suddenly showing up in seemingly trivial tasks we prefer to avoid breaking changes in the package. Instead when we need to make a breaking change in a widget, we create a new module for it `Nri.Ui.MyWidget.VX`. Similarly, when we build custom elements in JavaScript we create a file `lib/MyWidget/VX.js` and define a custom element `nri-mywidget-vX`.
 
+That said, we may remove modules that we're not moving occasionally. If you're using this library outside NoRedInk, this guarantee won't hold as strongly for you.
```
tesk9, line 23: Should "moving" be "using"?

tesk9: Also, mild preference to remove the "outside NRI" sentence entirely. It would be a mistake to use this library outside NRI; I don't want to suggest that we're thinking about external users at all, you know?

PR #489, opened 2020-04-01
re-enable major version check

BrianHicks:
I disabled this to publish 8.0.0 in #486, reenabling now.


BrianHicks:
I'm going to merge this without an approval once it passes since it was just a temporary override.

PR #487, opened 2020-03-31
Ink/email inputs

avh4:
This adds an `email` `InputType` for use in fixing https://www.pivotaltracker.com/story/show/172070228

See NOTE in the comments: we want to use `inputmode=email` instead of `type=email`

Also, it turned out to be easy to add a new InputType, so I decided to hold off on making TextInput.V6 and switching to an Attributes-style API.

> Hello there, wonderful author of widgets!
> This is `Nri.Ui.Friendly.AI.V1` speaking to you!
> Are you looking to get this awesome work reviewed as quickly as possible, so you can start putting it to use?
> Then feel free to ask Teambot to assign a reviewer to your PR
>
> https://nri-teambot.herokuapp.com
>
> It's best if we have separate pull requests for code changes and package version bumps. If you're just changing code, great! If you're just bumping the version, check out https://github.com/NoRedInk/noredink-ui#deploying.

PR #490, opened 2020-04-01
restore modules that were actually used

BrianHicks:
it turns out a number of modules I removed were actually used. Sigh. I think this restores all the ones we need.


BrianHicks:
wow, thanks for taking care of these! Are you planning to do more or shall we merge this? I'd like to not bump another major version right away if possible, but I think we're also blocking the upgrade on restoring the ones still here.

tesk9:
I'll probably do more when I have a chance. I think if anyone is blocked, we should merge this and do the major release. But as long as no one is blocked, I don't think there's any hurry. What do you think?

BrianHicks:
That sounds about right. I'll make noise in the right places.

elm.json
```diff
@@ -12,6 +12,8 @@
         "Nri.Ui.AssignmentIcon.V1",
         "Nri.Ui.BannerAlert.V6",
         "Nri.Ui.Button.V5",
+        "Nri.Ui.Button.V6",
```
tesk9, line 15: Okay so because of https://github.com/NoRedInk/noredink-ui/pull/493 I think we can now safely remove Button.V6, and plan to update the one spot that needs a special link type.

elm.json
```diff
@@ -12,6 +12,8 @@
...
+        "Nri.Ui.Button.V7",
```
tesk9, line 16: This update is done!

avh4:
Just a heads up, Ink will be needing a new noredink-ui release by Monday, after https://github.com/NoRedInk/noredink-ui/pull/487 is ready to be merged, which is being actively worked on now.

elm.json
BrianHicks: sounds like a plan

BrianHicks:
@avh4 we're ready for this to merge whenever. Go ahead and hit the button whenever you're ready.

avh4:
Merging now and publishing.  Thanks!

PR #491, opened 2020-04-03
Tessa/nri loading

tesk9:
Nri/Loading and Nri/Loading in the monolith and content creation code are practically identical. This PR extracts to a place that both applications can reference.

loading

cc @NoRedInk/design 
Sorry for the jerky gif. Should be easier to see what it looks like once it's published to Netlify. (IMO, these spinners are a bit fast)

### DIFF
--- a/styleguide-app/Examples/Page.elm
+++ b/styleguide-app/Examples/Page.elm
@@ -21,47 +21,22 @@ import Nri.Ui.Page.V3 as Page
 
 {-| -}
 type alias State =
-    { showLoadingFadeIn : Bool
-    , showLoading : Bool
-    }
+    {}
 
 
 init : State
 init =
-    { showLoadingFadeIn = False
-    , showLoading = False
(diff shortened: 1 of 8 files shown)


styleguide-app/Examples/Page.elm
```diff
@@ -31,7 +31,13 @@ example =
...
     , categories = List.singleton Pages
     , state = ()
-    , update = \_ state -> ( state, Cmd.none )
+    , update =
+        \msg model ->
+            let
+                _ =
+                    Debug.log "Clicked: " msg
+            in
+            ( model, Cmd.none )
```
Arkham, line 40: 🎨 do we want to keep this debug in?

tesk9: Yes, I think so -- many of the component examples currently use a debug.log to show that they do something 🤷‍♀

PR #492, opened 2020-04-08
Tessa/fix styleguide link

tesk9:
I'm removing the old divider component from the old styleguide (https://staging.noredink.com/style_guide#doodad/Nri%2FDivider.elm) and putting it into the netlify one.

I'm noticing that some of the styles differ: the netlify version has a gray45 title without a set font, and the /styleguide version has a navy title using baseFont (Multi).

Not sure whether to make changes or not -- is one of these behaviors right? cc @NoRedInk/design 

### /styleguide
Screen Shot 2020-04-08 at 3 14 20 PM

### New netlify addition
Screen Shot 2020-04-08 at 3 14 29 PM

### DIFF
--- a/src/Nri/Ui/Divider/V2.elm
+++ b/src/Nri/Ui/Divider/V2.elm
@@ -1,6 +1,8 @@
 module Nri.Ui.Divider.V2 exposing (view)
 
-{-|
+{-| Patch changes:
+
+  - Title is navy, Muli, fontWeight 600
 
 @docs view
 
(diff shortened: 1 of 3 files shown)


bendansby:
Navy and Muli is what we want

tesk9:
Screen Shot 2020-04-08 at 3 33 14 PM

Thanks! Look okay? (I also pulled in the fontweight from the other version because it looked a little anemic without)

bendansby:
Looks good!

PR #495, opened 2020-04-13
Ink/textinput v6

avh4:
TODO:
- [x] make V6 using attributes-style API
- [x] write elm-refactor script for TextInput.V5
- [x] add `custom`
- [x] add error messages
- [x] add disabled / loading

follow-up PR:
- [x] add topMargin attribute
- [ ] decide if there's anything beyond `custom` we want to add to support `autocomplete` attributes
- [ ] investigate if we want to try to set `name` attribute in general (see https://developers.google.com/web/fundamentals/design-and-ux/input/forms#recommended_input_name_and_autocomplete_attribute_values)
- [ ] write elm-refactor scripts for TextInput.V3,  TextInput.V4

>
>

### DIFF
--- a/src/Nri/Ui/TextInput/V6.elm
+++ b/src/Nri/Ui/TextInput/V6.elm
@@ -125,21 +125,15 @@ email toMsg =
 {-| An optional customization of a TextInput.
 -}
 type Attribute msg
-    = InputStyleAttribute InputStyles.Theme
-    | ErrorAttribute ErrorState
-    | HideLabelAttribute Bool
-    | PlaceholderAttribute String
-    | OnBlurAttribute msg
-    | AutofocusAttribute Bool
-    | Attribute (Config msg -> Config msg)
-    | CustomAttribute (Html.Attribute msg)
+    = Attribute (Config msg -> Config msg)
 
 
 {-| If not explicit placeholder is given, the input label will be used as the placeholder.
 -}
 placeholder : String -> Attribute msg
 placeholder text_ =
-    PlaceholderAttribute text_
+    Attribute <|
+        \config -> { config | placeholder = Just text_ }
 
 
 {-| This disables the input
@@ -163,12 +157,16 @@ If you are always passing `True`, then you don't need to use this attribute.
 -}
 errorIf : Bool -> Attribute msg
 errorIf isInError =
-    ErrorAttribute <|
-        if isInError then
-            Error { message = Nothing }
+    Attribute <|
+        \config ->
+            { config
+                | error =
+                    if isInError then
+                        Error { message = Nothing }
 
-        else
-            NoError
+                    else
+                        NoError
+            }
 
 
 {-| If `Just`, the field will be highlighted as having a validation error,
@@ -176,33 +174,41 @@ and the given error message will be shown.
 -}
 errorMessage : Maybe String -> Attribute msg
 errorMessage maybeMessage =
-    case maybeMessage of
-        Nothing ->
-            ErrorAttribute NoError
+    Attribute <|
+        \config ->
+            { config
+                | error =
+                    case maybeMessage of
+                        Nothing ->
+                            NoError
 
-        Just message ->
-            ErrorAttribute <| Error { message = Just message }
+                        Just message ->
+                            Error { message = Just message }
+            }
 
 
 {-| Hides the visible label. (There will still be an invisible label for screen readers.)
 -}
 hiddenLabel : Attribute msg
 hiddenLabel =
-    HideLabelAttribute True
+    Attribute <|
+        \config -> { config | hideLabel = True }
 
 
 {-| Causes the TextInput to produce the given `msg` when the field is blurred.
 -}
 onBlur : msg -> Attribute msg
 onBlur msg =
-    OnBlurAttribute msg
+    Attribute <|
+        \config -> { config | onBlur = Just msg }
 
 
 {-| Sets the `autofocus` attribute of the resulting HTML input.
 -}
 autofocus : Attribute msg
 autofocus =
-    AutofocusAttribute True
+    Attribute <|
+        \config -> { config | autofocus = True }
 
 
 {-| Add any attribute to the input.
@@ -213,7 +219,8 @@ consider adding to the TextInput API to support what you need.
 -}
 custom : Html.Attribute msg -> Attribute msg
 custom attr =
-    CustomAttribute attr
+    Attribute <|
+        \config -> { config | custom = attr :: config.custom }
 
 
 {-| This is private. The public API only exposes `Attribute`.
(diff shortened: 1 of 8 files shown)


styleguide-app/Examples/TextInput.elm
```diff
@@ -60,89 +60,106 @@ example =
...
+                , TextInput.view (exampleConfig.label ++ " (text)")
+                    (TextInput.text (SetTextInput 1))
+                    ([ TextInput.errorIf exampleConfig.isInError
+                     , TextInput.placeholder exampleConfig.placeholder
+                     ]
+                        ++ (if exampleConfig.showLabel then
+                                []
+
+                            else
+                                [ TextInput.hiddenLabel ]
```
tesk9, line 72: 💭 I wonder if `hiddenLabel : Bool -> Attribute msg` might make this a bit nicer?

Alternatively, maybe adding `custom : List (Html.Attribute msg) -> Attribute msg` might help us avoid some parens and brackets

src/Nri/Ui/TextInput/V6.elm
```diff
@@ -0,0 +1,316 @@
...
+    div
+        [ Attributes.css [ position relative ]
+        ]
+        [ input
+            (maybeStep
+                ++ [ Attributes.id idValue
+                   , css
+                        [ InputStyles.input config.inputStyle config.isInError
+                        , if config.inputStyle == InputStyles.Writing then
+                            Css.Global.withClass "override-sass-styles"
```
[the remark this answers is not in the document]
avh4: not sure... all I did in this function was path up how to get the config values from their new places.  I didn't want to accidentally break the Writing TextInput.  But I can try to investigate if this can be removed if you want me to.

styleguide-app/Examples/TextInput.elm
avh4: Yeah, I was thinking about this...   I think this is only weird in the style-guide.   In real code, each TextInput you create will either have a hidden label or not, it's not going to dynamically change.   I think this is a general issue with elm-debug-controls, in that it makes your "example" code way more dynamic that what you'd actually write when using the component.

I was thinking of just leaving this here for now, and thinking about this a bit longer.  But another option I thought of would be to add a `TextInput.hiddenLabelIf` which could be used here in the styleguide.

src/Nri/Ui/TextInput/V6.elm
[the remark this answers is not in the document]
tesk9: Hmm it might be a project to figure out, given that we added the override-sass-styles class to get out of actually fixing our inherited styles issues. 🤔 Feel free to investigate if you have the time & energy, but it's not a blocker.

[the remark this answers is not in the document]
avh4: okay, I'll leave this alone, then, since we want to get this merged to monolith tomorrow

styleguide-app/Examples/TextInput.elm
tesk9: Or perhaps we could add a `visibleLabel` attribute? In `Button`, we did expose helpers for the defaults even though they're defaults, and I don't think it's been bad.

I don't have a strong opinion about strategy, but I think it is nice if the style guide code is as easy-to-read as possible since people will probably look at it to get code examples.

styleguide-app/Examples/TextInput.elm
```diff
@@ -60,89 +60,106 @@ example =
                 [ Control.view UpdateControl state.control
                     |> Html.fromUnstyled
                 , Heading.h3 [] [ text "TextInput.view { type_ = TextInput.text }" ]
```
tesk9, line 62: Oh! One other thing -- could we update these headings? They're funky now given the glorious new API

avh4: Oh, you mean there's a new Heading API to switch to?  I'll give it a try.

styleguide-app/Examples/TextInput.elm
avh4: Ah, I realized the debug controls were out of date for the new API, so I changed those to look like this:

Screen Shot 2020-04-14 at 8 55 52 AM
Screen Shot 2020-04-14 at 8 56 03 AM

And the styleguide code is now like this:
```
                , TextInput.view (exampleConfig.label ++ " (writing)")
                    (TextInput.text (SetTextInput 4))
                    (List.filterMap identity
                        [ Just TextInput.writing
                        , exampleConfig.maybeErrorAttribute
                        , exampleConfig.maybePlaceholderAttribute
                        , exampleConfig.maybeShowLabelAttribute
                        ]
                    )
                    (Maybe.withDefault "" <| Dict.get 4 state.stringInputValues)
```

which imo is acceptable.  what do you think?

tesk9: Def acceptable! I think I'd vaguely prefer for the error and hidden APIs to match each other a little more closely?

Current version:
```
placeholder : String -> Attribute value msg
onBlur : msg -> Attribute value msg
autofocus : Attribute value msg
hiddenLabel : Attribute value msg
errorIf : Bool -> Attribute value msg
```

Suggested version:
```
placeholder : String -> Attribute value msg
onBlur : msg -> Attribute value msg
autofocus : Bool ->Attribute value msg
hiddenLabelIf : Bool -> Attribute value msg
errorIf : Bool -> Attribute value msg
```

And then rather than a dropdown, we could use the checkbox bool setting from Debug Controls for `hiddenLabelIf`, `autofocus`, and `errorIf`?

styleguide-app/Examples/TextInput.elm
tesk9: Oh! I meant "TextInput.view { type_ = TextInput.text }" was no longer quite correct

tesk9: Not a big deal though

avh4: Hmm, maybe you meant just that the spacing is messy because headers have no built-in margins?   So ...  Grid layout!  
Screen Shot 2020-04-14 at 10 56 15 AM

src/Nri/Ui/TextInput/V6.elm
```diff
@@ -125,21 +125,15 @@ email toMsg =
...
 type Attribute msg
-    = InputStyleAttribute InputStyles.Theme
-    | ErrorAttribute ErrorState
-    | HideLabelAttribute Bool
-    | PlaceholderAttribute String
-    | OnBlurAttribute msg
-    | AutofocusAttribute Bool
-    | Attribute (Config msg -> Config msg)
-    | CustomAttribute (Html.Attribute msg)
+    = Attribute (Config msg -> Config msg)
```
lukewestby, line 128: Love this

avh4: Taken from Labradoodle's work!

styleguide-app/Examples/TextInput.elm
avh4: I made another attempt at cleaning up the use of debug controls here.   It's still not the best, but it gave me an idea of what to add to elm-debug-controls itself to have better support for an attributes-style API.

I do agree the style guide code doesn't look great yet, but I'm also hesitant to add stuff to the public API prematurely just to support the style guide.  I think the API overall is stable enough that any of the additional functions you're suggesting could be added as a minor version bump if they turn out to be useful in real code.

I need to switch over to integrating this into the monolith so we can try to finish a story for our release today, so I'm gonna merge this now.  But I think I understand your concerns, and will continue to think about (and worst case, if we do need a V7, I'll volunteer to write the upgrade script to eliminate V6).

tesk9:
Thanks, AV! 👏 Yay improvements!!!

PR #502, opened 2020-04-27
Nri.Ui.Message.V1

avh4:
Nri.Ui.Message.V1 replaces Alert.V4 (becomes Message.tiny) and BannerAlert.V6 (becomes Message.banner), and adds a new Message.large

Preview of new styleguide examples: https://deploy-preview-502--noredink-ui.netlify.app/#/doodad/Nri.Ui.Message.V1

REVIEWER NOTE: there are a lot of commits, but I'm pretty sure this will be easiest to review commit-by-commit.  All code moved/copied between modules is directly moved/copied with no changes in one commit, and then any modifications are done in later commits.

Main questions for early code review:
- should the Theme constructors be private and have functions to create each type?
- should Content not exist, and instead the caller always passes HTML?  (this would clean up the API a bit, but would make it more annoying to pass a markdown string)

TODO:
- [x] convert Alert
- [x] convert AlertBanner
- [x] add `Message.large` new component

New component:
Screen Shot 2020-04-28 at 12 46 06 PM

Fixed alignment for `Message.tiny` (was `Alert.V4.*`):
Screen Shot 2020-04-28 at 1 06 08 PM
Screen Shot 2020-04-28 at 1 05 57 PM

>
>


src/Nri/Ui/Message/V1.elm
```diff
@@ -162,7 +162,7 @@ large theme content =
             case theme of
                 Error ->
                     { backgroundColor = Colors.purpleLight
-                    , fontColor = Colors.purple
+                    , fontColor = Colors.purpleDark
```
lukewestby, line 222: 🥇

avh4:
> It is my general opinion that markdown is more of a data concern than a UI concern, and in this case it really would clean up the API quite a bit to move markdown rendering out and just pass Html. I'll leave it to you to decide but I'm in favor of this change.

Good point!  From seeing the uses of the old API, it looks like it has been really nice to be able to quickly specify markdown, and the line of code to parse and render it is a bit non-intuitive, so I think this is a good open question, but I'm gonna lean toward leaving it as an option for now, and also can be a bit of an experiment to see if having direct markdown support in other modules might be useful.

avh4:
Labradoodle:  if you have later comments, lmk, and I can make a Message.V2 if necessary.

BrianHicks:
we have yet to define what it actually means for Lab to "own" noredink-ui. In the meantime, I don't see why refactors like this should require our special approval. 👍

PR #503, opened 2020-04-28
Make Nri.Ui.Tabs.V4 work with ProgramTest.clickButton

avh4:
Preview: https://deploy-preview-503--noredink-ui.netlify.app/#/doodad/Nri.Ui.Tabs.V4

This also applies the guidelines from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role#Best_practices
- elements with `role=tab` should be `button` elements
- `tabindex` should be `0` for the active tab, and `-1` for other tabs in the group

I tested using keyboard navigation in Chrome and in Safari (with Preferences -> Advanced -> Accessibility -> Press Tab to highlight each item on a webpage -> Enabled), and it appears to still work.

>
>

### DIFF
--- a/tests/Spec/Nri/Ui/Tabs/V4.elm
+++ b/tests/Spec/Nri/Ui/Tabs/V4.elm
@@ -1,0 +1,36 @@
+module Spec.Nri.Ui.Tabs.V4 exposing (all)
+
+import Accessibility.Styled as Html
+import Expect
+import List.Zipper.Extra
+import Nri.Ui.Tabs.V4 as Tabs
+import ProgramTest
+import Test exposing (..)
+
+
+all : Test
+all =
+    describe "Nri.Ui.Tabs.V4"
+        [ test "works with ProgramTest.clickButton" <|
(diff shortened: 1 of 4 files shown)


tests/Spec/Nri/Ui/Tabs/V4.elm
```diff
@@ -0,0 +1,36 @@
...
+                                    List.Zipper.Extra.from []
+                                        (Tabs.Tab "First tab" "ID_FIRST")
+                                        [ Tabs.Tab "Second tab" "ID_SECOND" ]
+                                , content = \_ -> Html.text ""
+                                , alignment = Tabs.Center
+                                }
+                                |> Html.toUnstyled
+                    }
+                    |> ProgramTest.start ()
+                    |> ProgramTest.clickButton "Second tab"
```
avh4, line 34: Before the change in this PR to use `Html.button`, this line would fail to find the tab.

PR #507, opened 2020-05-01
Tabs tweak

bendansby:
Adjusts tab styles to partially outfox: https://www.pivotaltracker.com/story/show/172591130

Shouldn't have a deleterious effect on other tab instances.

image


avh4:
@bendansby is it okay that the left-aligned tabs now will be flush to the page margin?  
Screen Shot 2020-05-01 at 10 01 21 AM

bendansby:
Yup!

avh4:
Looks like if we don't want left/right aligned flush, then adding `padding2 zero (px 20)` would put it back but still keep the centered ones centered properly

PR #520, opened 2020-05-20
Adjust llnk styles

bendansby:
Of large messages

Napkin 7 05-20-20, 10 45 49 AM


bendansby:
Text-decoration is uglier! Most nice looking sites implement underline with borders. It also has the benefit of not being overridden by Bootstrap

PR #521, opened 2020-05-20
Zamboni/maybe default segment

glitteringkatie:
This adds the ability to not have a default selection in a quick and dirty way so zamboni can wrap up https://www.pivotaltracker.com/story/show/172845604 faster. I'll be going back and doing clean up in my TCR with the intention of DRYing things up. Also broke out the example states so they'll all work uniquely.

- [x] Fix build
- [ ] Get approval
- [ ] merge
- [ ] bump
- [ ] publish
then we can use!

### No Default
image

### Each in unique states 🙌 
image

cc @bendansby

### DIFF
--- a/src/Nri/Ui/SegmentedControl/V8.elm
+++ b/src/Nri/Ui/SegmentedControl/V8.elm
@@ -1,8 +1,8 @@
-module Nri.Ui.SegmentedControl.V8 exposing (Config, Option, Width(..), view, viewSpa, ToggleConfig, viewToggle)
+module Nri.Ui.SegmentedControl.V8 exposing (Config, Option, Width(..), view, viewSpa, ToggleConfig, viewToggle, viewOptionalSelectToggle)
 
 {-|
 
-@docs Config, Option, Width, view, viewSpa, ToggleConfig, viewToggle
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/SegmentedControl/V8.elm
```diff
@@ -56,6 +56,19 @@ type alias ToggleConfig a msg =
     }
 
 
+{-| Same shape as ToggleConfig but with an optional selected. This would ideally
+be the same as ToggleConfig but we as Zambonis don't have time in the ticket to
+also upgrade all existing uses of viewToggle. Katie is mentally noting this as a
+good hackday clean up but if you find it and fix it first, that's great too!
+-}
```
asterite, line 63: 💭 I think it actually makes sense to have two different configs, because in the case of "there's always a selection" the selected shouldn't be a `Maybe`.

glitteringkatie: I don't want to change the config for the nav element because there _should_ always be a selection. But for the element that acts kind of like a select, we should think of it as a radio button collection--it's possible to not have a default! But it's easy for the end user to provide a default by passing `Just blah`.

PR #523, opened 2020-05-22
TCR: Segmented Toggle Cleanup

glitteringkatie:
Zamboni Tech Care Retreat

This aims to simplify the work done in https://github.com/NoRedInk/noredink-ui/pull/521 and better differentiate between the select-only component and the tab component. Reviewing commit by commit should be a good-enough experience.


src/Nri/Ui/SegmentedControl/V9.elm
```diff
@@ -0,0 +1,304 @@
...
+{-| Creates _just the segmented select_ when you need the ui element itself and
+not a page control
+-}
+viewSelect : SelectConfig a msg -> Html msg
+viewSelect config =
+    div
+        [ css
+            [ displayFlex
+            , cursor pointer
+            ]
```
asterite, line 106: css attributes should also contain `role.radioGroup`

styleguide-app/Examples/SegmentedControl.elm
```diff
@@ -55,17 +55,9 @@ example =
                 }
             , Html.h3 [] [ Html.text "Toggle only view" ]
```
glitteringkatie, line 56: Should say "Select only view"

PR #532, opened 2020-06-09
Lab/extend clickable helpers

tesk9:
clickable svg tooltips

cc @NoRedInk/design 

```
---- Nri.Ui.ClickableSvg.V1 - MINOR ----

    Added:
        withTooltipAbove :
            { id : String.String
            , isOpen : Basics.Bool
            , onShow : Basics.Bool -> msg
            }
            -> Nri.Ui.ClickableSvg.V1.Attribute msg
        withTooltipBelow :
            { id : String.String
            , isOpen : Basics.Bool
            , onShow : Basics.Bool -> msg
            }
            -> Nri.Ui.ClickableSvg.V1.Attribute msg
```

### DIFF
--- a/src/Nri/Ui/ClickableSvg/V1.elm
+++ b/src/Nri/Ui/ClickableSvg/V1.elm
@@ -6,6 +6,7 @@ module Nri.Ui.ClickableSvg.V1 exposing
     , width, height
     , disabled
     , custom, css
+    , withTooltipAbove, withTooltipBelow
     )
 
 {-|
@@ -14,6 +15,7 @@ module Nri.Ui.ClickableSvg.V1 exposing
 # Post-release patches
 
   - uses ClickableAttributes
+  - adds tooltip helpers
 
 
 # Create a button or link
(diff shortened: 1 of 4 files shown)


staceyadams:
🎉

tesk9:
Actually, wait -- I suppose for the _tabs_ we don't want an a tag or a button -- we want a regular svg. 🤔

BrianHicks:
so review now or wait?

tesk9:
Review now! This is in the direction we want, even if we don't end up using this the way I was anticipating us to 😅 Thanks

src/Nri/Ui/ClickableSvg/V1.elm
```diff
@@ -215,16 +216,59 @@ css styles =
...
-    set identity
+    withTooltip Tooltip.OnTop
 
 
 {-| -}
-withTooltipBelow : Attribute msg
+withTooltipBelow : { id : String, isOpen : Bool, onShow : Bool -> msg } -> Attribute msg
 withTooltipBelow =
-    set identity
+    withTooltip Tooltip.OnBottom
```
BrianHicks: I think it'd make more sense to accept the `Tooltip.Position` in here. Who knows if we'll need tooltips to the left or right someday.

Issue #300, opened 2019-06-21
Tabs.V4


tesk9:
Often, when using Tabs, the caller needs to do something like this:

```
   model.tabs
                        |> Zipper.first
                        |> Zipper.find (.id >> (==) tabId)
                        |> Maybe.withDefault model.tabs
```
because the Tabs API asks for `{ onSelect : id -> msg, ...}`

I think it might be nicer to have an API like this:

```
type alias Tabs =
    Zipper (Tab id) 

init : List (Tab id) -> Tab id -> List (Tab id) -> Tabs
init = 
    Zipper.Zipper

type alias Config id msg =
    { title : Maybe String
    , onSelect : Tabs -> msg
    , tabs : Tabs
    , content : id -> Html msg
    , alignment : Alignment
    }
```
So that the caller doesn't need to know about List.Zipper or touch it.

avh4:
On curriculum library, the design spec requires that sometimes no tab be selected -- so I think we'll have to not use a zipper to implement that.

Also, I think that it's generally turned out to be not ideal to store the UI state (which tab is selected) mixed together with the static data (the objects which are being represented by tabs),  so I really think the component should take a `List Tab` and `selected : Maybe id` instead of a zipper.

tesk9:
I looked around in the curriculum library product spec, and I don't fully understand what it would mean for there not to be a selected tab. Would the tabs still show, but without a rendered tab panel? just empty space?

avh4:
It's when the teacher is searching:  https://paper.dropbox.com/doc/Product-Spec-Curriculum-Library--AfiLHS1XYUJAoIoXIGF3_auvAg-CHaw3XzNKq4MePdinHDRc#:uid=054397348133745218381189&h2=Search  the three normal tabs show, but none of them are selected.  There is a rendered content area for the search results, and it's under the tab bar -- just none of the tabs appear selected.  (Maybe for accessibility purposes there is actually a selected "Search" tab, but it's invisible to sighted users?)

tesk9:
Ahh, thanks! I think that makes sense, yeah!

PR #533, opened 2020-06-10
Lab/tabs

tesk9:
Screen Shot 2020-06-09 at 6 13 39 PM

Part of https://www.pivotaltracker.com/story/show/173250938
See https://github.com/NoRedInk/noredink-ui/issues/300 for some background.


src/Nri/Ui/Tabs/V5.elm
```diff
@@ -0,0 +1,341 @@
...
+type Alignment
+    = Left
+    | Center
+    | Right
+
+
+{-| -}
+type alias Tab id msg =
+    { id : id
+    , idString : String
```
BrianHicks, line 44: wouldn't it be neater to ask for a `id -> String` function in `view`?

src/Nri/Ui/Tabs/V5.elm
```diff
@@ -0,0 +1,341 @@
...
+{-| -}
+view :
+    { title : Maybe String
+    , alignment : Alignment
+    , customSpacing : Maybe Float
+    , onSelect : id -> msg
+    , onFocus : String -> msg
+    , selected : id
+    , tabs : List (Tab id msg)
+    }
```
BrianHicks, line 60: it's starting to feel like there's enough config here to move to an HTML-like pattern.

src/Nri/Ui/Tabs/V5.elm
tesk9: I think if we do the id -> String pattern, we would also want to do id -> Html msg for the tab and id -> Html msg for the tabpanel. My concern with that strategy is that you end up with the tab view separated from the panel view in the code, even though conceptually they're tied together.

Like I'd expect using the API to look like this:
```
type TabId
    = Passage
    | Feedback
    | TipsAndExamples
    | Rubric


idToString : TabId -> String
idToString tabId =
    case tabId of
        Passage ->
            "Passage"

        Feedback ->
            "Feedback"

        TipsAndExamples ->
            "TipsAndExamples"

        Rubric ->
            "Rubric"


viewTab : TabId -> Html Msg
viewTab tabId =
    case tabId of
        Passage ->
            text "Passage"

        Feedback ->
            text "Feedback"

        TipsAndExamples ->
            text "Tips and Examples"

        Rubric ->
            text "Rubric"


viewPanel : TabId -> Html Msg
viewPanel tabId =
    case tabId of
        Passage ->
            div []
                [ text "Passage"
                , text "blah blah"
                ]

        Feedback ->
            div [] [ "Feedback" ]

        TipsAndExamples ->
            div [] []

        Rubric ->
            div [] []

```

Versus:

```

allTabs =
    [ { id = Passage
      , idString = "passage"
      , spaHref = Nothing
      , tabView = Html.text "Passage"
      , panelView =
            div []
                [ text "Passage"
                , text "blah blah"
                ]
      }
    , { id = Feedback
      , idString = ""
      , spaHref = Nothing
      , tabView = Html.text "Feedback"
      , panelView = div [] [ "Feedback" ]
      }
    , { id = TipsAndExamples
      , idString = ""
      , spaHref = Nothing
      , tabView = Html.text "Tips & Examples"
      , panelView = div [] []
      }
    , { id = Rubric
      , idString = "rubric"
      , spaHref = Nothing
      , tabView = Html.text "Rubric"
      , panelView = div [] []
      }
    ]
```

src/Nri/Ui/Tabs/V5.elm
tesk9: I'm not sure -- there are 4 mandatory values: `onSelect`, `onFocus`, `selected`, and `tabs`.

That leaves `alignment`, `title`, and `customSpacing`. I'm pretty sure that `title` should only be present when the `alignment` is `Right`... I could set a default for the two remaining properties, `alignment: Left` and `customSpacing: Nothing` but I'm not sure that at this point, that's better than the config-based API.

PR #539, opened 2020-06-17
small fixes for shared input transitions

BrianHicks:
the shared input styles were animating too much. In particular, they were animating border width, shadow size, etc, and making the text input jump around when it entered the page when they overrode sass styles in the monorepo.

This applies the fixes I patched in in the monorepo to the upstream library. Once this is in, I'll upgrade downstream and remove my `!important`.


tesk9:
Dang https://media.giphy.com/media/XBhzUNf7ta00w/giphy.gif

BrianHicks:
Ooohhh worth linking!

PR #541, opened 2020-06-19
Show keyboard shortcuts

tesk9:
- adds "atomic type" filters (see https://bradfrost.com/blog/post/atomic-web-design/.) I've heard some enthusiasm (cc @BrianHicks , @lukewestby) for us exploring this style of design system. I am not particularly familiar with it, and I didn't take too much time sorting the components into sections. Very MVP!
- adds links to the source files directly (the source link on the package site takes you to the root of the project, which is never where I want to go 🤷 )
- adds section for specifying the desired keyboard behavior. I only added keyboard support information to Nri.Ui.Tabs.V5's examples, but now it should be straightforward to add for any component. And also we ought to! In the future, I'd also like for the keyboard support details section to get cute -- currently, pretty ugly 😅 

My hope is that the keyboard support section can help Design, Eng, and QA to agree on desired behavior, and then get it implemented and tested and out into the world! ⌨️ 

Screen Shot 2020-06-19 at 2 50 33 PM

cc @NoRedInk/design 

also cc @krinhorn (this PR is largely inspired by the Juneteenth conf discussion, and in particular your comment here https://noredink.slack.com/archives/C0162CRJ9KK/p1592592178026100?thread_ts=1592589882.015000&cid=C0162CRJ9KK . Sounds like we need more visibility! Would love your thoughts on this 🙏 )

### DIFF
--- a/styleguide-app/AtomicDesignType.elm
+++ b/styleguide-app/AtomicDesignType.elm
@@ -1,0 +1,50 @@
+module AtomicDesignType exposing
+    ( AtomicDesignType(..)
+    , sorter
+    , toString
+    )
+
+{-|
+
+@docs AtomicDesignType
+@docs sorter
+@docs toString
+
+-}
+
+import Sort exposing (Sorter)
+
+
+{-| -}
+type AtomicDesignType
+    = Atom
+    | Molecule
+    | Organism
+    | Template
+    | Page
+
+
+{-| -}
+sorter : Sorter AtomicDesignType
+sorter =
+    Sort.by toString Sort.alphabetical
+
+
+{-| -}
+toString : AtomicDesignType -> String
+toString atomicDesignType =
+    case atomicDesignType of
+        Atom ->
+            "Atom"
+
+        Molecule ->
+            "Molecule"
+
+        Organism ->
+            "Organism"
+
+        Template ->
+            "Template"
+
+        Page ->
+            "Page"
--- a/styleguide-app/AtomicDesignType.elm
+++ b/styleguide-app/AtomicDesignType.elm
@@ -36,7 +36,25 @@ all =
 {-| -}
 sorter : Sorter AtomicDesignType
 sorter =
-    Sort.by toString Sort.alphabetical
+    Sort.by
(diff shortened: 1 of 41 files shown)


staceyadams:
> adds section for specifying the desired keyboard behavior. 

Does "desired" mean the behavior we eventually want or the behavior that currently exists?

>  In the future, I'd also like for the keyboard support details section to get cute -- currently, pretty ugly 😅

It's visible and readable! This will make a huge difference to me in terms of what I take into account when I'm looking at something in the style guide.

tesk9:
Desired behavior == behavior that currently exists, and behavior that we should double-check continues to exist 😅 

YAY glad this will be useful! <3 <3

styleguide-app/AtomicDesignType.elm
```diff
@@ -0,0 +1,50 @@
...
+    | Molecule
+    | Organism
+    | Template
+    | Page
+
+
+{-| -}
+sorter : Sorter AtomicDesignType
+sorter =
+    Sort.by toString Sort.alphabetical
```
BrianHicks, line 39: 💭 this will sort Atom, Molecule, Organism, **Page**, Template. Should it sort Page last?

styleguide-app/Examples/Button.elm
```diff
@@ -27,6 +28,7 @@ example =
     , subscriptions = \_ -> Sub.none
     , view = \state -> [ viewButtonExamples state ]
     , categories = [ Buttons ]
+    , atomicDesignType = AtomicDesignType.Molecule
```
BrianHicks: wouldn't this be an Atom?

styleguide-app/Examples/Confetti.elm
```diff
@@ -22,6 +23,7 @@ example : Example State Msg
 example =
     { name = "Nri.Ui.Confetti.V2"
     , categories = [ Animations ]
+    , atomicDesignType = AtomicDesignType.Atom
```
BrianHicks: maybe Molecule because animation + colors?

PR #542, opened 2020-06-22
use ugFont for standard input styles

BrianHicks:


BrianHicks:
@NoRedInk/design does the Percy diff here look OK to you?

staceyadams:
LGTM 👍

tesk9:
What is happening here? We're changing all of our input styles from basefont to ugfont?

BrianHicks:
so here's the deal… I'm not sure this is the right thing to do. I thought it'd be fine but the Percy diff is making me think it's not. I requested reviews before it was done; sorry for wasting y'all's time!

The point of what I'm doing is to get the ugFont for the writing prompt and passage editor. Buuuut I don't want to have an override on those specific editors. @tesk9 would it match your intent in designing this module to add a specific `UserGenerated` case?

tesk9:
I don't think I made InputStyles -- I don't have strong opinions on how these styles ought to work. Having a separate UserGenerated style sounds fine to me

BrianHicks:
ok! now it's separate and the other text inputs won't change.

script/deprecated-imports.csv
```diff
@@ -8,8 +8,10 @@ src/Nri/Ui/Page/V3.elm,Nri.Ui.Text,2
 src/Nri/Ui/SlideModal/V2.elm,Nri.Ui.Button,8
 src/Nri/Ui/SlideModal/V2.elm,Nri.Ui.Icon,3
 src/Nri/Ui/SlideModal/V2.elm,Nri.Ui.Text,2
+src/Nri/Ui/TextInput/V6.elm,Nri.Ui.InputStyles,2
```
tesk9, line 11: Should we update these..?

BrianHicks: I mean, yeahhhhhhhhhhhh but also



If you really think we should, we can, but it seems to require bumping both versions to their next `Vwhatever` and this is already going kind of off our critical path for TDA right now.

tesk9: Oh -- i'm surprised that's not a patch change! happy to wait then

PR #557, opened 2020-07-24
no margin for message buttons

bendansby:
Fix to hopefully address the elevated button in this banner

Screen Shot 2020-07-24 at 9 23 04 AM

### DIFF
--- a/src/Nri/Ui/Message/V1.elm
+++ b/src/Nri/Ui/Message/V1.elm
@@ -280,6 +280,8 @@ large theme content =
                 , borderBottom3 (px 1) solid Colors.azure
                 , visited [ color Colors.azure ]
                 ]
+            , Css.Global.button
+                [ margin zero ]
             ]
         ]
         []
@@ -445,6 +447,8 @@ banner theme content attr =
                         , borderBottom3 (px 1) solid Colors.azure
                         , visited [ color Colors.azure ]
                         ]
+                    , Css.Global.button
+                        [ margin zero ]
                     ]
                 ]
                 []


tesk9:
@bendansby I don't think it's the margin that's causing the alignment issue. I'm pretty sure that the problem is that we're using a border-bottom rather than text-underline to make the ClickableText look like a link, and that pushes up the text by 1px.

I made a commit removing the border bottom from just that one place here, just to verify the issue: https://github.com/NoRedInk/NoRedInk/compare/tessa/activate-premium-button

without border bottom

I'm not sure what the actual fix is here -- seems like we do need either the border bottom or the underline.

bendansby:
The button has default margin on it though and zeroing it out fixes it?

Screen Shot 2020-07-24 at 9 47 25 AM

tesk9:
Screen Shot 2020-07-24 at 9 53 13 AM

Screen Shot 2020-07-24 at 9 54 06 AM

I'm seeing it as not lined up?

bendansby:
Ahh beaten at my own persnickety game! In that case it's an issue of both margin and border? Doing one or the other only gets it part of the way there for me.

tesk9:
However we want to fix it, I think it probably makes sense to do with `ClickableText.css []` over here https://github.com/NoRedInk/NoRedInk/blob/master/monolith/ui/src/Page/Teach/TeacherDashboard/View.elm#L69 rather than in the Message component.

bendansby:
That makes sense! I'll close this PR and open one in the monolith.

PR #560, opened 2020-07-29
Tessa/expand segmented control example

tesk9:
Adds setting for demoing 2 to 8 segmented controls.

Screen Shot 2020-07-29 at 11 10 11 AM

(Note: done as investigation for https://paper.dropbox.com/doc/Discussion-TDA-Multiple-Passages-Next-Steps--A4usM7i4rAFd2_ZFa5MKBD4mAg-qxJkBGoZ1C4TaUsmJ4NX3#:uid=634876250647832936652806&h2=8-segmented-controls-explorati)

cc @staceyadams -- making this change so that you can see the segmented control behavior with lots of options directly


styleguide-app/Examples/SegmentedControl.elm
```diff
@@ -70,6 +70,32 @@ example =
     }
 
 
+coloredIcons : List Svg
+coloredIcons =
+    [ UiIcon.flag
+    , UiIcon.sprout
+    , UiIcon.star
+    , UiIcon.sapling
+    , Svg.withColor Colors.greenDark UiIcon.attention
```
BrianHicks, line 79: this one overflows the bounds of the container, even in your screenshot. Should that be possible?

tesk9: It does indeed! it's an issue with the segmented control itself -- see:
 
Screen Shot 2020-07-29 at 1 15 01 PM

I'm not sure if it matters to us yet, but I think there's some value in showing limitations within the styleguide too. Easily fixable on the use-side by setting a height on the icon too, I think..

staceyadams:
> cc @staceyadams -- making this change so that you can see the segmented control behavior with lots of options directly

Thank you! 🙏

styleguide-app/Examples/SegmentedControl.elm
staceyadams: The ❗ is just really excited!

PR #569, opened 2020-08-05
Lab/stickyable segmented control

tesk9:
https://www.pivotaltracker.com/story/show/174178319

Add SegmentedControl V10.

Screen Shot 2020-08-04 at 9 38 17 PM

  - hides non-displayed content rather than fully removing from the DOM, allowing for the segmented control to be used as a sticky-header that doesn't lose scroll position
  - :skull: removes NavConfig and SelectConfig
  - combines `view` and `viewSpa` (for V9 `view` behavior, be sure `toUrl` is Nothing. for V9 `viewSpa` behavior, pass through a Just as `toUrl`)
  - add custom attributes hole to the Option (in order to make SegmentedControls compatible with the Modal component)
  - combine `css` attributes into one to prevent class-name-order-change css :bug:s
  - :bug: fix overflowing-y svg icon issue

### Overflow/stickiness demo

overflow

cc @NoRedInk/design 

(@krinhorn  -- this might be a good opportunity for 🍐 on accessibility tools!)


src/Nri/Ui/SegmentedControl/V10.elm
```diff
@@ -0,0 +1,305 @@
+module Nri.Ui.SegmentedControl.V10 exposing
+    ( Option, view
+    , SelectOption, viewSelect
+    , Width(..)
+    )
+
+{-| Changes from V9:
+
+  - hides non-displayed content rather than fully removing from the DOM, allowing for the segmented control to be used as a sticky-header that doesn't lose scroll position
```
BrianHicks, line 9: maybe this is a nitpick, but I haven't seen "sticky" used this way before. Wouldn't we say something is "scroll-preserving" if we do this? To me, sticky means that it will stick in it's position on the screen (like `position: sticky`.)

src/Nri/Ui/SegmentedControl/V10.elm
```diff
@@ -0,0 +1,305 @@
...
+  - :skull: removes NavConfig and SelectConfig
```
BrianHicks, line 10: *definitely* a nitpick, but if you want this  to show up as 💀 it needs to be the literal emoji, as the package site doesn't parse Slack-style emoji names.

src/Nri/Ui/SegmentedControl/V10.elm
```diff
@@ -0,0 +1,305 @@
...
+    | FillContainer
+
+
+{-| -}
+type alias SelectOption value msg =
+    { value : value
+    , label : String
+    , attributes : List (Attribute msg)
+    , icon : Maybe Svg
+    }
```
BrianHicks, line 53: how would you feel about inlining this for better error messages? It'd make the signature of `viewSelect` pretty complex (but maybe being explicit about the complexity is fine?)

src/Nri/Ui/SegmentedControl/V10.elm
```diff
@@ -0,0 +1,305 @@
...
+
+
+{-| -}
+type alias Option value msg =
+    { value : value
+    , label : String
+    , attributes : List (Attribute msg)
+    , icon : Maybe Svg
+    , content : Html msg
+    }
```
BrianHicks, line 113: same comment about inlining here.

src/Nri/Ui/SegmentedControl/V10.elm
tesk9: I somewhat believe that users may want to use the `SelectOption` and `Option` types. We're using it in the Example currently.. and for the Tab component, I definitely see people using the `Tab` type.

BrianHicks: seems fine!

PR #570, opened 2020-08-05
Axe/segmented control and tabs

tesk9:
@krinhorn and I paired on accessibility checking for Segmented Controls, and we did find some issues.

This PR fixes what we found:
- tabpanels weren't focusable, meaning voiceover wouldn't read the content
- the main segmented controls view and the tabs component followed 2 totally different UX patterns for keyboard navigability, despite fulfilling the same function.
- the secondary viewSelect segmented controls function used aria attributes to emulate radio buttons, but only part of the way (keyboard behavior was quite different than radio buttons). Underlying implementation is now actually radio buttons

## Segmented Control
segmented control keyboard

## Tabs
tabs keyboard

### DIFF
--- a/src/Nri/Ui/SegmentedControl/V11.elm
+++ b/src/Nri/Ui/SegmentedControl/V11.elm
@@ -4,14 +4,12 @@ module Nri.Ui.SegmentedControl.V11 exposing
     , Width(..)
     )
 
-{-| Changes from V9:
+{-| Changes from V10:
 
-  - hides non-displayed content rather than fully removing from the DOM, allowing for the content the SegmentedControl controls to have overflowY: auto & maintain scroll position
-  - 💀 removes NavConfig and SelectConfig
-  - combines `view` and `viewSpa` (for V9 `view` behavior, be sure `toUrl` is Nothing. for V9 `viewSpa` behavior, pass through a Just as `toUrl`)
-  - add custom attributes hole to the Option (in order to make SegmentedControls compatible with the Modal component)
-  - combine `css` attributes into one to prevent class-name-order-change css :bug:s
-  - :bug: fix overflowing-y svg icon issue
+  - change selection using left/right arrow keys
+  - only currently-selected or first control is tabbable
+  - `viewSelect` renamed to `viewRadioGroup`, `SelectOption` renamed to `Radio`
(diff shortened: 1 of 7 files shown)


BrianHicks:
😮 this sounds like a really interesting pairing session. Are y'all planning to do a demo on how you found these things out?

src/TabsInternal.elm
```diff
@@ -0,0 +1,205 @@
+module TabsInternal exposing (Config, Tab, views)
```
BrianHicks, line 1: thinking in terms of atomic design, why aren't the things in this module an atom? Are they just harder to use correctly so we wrap them in preset ways?

src/Nri/Ui/SegmentedControl/V11.elm
```diff
@@ -0,0 +1,268 @@
...
+    )
+
+{-| Changes from V10:
+
+  - change selection using left/right arrow keys
+  - only currently-selected or first control is tabbable
+  - tabpanel is tabbable
+  - Uses TabsInternal under the hood
+  - `viewSelect` renamed to `viewRadioGroup`, `SelectOption` renamed to `Radio`
+  - `viewRadioGroup` uses native HTML radio input internally
```
BrianHicks, line 14: thank you for making these goals so clear!

src/Nri/Ui/SegmentedControl/V11.elm
```diff
@@ -60,14 +60,14 @@ not a page control
   - `width`: how to size the segmented control
 
 -}
-viewSelect :
+viewRadioGroup :
```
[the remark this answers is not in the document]
tesk9: We checked the monorepo, and we're often using the `viewSelect` helper when it would be better to use `view`. I want users of this function to know what they're getting -- if they wouldn't use radio buttons, they shouldn't user viewRadioGroup

src/TabsInternal.elm
tesk9: From my perspective, we wrap them in order to have our exposed components match the Designer's styleguide. I don't know that we'll want another tab component apart from SegmentedControls and Tabs? I suppose we may someday...

tesk9:
Re demo -- isn't the zoom link for the PST meeting still broke? 😅 Happy to demo when the meeting re-occurs

PR #581, opened 2020-08-13
add checked attribute to SegmentedControl radio view

BrianHicks:
# ⚠️ don't merge until #579 is in `master`.

this adds `checked=true` to segmented controls so we can have a programmatic way of determining which thing is selected that doesn't depend on class names, etc. QA needs it to un-break their tests!

### DIFF
--- a/src/Nri/Ui/SegmentedControl/V11.elm
+++ b/src/Nri/Ui/SegmentedControl/V11.elm
@@ -29,6 +29,7 @@ import EventExtras
 import Html.Styled
 import Html.Styled.Attributes as Attributes exposing (css, href)
 import Html.Styled.Events as Events
+import Json.Encode as Encode
 import Nri.Ui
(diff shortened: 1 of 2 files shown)


BrianHicks:
wait, that's really weird… I assumed that it wasn't set because when I use SegmentedControl I see this in the DOM:

image

for this rendering:

image

any ideas why `checked` would be totally absent there?

BrianHicks:
🤔 looks like that's also true in the deploy preview here, so the additional work in this PR isn't actually adding anything

BrianHicks:
@tesk9 I paired with Kristine this morning and it's annoyingly hard to make Cypress grab a checked/not-checked value, so I'm gonna add a `data-nri-checked` attribute to get around testing limitations. This shouldn't make the control less accessible; just more testable.

All that to say: ready for another review!

PR #589, opened 2020-08-21
use text underlines instead of bottom-borders for multiline buttons

BrianHicks:
Sometimes `ClickableText` gets line-wrapped. When that happens, the hover styles don't look right:

| Screen Shot 2020-07-07 at 2 25 38 PM | Screen Shot 2020-07-07 at 2 23 04 PM | Screen Shot 2020-07-07 at 2 24 48 PM
|-|-|-|

the way to fix this is to use `text-decoration: underline` instead of `border-bottom: 1px solid $azure`. This looks fine for both single-line and multiple-line buttons:

image

image

… but it also has some other effects. Namely, it creates a conflict in our text styles, since most of the things in `Nri.Ui.Text.V4` set a blue `border-bottom` on links. When things are inline, this works OK, but it means that we shouldn't combine `ClickableText` with `Text`, which seems silly (and, in fact, the style guide has an example of this…)

image

So, while this PR (currently) fixes the multiline case, I think more changes are needed to make underlined links work consistently (although it'd be reasonable… changing `border-bottom` to a `text-decoration` in `Nri.Ui.Text.V4` and the other places it needs to match.)

### DIFF
--- a/src/Nri/Ui/ClickableText/V3.elm
+++ b/src/Nri/Ui/ClickableText/V3.elm
@@ -16,7 +16,7 @@ module Nri.Ui.ClickableText.V3 exposing
 
   - uses ClickableAttributes
   - adds `css` helper
-  - add bottom border on hover instead of text decoration
+  - removes underline on hover and recolors to azureDark
 
 
 # Changes from V2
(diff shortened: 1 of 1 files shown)


BrianHicks:
I guess that'd work, yeah!

BrianHicks:
alright, I've pushed a commit that does that. Thanks!

BrianHicks:
done!

BrianHicks:
mmm, makes sense… I didn't know that had caused a headache before. I'll make that change and ship it.

PR #593, opened 2020-09-02
Raven/modal improvements

tesk9:
- Adds new Modal version with API and focus management improvements
- Changes the example to use Debug.Control rather than a bajillion buttons and Msgs
- Adds KeyboardSupport notes to the example
- Hides the Skip to Content Button when it's not focused
- adds `custom` helper for adding arbitrary html attributes (primarily useful to make limiting the scope of selectors in tests easier by adding ids to modals)
- tab and tabback events stop propagation and prevent default 
- Refactors so that the ids don't overwrite each other -- this pushes responsibility for managing focus on to engineers a bit more. Introduces new FocusTrap module. (Hopefully this will be useful for banners like the one for writing connection loss errors)

Screen Shot 2020-09-02 at 2 19 48 PM

hide skip button

### Known Issues:

With focus in the URL bar (CMD+L), hitting tab takes you under the overlay. I'm tempted to further mess with the tabIndex, unless people have other ideas...?

~~The firstFocusable/lastFocusable/etc attributes overwrite the id of the element to which they're attached. Is there a better alternative here..?~~

Sometimes a modal's outcomes will impact what buttons appear on the page (including whether the button that launched it is present or not). E.g., "Add" button disappears if the user saves from the modal, but doesn't disappear if the user cancels from the modal.

### DIFF
--- a/src/Nri/Ui/Modal/V11.elm
+++ b/src/Nri/Ui/Modal/V11.elm
@@ -13,41 +13,152 @@ module Nri.Ui.Modal.V11 exposing
   - change `open`, `close` to return `(Model, Cmd Msg)` rather than `Msg`
   - make info and warning themes
 
+```
+import Browser exposing (Program, element)
+import Html.Styled exposing (..)
+import Html.Styled.Attributes exposing (id)
+import Html.Styled.Events as Events
+import Nri.Ui.Modal.V11 as Modal
+
+main : Program flags model msg
+main =
+    { init = \_ -> init
+    , view = view
+    , update = update
+    , subscriptions = \model -> Modal.subscriptions model.modalState
+    }
+
+type ModalKind
+    = FirstKindOfModal
+    | SecondKindOfModal
+
+type alias Model =
+    { modal : ModalKind
+    , modalState : Modal.Model
+    }
+
+init : Model
+init =
+    let
+        ( modalState, cmd ) =
+            -- When we load the page with a modal already open, we should return
+            -- the focus someplace sensible when the modal closes.
+            -- [This article](https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/) recommends
+            -- focusing the main or body.
+            Modal.open "maincontent"
+    in
+    ( { modal = FirstKindOfModal
+      , modalState = modalState
+      }
+    , Cmd.map ModalMsg cmd
+    )
+
+type Msg
+    = OpenModal ModalKind String
+    | ModalMsg Modal.Msg
+    | CloseModal
+
+update : Msg -> Model -> ( Modal, Cmd Msg )
+update msg model =
+    case msg of
+        OpenModal modalKind returnFocusTo ->
+            let
+                ( modalState, cmd ) =
+                    Modal.open returnFocusTo
+            in
+            ( { modal = modalKind
+              , modalState = modalState
+              }
+            , Cmd.map ModalMsg cmd
+            )
+
+        ModalMsg modalMsg ->
+            let
+                ( modalState, cmd ) =
+                    Modal.update
+                        { dismissOnEscAndOverlayClick = True }
+                        modalMsg
+                        model.modalState
+            in
+            ( { model | modalState = modalState }
+            , Cmd.map ModalMsg cmd
+            )
+
+        CloseModal ->
+            let
+                ( modalState, cmd ) =
+                    Modal.close model.modalState
+            in
+            ( { model | modalState = modalState }
+            , Cmd.map ModalMsg cmd
+            )
+
+view : Model -> Html Msg
+view model =
+    main_ [ id "maincontent" ]
+        [ button
+            [ id "open-first-kind-of-modal-button"
+            , Events.onClick (OpenModal FirstKindOfModal "open-first-kind-of-modal-button")
+            ]
+            [ text "Open FirstKindOfModal" ]
+        , button
+            [ id "open-second-kind-of-modal-button"
+            , Events.onClick (OpenModal SecondKindOfModal "open-second-kind-of-modal-button")
+            ]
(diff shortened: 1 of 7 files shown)


src/Nri/Ui/Modal/V11.elm
```diff
@@ -116,20 +132,19 @@ view model =
...
-                                    [ closeButton firstFocusableElement
-                                    , text "Modal Content"
-                                    ]
-                                , footer =
-                                    [ button
-                                        (Events.onClick CloseModal :: lastFocusableElement)
-                                        [ text "Close" ]
-                                    ]
-                                }
+                    , content =
```
juanedi, line 135: I like the change! I liked how the previous API was a bit more strict but this one is definitely more ergonomic!

juanedi: 💭 I wonder if something like the following would be feasable as a compromise between both:

```elm
Modal.view
  {...
  -- no need to use firstFocusable and friends in the view. just set the ids.
  , content : []
  , footer = []
  }
  [ -- set the focus manager as an attribute, separated from the actual content
    Modal.singleFocusableElement "element-id"
  , -- alternative for multiple focusable elements
    Modal.multipleFocusableElements
      { firstFocusableElement = "first-focusable-element"
      , lastFocusableElement = "last-focusable-element"
      }
  ]
```

Of course, the user could still use it in a way that the supplied ids point nowhere in the view, but I like that it's a bit more explicit about how focus management works, and there's no room for the user to mess up using the multiple helpers in an inconsistent way.

I guess we could implement that by adding a single event handler that captures key presses and prevents defaults only on certain key codes and target element ids. Would that make sense?


(Not asking you to follow this path but curious if it's something you considered and if it makes sense at all).

juanedi: Also... if we did this I guess we could move the close button configuration to an attribute instead of expecting the user to call `closeButton` function in the view and know that it will be placed using absolute positioning.

juanedi: 🗒️ I wrote all of that before seeing the `FocusTrap` change. I think doing something among this lines might still make sense! Like.. the focus trap abstraction/module could be in charge of configuring this single event handler and that way it's even more decoupled from the view (assuming the ids are valid).

juanedi (commented):
This is looking good! I'm curious about your thoughts on that alternative I proposed below.

Also, I didn't get to look into the CMD+L issue... will do tomorrow. Sorry!

src/Nri/Ui/Modal/V11.elm
tesk9: I'm interested but not sure I'm following. When/how would we attach the tab and tab back listeners...?

Would you like to pair on this? Or if not, want to just do it? :) I like the API suggestions, as I understand them -- just not sure how to make them a reality

src/Nri/Ui/Modal/V11.elm
```diff
@@ -438,7 +439,7 @@ viewModal config styles =
                     toContentAndFooter
                         { onlyFocusableElement =
                             List.map (Attrs.map config.wrapMsg)
-                                [ Key.onKeyDown
+                                [ onKeyDownPreventDefault
```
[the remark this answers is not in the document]
tesk9: I was seeing some weird behavior on the add-a-passage branch where the wrapping would correctly take me to the titleEditor, but then something would cause my focus to move to the URL bar. (Which, since focus then incorrectly moves under the modal rather than on to it, really breaks the experience). I'm not actually sure that this is the fix though, and I'm happy to walk it back and do more verification work before making a change like htis

src/Nri/Ui/Modal/V11.elm
juanedi: I'll do a quick spike to see if it's actually doable and maybe ask you to pair on the API if it works!

juanedi: hmmm... it looks like it can't be done in pure Elm.

i was thinking about adding a single attribute to the top level container with something like:

```elm
multipleFocusableElements : { firstId : String, lastId: String, focus : String -> msg } -> Attribute msg
multipleFocusableElements {firstId, lastId, focus } =
  -- intention: use capture instead of bubble to run our event handler
  -- before the native focus change has occurred
  VirtualDom.onCapture "keydown"
  (VirtualDom.Custom
    (Decode.map3
      (\elementId keyCode shiftKey ->
          case (elementId, keyCode, shiftKey) of
            (firstId, tabKeyCode, True) ->
              { message : focus lastId
              , preventDefault : True
              , stopPropagation : True
              }
            (lastId, tabKeyCode, False) ->
              { message : focus firstId
              , preventDefault : True
              , stopPropagation : True
              }
            _ ->
              Decode.fail "..."
      )
      (Decode.at ["target", "id"] Decode.string)
      (Decode.field "key" Decode.int)
      (Decode.field "shiftKey" Decode.bool)
    )
  )
```

But the [VirtualDom docs](https://package.elm-lang.org/packages/elm/virtual-dom/latest/VirtualDom#on) say there is no support for capture phase events, and I don't think there's another way to prevent default tab behavior by attaching an event handler to a container element :-(

src/Nri/Ui/Modal/V11.elm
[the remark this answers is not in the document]
juanedi: I don't think this change will hurt so I'm fine with merging as it is. Maybe doing the symlink dance to verify locally before doing the release dance can save us from some more dancing in the future, but up to you!

juanedi (approved):

src/Nri/Ui/Modal/V11.elm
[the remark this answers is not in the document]
tesk9: Thanks!

PR #597, opened 2020-09-04
Changes to FocusTrap API

juanedi:
We now implement focus trap using a single event handler in the widget's top level container (instead of attaching handlers to the focusable elements as we did before).

This should simplify the API a bit, at the expense of assuming that the user will use valid ids when defining the focus trap.


src/Nri/Ui/Modal/V11.elm
```diff
@@ -319,13 +322,25 @@ warning =
...
+
+
 {-| This is the default setting.
 -}
 showTitle : Attribute
 showTitle =
     Attribute (\attrs -> { attrs | visibleTitle = True })
 
 
+focusTrap : FocusTrap -> Attribute
```
juanedi, line 339: 💭 @tesk9 how would you feel about making the focus trap a required part of the config instead of an attribute of the modal? I think I liked how the FocusManager forced me to think about focus management for the modal.

juanedi: I did it! Let me know if you disagree though, we can totally revert.

src/Nri/Ui/FocusTrap/V1.elm
```diff
@@ -20,34 +19,68 @@ import Json.Decode as Decode exposing (Decoder)
...
+            else if elementId == lastId && not shiftKey then
+                -- if the user tabs forward while on the last id,
+                -- we want to wrap around to the first id.
+                Decode.succeed
+                    { message = focus firstId
+
+                    -- Will this preventDefault break anything
+                    -- if the element is an input
+                    -- or dropdown (if the element has key-based
+                    -- behavior already)?
```
juanedi, line 59: I'm pretty sure it won't break anything because:
  - preventDefault should only affect whether or not the native behavior runs, but it shouldn't mess up with event propagation
  - we are handling the event as it is bubbling up, and at this point event handlers on descendant elements should have already been fired

With that said, testing locally before merging shouldn't hurt. I'll try it out!

juanedi: I just added a text input to the example in the style guide and verified that we can still attach event handlers to it as usual. All seems to be working fine!

tesk9 (approved):

PR #599, opened 2020-09-09
Tessa/percy screenshots per widget

tesk9:
Change the Percy screenshots to be per doodad/widget/dealio instead of per category.


juanedi (approved):
Niiiiiice!

PR #600, opened 2020-09-10
don't shrink icons in messages in narrow parents

BrianHicks:
Icons were being smooshed in messages. Now they're not!

## Before

image

## After

image


tesk9 (approved):
Thanks for fixing!

PR #602, opened 2020-09-10
Make the X button work when dismissOnEscAndOverlayClick=False

juanedi:
Before this we were ignoring `CloseModal` events when the value was set to False.


tesk9 (approved):
WOO 🎸 🎸 🎸

PR #603, opened 2020-09-10
Bump 12.2.2

juanedi:
Patch to fix a bug in our Modal's close button.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "NoRedInk/noredink-ui",
     "summary": "UI Widgets we use at NRI",
     "license": "BSD-3-Clause",
-    "version": "12.2.2",
+    "version": "12.3.0",
     "exposed-modules": [
         "Nri.Ui",
         "Nri.Ui.Accordion.V1",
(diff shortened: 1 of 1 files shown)


tesk9 (commented):
@juanedi I think there's some accidentally-committed files in here -- this is meant to just be elm.json, right?

juanedi:
ouchhhhh! fixing 😅

src/ModalExample.elm
```diff
@@ -0,0 +1,116 @@
+module ModalExample exposing (..)
```
tesk9, line 1: I don't think we want this file either, do we...?

juanedi: not sure what's wrong! git kept re-adding that file. should be fixed now.

tesk9 (approved):
Woo!

tesk9:
Rebased on master, so this will now do the following:

```
This is a MINOR change.

    Nri.Ui.SegmentedControl.V13
    Nri.Ui.Tabs.V7
    Nri.Ui.Tooltip.V2
```

tesk9:
Oh wait but then that means this is a minor change rather than a patch

tesk9:
Published!

PR #604, opened 2020-09-11
Fab/accordion v2

tesk9:
- Combine view and viewKeyed so that nodes are always keyed
  - Removes viewCaret from the API -- it's possible to use the DisclosureIndicator directly
  - Changed implementation to follow recommendations from <https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html>
      - Adds Up and Down keyboard handling for navigating between the accordions
      - Adds aria-expanded and aria-controls
      - Changes accordion container to a section
      - removes tablist/tab/tabpanel roles
      - Adds heading levels for the accordion header (including style resets)

accordion

### DIFF
--- a/styleguide-app/Examples/Accordion.elm
+++ b/styleguide-app/Examples/Accordion.elm
@@ -52,7 +52,10 @@ view model =
             ]
                 |> List.map
                     (\entry ->
-                        ( entry, Dict.get entry.id model |> Maybe.withDefault False )
+                        ( "accordion-entry__" ++ String.fromInt entry.id
+                        , entry
+                        , Dict.get entry.id model |> Maybe.withDefault False
+                        )
                     )
         , viewHeader = .title >> Html.text
         , viewContent = \{ content } -> Text.smallBody [ Html.text content ]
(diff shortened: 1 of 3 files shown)


tesk9:
Ping, @rtfeldman :)

styleguide-app/Examples/Accordion.elm
```diff
@@ -143,5 +164,13 @@ type alias State =
...
-    ( Dict.insert id toExpanded model, Cmd.none )
+update msg model =
+    case msg of
+        Toggle id toExpanded ->
+            ( Dict.insert id toExpanded model, Cmd.none )
+
+        Focus id ->
+            ( model, Task.attempt Focused (Dom.focus id) )
+
+        Focused _ ->
```
rtfeldman, line 175: 🎨 If we're never going to use this `Result`, I'd personally rather leave it out of the data model.

In that world, instead of `Task.attempt Focused (Dom.focus id)` it would be:

```
Dom.focus id
    |> Task.onError (\_ -> Task.succeed FocusAttempted)
    |> Task.perform (\_ -> FocusAttempted)
```

…which is more verbose, but I'd pay that price for the simpler `| FocusAttempted` variant in `Msg`.

Up to you though!

rtfeldman (approved):
Looks FABulous! Two 🎨 s, but feel free to merge as-is if you like.

styleguide-app/Examples/Accordion.elm
tesk9: My hope is that eventually, in our non-example code, we can report when focusing on an element by id fails. I think someday, when we're fully on modal v11, that will tell us when focus traps are broken/etc. Right now it would be very noisy tho 😅

PR #607, opened 2020-09-14
stop icons from shrinking and top-align them

BrianHicks:
More icons were shrinking (see https://github.com/NoRedInk/NoRedInk/issues/29526) and Stacey and Ben prefer they be top-aligned (see https://github.com/NoRedInk/NoRedInk/issues/29527.)

This PR fixes both!

image

image

image

image

@NoRedInk/design does that look ok to y'all?

### DIFF
--- a/src/Nri/Ui/Message/V2.elm
+++ b/src/Nri/Ui/Message/V2.elm
@@ -112,7 +112,7 @@ view attributes_ =
                     "Nri-Ui-Message--tiny"
                     [ displayFlex
                     , justifyContent start
-                    , alignItems center
+                    , alignItems flexStart
                     , paddingTop (px 6)
                     , paddingBottom (px 8)
                     , fontSize (px 13)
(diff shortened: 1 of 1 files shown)


bendansby:
Can you post a single-line example too?

BrianHicks:
oh, good catch… that looks worse. I'll see if I can get the best of both worlds real quick, mm?

image

BrianHicks:
There. Now there's no change.

image

Oh… while I'm in here, do you want the icons in the large and banner-style warnings to be aligned top too? Now they look like so:

image

BrianHicks:
that is, would it be better for them to look like this?

image

staceyadams:
I like it in the middle in the large and banner styles because they are in a container.
I don't feel strongly either way though.

bendansby:
Yeah, agreed

BrianHicks:
works for me. No further work needed to keep them centered.

If y'all wanna check it out further the deploy preview is in the deploy/netlify check on this PR, currently https://deploy-preview-607--noredink-ui.netlify.app

juanedi (approved):

PR #613, opened 2020-09-21
Button font size

bendansby:
Tweaks the font size of medium buttons so they match other similarly-sized widgets like segmented controls, selects, etc. 

image

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -776,7 +776,7 @@ sizeStyle size width =
                     }
 
                 Medium ->
-                    { fontSize = 17
+                    { fontSize = 15
                     , height = 45
                     , imageHeight = 15
                     , shadowHeight = 3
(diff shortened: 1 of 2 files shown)


tesk9 (commented):
Looks nice, but I'm a little worried that this is going to cause regressions on the monolith.

Is there a difference between ClickableText.small and ClickableText.medium anymore? Since Clickable Text is sometimes used in-line, I could see this leading to mis-matched font sizes in a paragraph

bendansby:
ClickableText medium already didn't match Nri.Ui.Text mediumBody, though it was closer, so it's possible any mismatches are harder to spot. I'll go ahead and undo the CT changes though, just in case, as the the button change is more salient. Thanks!

tesk9 (approved):
Sweet, works for me! I'm gonna throw a note into the frontend ops changes channel so folks know to add this to their qa instructions if they release a new version.

PR #614, opened 2020-09-28
Kraken/clickable svg with border

tesk9:
for clickable svgs:
- adds withBorder setting 
- adds primary, secondary, and danger themes

Screen Shot 2020-09-28 at 12 28 28 PM

Questions:
- "secondary" is currently the navy color -- is that what secondary should mean here?
- do we want background color changes on hover?

### DIFF
--- a/src/Nri/Ui/ClickableSvg/V1.elm
+++ b/src/Nri/Ui/ClickableSvg/V1.elm
@@ -445,7 +445,7 @@ buttonOrLinkStyles config =
                         ( Colors.navy, Colors.navy, Css.pointer )
 
                     Danger ->
-                        ( Colors.red, Colors.red, Css.pointer )
+                        ( Colors.red, Colors.redDark, Css.pointer )
     in
     [ Css.property "transition"
         "background-color 0.2s, color 0.2s, border-width 0s, border-color 0.2s"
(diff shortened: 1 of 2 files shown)


bendansby:
Woo-hoo! I think primary and secondary could match the corresponding button styles, so white icon on azure button and azure icon on white button with azure border.

Maybe a px or two more of padding? The eye icon looks good as is but some of the other icons look a little big within their confines.

tesk9:
@bendansby sounds good! I wonder about the `danger` behavior then -- if we're matching the `Button` behavior, the interior would be red rather than white. Maybe `dangerInverted`? Or something like that...?

bendansby:
Yup, that makes sense. We originally had a "danger secondary" button style in Sketch but it never ended up getting used.

tesk9:
Oh I like `dangerSecondary` 👍 thanks!

tesk9:
Screen Shot 2020-09-28 at 2 58 11 PM

Is redLight okay for the hover?

bendansby:
Yup! Love the grid!

allymcknight (commented):
Noice

allymcknight (approved):

PR #615, opened 2020-09-28
Adds couple for coteacher icon

tesk9:
Adds co-teach icon (now called `couple`, open to suggestions on alternate names!) to UiIcons so that we can re-add the icon to https://github.com/NoRedInk/NoRedInk/pull/29742.

Screen Shot 2020-09-28 at 3 18 40 PM

cc @bendansby


bendansby:
Nice!

Arkham (approved):
Sweet!

Issue #598, opened 2020-09-09
modals should not expand proportional to their surroundings


BrianHicks:
modals currently expand vertically on super tall viewports. They shouldn't do that!

![image (3)](https://user-images.githubusercontent.com/355401/92607730-2215b000-f27a-11ea-899b-c37560727a02.png)

staceyadams:
The expanding is even more obvious/egregious on modals that only have a single line of text in the body:
![image](https://user-images.githubusercontent.com/10527520/92623247-d44d7780-f27a-11ea-8c0e-573036c921ed.png)


The modals are still usable though and it's unlikely that users have viewports this tall so historically, this hasn't been prioritized.

PR #618, opened 2020-10-01
remove/lower min heights

bendansby:
This removes the minimum height for the modal container and reduces it for the modal content area so that modals with a small amount of content don't have a bunch of empty space and modals within a huge viewport don't grow for no good reason. Also tweaks footer spacing. I'm hoping it's OK to keep this on V11 since it's just styling, but let me know.

Outfoxes: https://github.com/NoRedInk/noredink-ui/issues/598

image

image

image


BrianHicks:
looks good to me!

PR #629, opened 2020-10-14
Fix bottom button alignment

tesk9:
Screen Shot 2020-10-14 at 2 47 42 PM


Arkham:
hey @tesk9 did you catch up with ben about this? I saw that in some designs the clickable test is inline with the rest of the text even when the text is underscored. In our implementation, adding the underscore makes the text jump up slightly instead.

tesk9:
@Arkham yes! Ben and I paired on this

Arkham (approved):
Awesome, let's get them changes in :)

PR #633, opened 2020-10-21
Kraken/clickable svg improvements

tesk9:
Screen Shot 2020-10-21 at 10 42 55 AM

Ookkkkkkkkayyyyy I made a new version of ClickableSvg with a different API for width/height. Instead of always having to do `ClickableSvg.height` and `ClickableSvg.width`, there are preset sizes (`small`, `medium`, `large`), and then you can set the width of the button to be something exact with `exactWidth` if you want.

@NoRedInk/design I'm not sure what the presets should actually be! I picked some values somewhat at random. Would love instructions or commits! 🙏

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -15,6 +15,7 @@
         "Nri.Ui.Callout.V1",
         "Nri.Ui.Checkbox.V5",
         "Nri.Ui.ClickableSvg.V1",
+        "Nri.Ui.ClickableSvg.V2",
         "Nri.Ui.ClickableText.V3",
         "Nri.Ui.Colors.Extra",
         "Nri.Ui.Colors.V1",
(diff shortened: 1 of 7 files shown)


bendansby:
Nice! Let's have the height and borders match Nri.Ui.Button

staceyadams:
Those danger eyes feel very spoooooky and Halloweeny!

tesk9:
With the same padding & border per size styles as from Button:

Screen Shot 2020-10-21 at 11 56 04 AM
Screen Shot 2020-10-21 at 11 54 21 AM
Screen Shot 2020-10-21 at 11 54 15 AM
Screen Shot 2020-10-21 at 11 54 10 AM
Screen Shot 2020-10-21 at 11 54 02 AM

bendansby:
Just pushed some padding tweaks!

bendansby:
Other thing I noticed is that the bottom borders for the primary and danger themes are white.

tesk9:
> Other thing I noticed is that the bottom borders for the primary and danger themes are white.

Yeeech not great. Good eye. Should the bottom borders be transparent or match the background color?

bendansby:
Let's make it match Button, so for the bottom borders

primary = azureDark
secondary = azure
danger = redDark
dangerSecondary = red

Arkham (approved):
Looks purrfect!

PR #637, opened 2020-10-26
bounded-width button

BrianHicks:
# 🚧 depends on #636 

I needed a button with a `min-width` instead of a `width`. This adds that, and caps how much the button can expand at the same time!

bounded-width-button

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -5,7 +5,7 @@ module Nri.Ui.Button.V10 exposing
     , onClick
     , href, linkSpa, linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
     , small, medium, large
-    , exactWidth, unboundedWidth, fillContainerWidth
+    , exactWidth, boundedWidth, unboundedWidth, fillContainerWidth
     , primary, secondary, danger, premium
     , enabled, unfulfilled, disabled, error, loading, success
     , delete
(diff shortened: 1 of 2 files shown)


BrianHicks:
@NoRedInk/design does the padding on these when they're above the minimum width look OK to you?

Also, here's my motivation:

image

bendansby:
Sorry, could you explain the problem and solution more step by step?

BrianHicks:
The problem: we have some the error text above as a replacement for the "Continue" button on the assignment form during validation. But it line breaks and just kind of looks bad. I could set both buttons to be 190px wide but then it looks meh when the labels are small (cf. "Preview" below.)

image

So what I want is actually to let the validation message grow the button a little, but within a bound. This PR lets me say "the button should be 150px wide, growing up to 200px if necessary for the contents of the label.

BrianHicks:
original context where Stacey and I talked about this: https://github.com/NoRedInk/NoRedInk/pull/30315

bendansby:
Gotcha. Not that more flexibility with buttons ain't cool, but in this case could you alternately set the Continue button to be unbounded and set the Preview button to the exact width of the Continue button in its default state?

bendansby:
I.e. would that solve the problem and if so would that avoid any complication that might arise from modifying the button component? (maybe there's not really any complication, just wanted to provide some checks and balances)

BrianHicks:
That'd make the preview button exactly 150px and the unbounded one a minimum of 75px (the default min-width for unbounded.) We still need the button to be a reasonable width when it says "Continue" instead of an error message, so I want to be able to set the minimum.

But TBH I hadn't considered it at all before now so thanks for the checks slash balances!

bendansby:
Ah, right. Cool, sounds good!

juanedi (approved):

PR #639, opened 2020-10-28
port build system to shake

BrianHicks:


script/run-axe.sh
```diff
@@ -3,6 +3,10 @@ set -euo pipefail
 
 # start a web server in the background and tear it down when exiting
 ./script/serve.sh public &
+
+# Wait for the python server to launch since we're launching it async
+sleep 3
```
BrianHicks, line 8: how would you feel about waiting on `lsof` or `curl` to test the port instead of waiting a set amount of time? Better or worse?

omnibs: For sure better, I was lazy!

BrianHicks:
OMG IT PASSED THANK YOU @omnibs @rtfeldman

BrianHicks:
actually, since y'all have looked at this a little… I wonder if you'd mind reviewing it too? I'll request your reviews, please feel free to decline if it's not great right now.

script/netlify.sh
```diff
@@ -6,8 +6,8 @@ set -xeuo pipefail
 # https://github.com/netlify/build-image) but for now it's way simpler to just
 # accept that Netlify will need separate instructions.
 
-# get our dependencies
-npm install
+# get our dependencies (--ignore-scripts=false is needed for puppeteer)
+npm install --ignore-scripts=false
```
BrianHicks, line 10: why do we need puppeteer to build for Netlify? I think it's just `elm make` and `browserify`.

omnibs:
> I wonder if you'd mind reviewing it too?

Stopped for the day but can look into it tomorrow!

BrianHicks:
cool cool cool. I'm going to restart the build in the meantime to get a sense of if caching /nix is worthwhile (it sometimes is and sometimes isn't on Travis.)

script/develop.sh
```diff
@@ -32,4 +34,4 @@ cleanup() {
 trap cleanup EXIT INT
 
 # start a watcher. This loops forever, so we don't need to loop ourselves.
-find src styleguide-app -type f -not -ipath '*elm-stuff*' | entr -c -p make public
+git ls-files | entr -c -p shake --compact "$SHAKE_TARGET"
```
omnibs, line 37: For some reason I'm getting `Shakefile.hs: getCurrentDirectory:getWorkingDirectory: resource exhausted (Too many open files)` when running shake with entr (and it's only with entr) and when setting `shakeThreads = 0,` or any number higher than 13 in my system.

With 14 even a Shakefile that does basically nothing fails with that error.

I'm gonna open an issue about it in the shake repo.

BrianHicks: What's your ulimit like? Maybe `git ls-files` has juuuuuuuust enough to bump you close to the limit. `entr` opens a ton of handles.

omnibs:
```
$ ulimit
unlimited
$ ulimit -s
8192
$ ulimit -u
11136
```

Does it seem low? The repo is so small, `git ls-files` has 174 files.

omnibs: Oh, it's:

```
$ ulimit -n
256
```

Raising it to like 1k fixes it. Still, how do we manage to watch stuff in the monorepo without having to resort to this?

BrianHicks: I think that's the only way to do it. If `entr` fails with that it links to a help page.

Maybe `find` produced fewer files though? 🤷

omnibs: Oh, I'm going to guess it's bc we don't use `entr`, but `watchman`, from this review of watchers: https://anarc.at/blog/2019-11-20-file-monitoring-tools/#entr

omnibs: So 💭 either set a high `ulimit -n` in this file (1k?) or ditch `entr`?

BrianHicks: ok, switched to `watchexec` in 6236588. How's that work for you?

script/develop.sh
```diff
@@ -32,4 +32,4 @@ cleanup() {
 trap cleanup EXIT INT
 
 # start a watcher. This loops forever, so we don't need to loop ourselves.
-find src styleguide-app -type f -not -ipath '*elm-stuff*' | entr -c -p make public
+exec watchexec --clear --postpone -- shake --compact "$SHAKE_TARGET"
```
omnibs, line 35: `exec` here makes the trap above never hit. I was really confused why but [this SO](https://serverfault.com/questions/803247/parent-bash-script-not-receiving-trap-despite-process-still-running) made it clearer for me, *I think*:

> The trap command only works as long as the script is running.

I think the reason is `exec` makes `watchexec` the running process and `ctrl+c` is sent to it, instead of the script.

Was there a specific reason for adding `exec` here? I usually only do it if the launched process has some trouble without it (e.g. docker doesn't clean up on ctrl+c unless we `exec` it when doing `docker run`, I think).

I'm gonna go ahead and remove it, but feel free to add it back in case it's necessary!

BrianHicks: oh duh, sorry… I'm just used to adding `exec` in front of the last command if the point of the script is to set something up. I totally forgot about the teardown. Yes, you've got the reason right—`exec` makes the invoked command take over the whole process, so the bash `trap` never gets called.

omnibs:
(oops sorry for leaving the debug echo hahah)

omnibs (approved):
This looks great! Sorry for the long review time!

BrianHicks:
no worries, and thank you for the review!

PR #643, opened 2020-11-06
FAB/checkbox

tesk9:
FAB Friday checkbox accessibility improvements!

# Change set 1: 

These changes came from FAB Friday on [November 6, 2020](https://paper.dropbox.com/doc/FAB-Fridays-log-and-upcoming-topics--A_WVewOq7TOzbe8LpdckKeyaAg-Yr853teky2oypzanPfSKP#:uid=434012004652331945797748&h2=2020-11-06
)

- When focusing on the “checkbox” element, VoiceOver should recognize it as a checkbox and should instruct you to VO-space bar to select it rather than saying it's a group
- tabbing off the checkbox element, you shouldn't hear VoiceOver reading the label a second time

Screen Shot 2020-11-06 at 2 55 31 PM

Caveat: I don't think our approach for hiding focus indicators for mouse users will work for checkboxes anymore.

cc @NoRedInk/design 

# Change set 2: 

These changes came from FAB Friday on (Friday 😱) [November 13, 2020](https://paper.dropbox.com/doc/FAB-Fridays-log-and-upcoming-topics--A_WVewOq7TOzbe8LpdckKeyaAg-Yr853teky2oypzanPfSKP#:uid=912008681740817975415360&h2=2020-11-13)

Disabled checkboxes should still be read aloud by voiceover.

### DIFF
--- a/src/Nri/Ui/Checkbox/V5.elm
+++ b/src/Nri/Ui/Checkbox/V5.elm
@@ -303,11 +303,12 @@ viewIcon styles icon =
             , border3 (px 2) solid transparent
             , padding (px 2)
             , borderRadius (px 3)
-            , Css.batch styles
             ]
         , Attributes.class "checkbox-icon-container"
         ]
-        [ Nri.Ui.Svg.V1.toHtml icon
+        [ Html.div [ css [ backgroundColor Colors.white ] ]
+            [ Nri.Ui.Svg.V1.toHtml (Nri.Ui.Svg.V1.withCss styles icon)
+            ]
         ]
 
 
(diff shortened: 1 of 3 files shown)


bendansby:
Would be great if we could solve the focus outline problem. Can be very distracting.

tesk9:
I'll try to fix the focus behavior on the monolith side!

tesk9:
Ping, @Arkham :) (we made some more changes in FAB Friday today and it was actually quite convenient that this hadn't merged yet! 😄)

Arkham:
Ah, good timing! :)

I've noticed that when focusing a checkbox pressing Space does select and unselect it, but pressing Enter doesn't do anything, is that a regression?

tesk9:
It's a change in behavior for sure! I was actually just wrong about what behavior checkboxes are supposed to have, I think? 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox is actually only interact-able with space, not enter. I changed the docs.

Thanks for fixing the specs!

Arkham (approved):
Looking fab!!

Issue #573, opened 2020-08-07
Radio button loses user focus


tesk9:
![radio keyboard nav](https://user-images.githubusercontent.com/8811312/89664023-5c5eec80-d88b-11ea-8591-cf64556a07ed.gif)

Make sure that there's a visual indication of the user's focus. https://www.deque.com/blog/accessible-focus-indicators/

PR #644, opened 2020-11-06
Show an outline for radio buttons on selection

tesk9:
Don't lose focus on radio buttons & use aria-disabled instead of disabled attribute.

Fixes https://github.com/NoRedInk/noredink-ui/issues/573

Uses same approach for showing outlines as https://github.com/NoRedInk/noredink-ui/pull/643

NOTE: this changes the tabindex behavior! We were overriding tab behavior in kind of a weird way: radio buttons within a group should be navigable with the arrow keys, and not internally tab-able. I also had to remove an event prevent-default event swallower to stop breaking the default radio behavior.


allymcknight (approved):
whoops sorry this got lost in the review pile. Looks good!

PR #648, opened 2020-11-13
allow overriding the generated ID in a text input

BrianHicks:
Allow overriding automatically-generated IDs in `TextInput.V6`. This will result in a minor version bump when released.

(this should help solve https://github.com/NoRedInk/NoRedInk/issues/29383)


juanedi (approved):
Looks good!

PR #657, opened 2020-12-08
use cachix

BrianHicks:
Could potentially speed up builds a lot! Even if we have to pull stuff over the network, at least we won't be building `niv` and `elm-forbid-import` except when we update them.

| Step | Time |
|-|-|
| cold cachix cache, no travis caching (but pushing to cachix) | 13m6s |
| warm cachix cache, no travis caching | 3m13s |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: nix
-sudo: false
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,20 +11,45 @@ cache:
     - ~/.elm
 
 install:
-  # set up cachix
-  # see https://docs.cachix.org/continuous-integration-setup/travis-ci.html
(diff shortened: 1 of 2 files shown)


BrianHicks:
I'm not sure why it says it was waiting… maybe github didn't get the webhook event. Restarting CI to get it to be all green.

juanedi:
tbh I don't think I understand exactly what's going on here to review 😅 
i'm leaving this one for our next pairing session!

BrianHicks:
oh! In that case, let me annotate the CI file so that the next person who needs to touch it understands what's going on as well.

BrianHicks:
we can still totally review it while pairing, though 😄

juanedi (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #659, opened 2020-12-10
add Nri.Ui.Switch.V1

BrianHicks:
We finally have a reason to add switches! Hooray!

These switches look nice, have nice animations, and should be reasonably accessible (at least as far as I can tell.) The animation also stops if the user-agent says the user prefers reduced motion—I'm not sure what the threshold for those things is, so I just turned off animation altogether if it's disabled. Apple seems to just do color fades instead, so I may be being too conservative here.

Here's how keyboard and mouse interaction goes:

switch-interactions

### DIFF
--- a/src/Nri/Ui/Switch/V1.elm
+++ b/src/Nri/Ui/Switch/V1.elm
@@ -100,6 +100,13 @@ view attrs isOn =
                 [ Global.descendants
                     [ Global.svg [ Css.borderColor Colors.azure ] ]
                 ]
+            , Css.cursor
+                (if config.onSwitch == Nothing then
+                    Css.default
+
+                 else
+                    Css.pointer
+                )
             ]
         , Aria.controls config.id
         , Widget.checked (Just isOn)
@@ -113,7 +120,6 @@ view attrs isOn =
             (viewSwitch
                 { id = config.id
                 , isOn = isOn
-                , enabled = config.onSwitch /= Nothing
                 }
             )
         , case config.label of
@@ -159,7 +165,6 @@ viewCheckbox config =
 viewSwitch :
     { id : String
     , isOn : Bool
-    , enabled : Bool
     }
     -> Svg
 viewSwitch config =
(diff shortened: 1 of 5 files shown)


BrianHicks:
@staceyadams how's this look? You can poke it at https://deploy-preview-659--noredink-ui.netlify.app/#/doodad/Switch

staceyadams:
Looking spiffy!

Can you add the inset shadows to the switch background (idk what it's called; the thing it's sliding on)? The same styling is used on checkboxes and radio buttons.

staceyadams:
p.s. I love being able to poke at it on the share link!

staceyadams:
Does the disabled state need to be focusable? Is that so the label can be read? The focus ring feels a little misleading in this case since you can't move the switch.

BrianHicks:
> Can you add the inset shadows to the switch background (idk what it's called; the thing it's sliding on)? The same styling is used on checkboxes and radio buttons.

Oh, oops! That looks like a bug. It should be there.

> Does the disabled state need to be focusable? Is that so the label can be read? The focus ring feels a little misleading in this case since you can't move the switch.

I agree. I'm not sure why this needs to happen but checkbox does it so I copied it here. @tesk9 do you know?

BrianHicks:
ok, shadow's fixed. Same preview link!

staceyadams:
Woo!

One more small thing: border on the knob of the off state should be gray75/#bfbfbf

BrianHicks:
biffed!

tesk9:
> Does the disabled state need to be focusable? Is that so the label can be read? The focus ring feels a little misleading in this case since you can't move the switch.

I'm so excited you asked this question!!! FAB Friday just learned about this, and we are in the process of implementing across noredink-ui. (cc @krinhorn !!)

We determined that _yes_, we want disabled inputs to be focusable. Additionally, we want to use [`aria-disabled`](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-disabled) instead of `disabled` so that screenreader users will know (a) that there's a disabled slider doohickey on the page and (b) what the value is disabled as.

I tested the deploy preview with VoiceOver, and it sounded good to me!

src/Nri/Ui/Switch/V1.elm
```diff
@@ -293,11 +296,18 @@ viewSwitch config =
...
                         ]
                     ]
                     []
                 ]
             ]
         ]
         |> Nri.Ui.Svg.V1.fromHtml
+
+
+transition : String -> Css.Style
```
juanedi, line 328: 🤘

juanedi (approved):

krinhorn:
Yay! Thanks, Brian!

> The animation also stops if the user-agent says the user prefers reduced motion—I'm not sure what the threshold for those things is

I think as little animation as possible without losing required information. If the animation doesn't convey anything meaningful, then best to reduce it as close to no animation as we can.

krinhorn:
image
image
image


From https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html, where there are some examples of essential and non-essential animation and a bit more context. :)

PR #666, opened 2021-01-05
Use same shape of star filled for star outline

Arkham:
Adjust the starOutline so it's not cut off on the edges:

### Before:
image

### After:

image

cc @NoRedInk/design


tesk9 (approved):
Woo! Will merge & publish a new version once there's a 👍 from design 😈

bendansby:
lgtm 👍

PR #669, opened 2021-01-19
Kraken/message v3

tesk9:
Add Message.V3 in order to address feedback here https://github.com/NoRedInk/NoRedInk/pull/31399#discussion_r560306476.

Primary goal was to add `css` helper that would allow us to round top corners and keep bottom corners square on a banner Message.

Screen Shot 2021-01-19 at 1 15 38 PM


Arkham (approved):
Looks great! I like that now we can override the default icon of a theme :)

PR #670, opened 2021-01-22
Raven/reduced motion support 3

juanedi:
Don't render confetti if `prefers-reduced-motion` is set!

This won't avoid the computation, but I figured it was better to do it all using css only to keep the widget port-free.

See https://paper.dropbox.com/doc/Reduce-motion-for-site-animations--BDuE8Mq0HRjcjs3jjW_4O4SrAg-UQ2ngQW9PORW1SMa1qapL


tesk9 (approved):
Woo works great! Confetti is beautiful when it shows, and very absent when it doesn't show 👍

Thanks!

Issue #665, opened 2020-12-23
Fix labels on Nri.Ui.Switch so it plays well with elm-program-test


juanedi:
`Switch` is backed by a checkbox input. The usual way to interact with those in a `ProgramTest` is via the `check` function, but this is what you get when you try to use it with a switch:

```
    ▼ Query.has [ text "HTML expected by the call to: check "Apply this scale across all rubric criteria"" ]
    
    ✗ has text "HTML expected by the call to: check "Apply this scale across all rubric criteria""
    
    Expected one of the following to exist and have an "onchange" handler:
    - <label for="nri-ui-switch-with-default-id"> with text "Apply this scale across all rubric criteria" and an <input id="nri-ui-switch-with-default-id">
    - <input aria-label="Apply this scale across all rubric criteria" id="nri-ui-switch-with-default-id">
```

We should probably do that!

tesk9:
@juanedi this looks like it's accepted in the Raven project. Can this be closed?

juanedi:
Yeeeees! 🤘

PR #673, opened 2021-02-02
add "for" attribute to the switch's label

juanedi:
this will associate the label with the input, giving us some nice things such as:
  - screen readers will read the label when the input is focused
  - clicking on the label toggles the switch
  - toggling the switch with elm-program-test is now possible without manually simulating dom events

For https://github.com/NoRedInk/noredink-ui/issues/665

NOTE: I tested that this plays well with elm-program-test by symlinking on the monolith's `CustomizeRubricSpec`


Arkham (approved):
Thank you!

PR #676, opened 2021-02-09
Increase label left padding on checkboxes

Arkham:
From https://github.com/NoRedInk/NoRedInk/issues/31315#issuecomment-776149088

Left old, right new.

Screenshot 2021-02-09 at 18 51 08


tesk9 (approved):
Looks lovely! Fingers crossed we don't rely too much on the checkbox width anywhere 😄

PR #683, opened 2021-04-06
allow tooltips in segmented controls

BrianHicks:
does what it says on the tin!

image

### DIFF
--- a/styleguide-app/Examples/SegmentedControl.elm
+++ b/styleguide-app/Examples/SegmentedControl.elm
@@ -21,7 +21,7 @@ import Html.Styled.Attributes as Attributes exposing (css)
 import Html.Styled.Events as Events
 import KeyboardSupport exposing (Direction(..), Key(..))
 import Nri.Ui.Colors.V1 as Colors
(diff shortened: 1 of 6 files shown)


styleguide-app/Examples/SegmentedControl.elm
```diff
@@ -160,18 +161,33 @@ buildRadioOptions content =
             , label = label
             , value = value
             , idString = String.fromInt value
+            , tooltip =
+                [ Tooltip.plaintext text
+                , Tooltip.open (currentlyHovered == Just value)
+                , Tooltip.fitToContent
+                , Tooltip.onHover
```
nunocf, line 168: I've noticed that in most places, we're using `Tooltip.smallPadding`. Perhaps it's worth checking if we need that?
we use it on the `sidebar tabs` and the `expand/collapse buttons`

nunocf (approved):

nunocf:
I tried getting the best example of a tooltip.
Sorry!

nunocf:
@NoRedInkBot r+

PR #685, opened 2021-04-07
Fix ClickableSvg example

tesk9:
ClickableSvg tooltip example should print "You clicked the preview button!" when the button is clicked.


nunocf (approved):
Oh, this is nice!

Thank you for looking into this, I think using the same example should work on the PR I found this bug on!

nunocf:
@NoRedInkBot r+

PR #688, opened 2021-04-12
SegmentedControl honours FillContainer

juanedi:
`viewRadioGroup` was ignoring it when tooltips were being used.

**NOTE: ** this is a bug, but no use cases of `viewRadioGroup` were actually using tooltips (since they were recently introduced for a new feature).

The following combination should now result in full-width controls!

Screen Shot 2021-04-12 at 18 17 24


nunocf (approved):

juanedi:
The output of the CI build looks fine, and it seems like integration with GH has been failing for a long time. I'm merging this and will add a story to figure out what's going on here.

Screen Shot 2021-04-13 at 18 51 08

PR #689, opened 2021-04-13
bump to 12.18.1

juanedi:
- Patches `SegmentedControl.viewRadioGroup` to honour `FillContainer` when tooltips are being used (#688)
- Fixes `ClickableSvg` example so that the content is actually clickable when tooltips are being used (#685)


tesk9 (approved):
Looks good! (Also I won't tell anybody if you don't if you want to do a merge-sans-review on this kind of change 🙊)

PR #694, opened 2021-05-14
Kraken/http message

tesk9:
Update the Message component to support taking Http.Errors directly & producing nicely formatted, user-friendly error views.

You can play with this experience [here](https://deploy-preview-694--noredink-ui.netlify.app/#/doodad/Message). cc @NoRedInk/design 

# Bad Url, Bad Status, Bad Body

Screen Shot 2021-05-14 at 3 00 44 PM

# 404

Screen Shot 2021-05-14 at 3 01 30 PM

# 401

Screen Shot 2021-05-14 at 3 01 20 PM

# Network Error

Screen Shot 2021-05-14 at 3 01 12 PM

# Timeout

Screen Shot 2021-05-14 at 3 00 54 PM

### DIFF
--- a/src/Nri/Ui/Message/V3.elm
+++ b/src/Nri/Ui/Message/V3.elm
@@ -182,6 +182,12 @@ view attributes_ =
 
                     Just msg ->
                         largeDismissButton msg
+                , case attributes.codeDetails of
+                    Just details ->
+                        viewCodeDetails details
+
+                    Nothing ->
+                        text ""
                 ]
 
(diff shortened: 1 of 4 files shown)


bcardiff (approved):
Looks good, the only thing I notice is that when expanding the errors the layout may change. In NRI pages this will not be a problem, but I am not sure.

3GJ1B7JvHw

tesk9:
I don't think it will be a problem, thank you!

PR #696, opened 2021-05-20
Raven/tooltip without tail

juanedi:
Adds the option to render a tooltip without a tail!

The API change is backwards compatible so this should be a painless upgrade 👍 

https://user-images.githubusercontent.com/753421/119056178-a17a8800-b9a0-11eb-9115-d04c22a87d82.mov


nunocf (approved):
Very nice!

PR #697, opened 2021-05-25
Raven/disabled tabs

juanedi:
Adds the ability to disable tabs to `Nri.Ui.Tabs.V7`.

Disabled tabs can't be selected and also can't be focused with the keyboard using the left/right arrow keys.

This is needed for https://github.com/NoRedInk/NoRedInk/issues/33574. We'll use this so that keypresses will skip to focusing on the tab we want users to focus instead of whatever tab comes after the current one.

### Preview

https://user-images.githubusercontent.com/753421/119571177-6b198000-bd87-11eb-8fd9-b7facd335274.mov

### DIFF
--- a/src/Nri/Ui/Tabs/V7.elm
+++ b/src/Nri/Ui/Tabs/V7.elm
@@ -2,9 +2,9 @@ module Nri.Ui.Tabs.V7 exposing
     ( view
     , Alignment(..)
     , Tab, Attribute, build
-    , tabString, tabHtml, withTooltip
-    , panelHtml
+    , tabString, tabHtml, withTooltip, disabled
     , spaHref
+    , panelHtml
     )
 
 {-| Changes from V6:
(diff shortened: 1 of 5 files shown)


src/Nri/Ui/Tabs/V7.elm
```diff
@@ -16,8 +16,8 @@ module Nri.Ui.Tabs.V7 exposing
 @docs view
 @docs Alignment
 @docs Tab, Attribute, build
-@docs tabString, tabHtml, withTooltip
-@docs panelHtml
+@docs tabString, tabHtml, withTooltip, disabled
+@docs th@docs panelHtml
```
BrianHicks, line 20: 🐛 little typo here that's gonna break publishing

BrianHicks (changes requested):
looks pretty good! We need to fix the `th@docs` issue I commented on, and CI is also red. Once those are good, we're OK to merge!

PR #698, opened 2021-05-26
Added labelledBy attribute to tabs

nunocf:
This PR adds the ability to override existing tab tooltips with external tooltips by setting a `labelledBy` attribute, making external tooltips still accessible.


elm.json
```diff
@@ -3,7 +3,7 @@
     "name": "NoRedInk/noredink-ui",
     "summary": "UI Widgets we use at NRI",
     "license": "BSD-3-Clause",
-    "version": "12.20.0",
```
BrianHicks, line 6: would you mind reverting this? We don't bump versions until the change is on the main branch, just so we can bundle a bunch of stuff together without getting conflicting bumps

BrianHicks (changes requested):
looks good so far! Once CI is green I think we're ready to go!

elm.json
nunocf: Oh sorry! This was not supposed to be here!

nunocf: we're already on 12.21.0, which is odd!

nunocf: I've just removed this as it was an unintended change.
Thank you!

BrianHicks:
green! merging!

PR #715, opened 2021-08-20
Raven/remove modal shadows

micahhahn:
Remove shadows from modal dialogs as they were escaping corners on Safari.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -2,7 +2,7 @@
     "elm-forbid-import": {
         "branch": "main",
         "repo": "https://git.bytes.zone/brian/elm-forbid-import.git",
-        "rev": "16b31af7f6f5902f6c61dcab46224595c5dcf8e7",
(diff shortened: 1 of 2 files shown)


micahhahn (commented):
🍐 with @BrianHicks on these changes.

BrianHicks (changes requested):
looks like an elm-format failure. As soon as you fix that and CI passes, I think we're good to go!

BrianHicks:
hmm, looks like we need to update some other deps too. I'll have a look at that on Monday!

BrianHicks:
argh, I have no idea why it's doing this. One more travis retry. If it doesn't work we'll ask in #eng-frontend for help.

BrianHicks:
OK, so it's always failing the integration with `master`, but always passing the build itself 🤔

BrianHicks:
except… those *should* be the same thing? Very weird.

PR #720, opened 2021-08-30
Menu.V3

waj:
This introduces a new version of the `Menu` component adopting the attribute pattern. It also adds an option to customize the `z-index` used for the menu. This is required to fix some issues we have at NRI with zIndex layering.

### DIFF
--- a/styleguide-app/Examples/Menu.elm
+++ b/styleguide-app/Examples/Menu.elm
@@ -17,7 +17,7 @@ import KeyboardSupport exposing (Direction(..), Key(..))
 import Nri.Ui.ClickableSvg.V2 as ClickableSvg
 import Nri.Ui.ClickableText.V3 as ClickableText
 import Nri.Ui.Heading.V2 as Heading
-import Nri.Ui.Menu.V2 as Menu
+import Nri.Ui.Menu.V3 as Menu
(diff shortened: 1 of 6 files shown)


styleguide-app/Examples/Menu.elm
```diff
@@ -120,27 +122,30 @@ view state =
...
-            }
-            { title = "1st Period English with Mx. Trainer"
-            , icon = viewConfiguration.icon
-            , hasBorder = viewConfiguration.hasBorder
-            , wrapping = viewConfiguration.wrapping
-            , buttonWidth = viewConfiguration.buttonWidth
+            , button =
+                Menu.button
+                    ([ Menu.hasBorder viewConfiguration.hasBorder
+                     , Menu.wrapping viewConfiguration.wrapping
```
stoeffel, line 128: 🎨 How about adding `Menu.emptyAttr` so that we can use existing "generic" maybe functions or casing?
`Maybe.withDefault Menu.emptyAttr viewConfiguration.icon` inside the list of attributes
?

stoeffel: oh sorry, I thought that `withOptionalAttribute` was in the API of `Menu`.

stoeffel:
published with 13.2.0

PR #724, opened 2021-09-01
Modal: add footer background and border

bendansby:
This adds a background color and top border to the V11 modal footer to give it some structure after we removed the shadow. I checked several modals in the monolith with the new styles and didn't see a need for a new version.

image


tesk9 (approved):
Are you okay with how it looks with short content? If so, looks good to me!

image

bendansby:
Yeah you know, it's kind of a two tone thing, very stylish. Thanks!

PR #725, opened 2021-09-01
Add border radius to checkbox

bendansby:
So little white corners don't show when the background is a color (like when a topic is selected in the Assignment Library)

image


tesk9 (approved):
Woo wayyy nicer thank you!

PR #726, opened 2021-09-02
Make toggleButton clickable in style guide

LearningNerd:
Related to [this story in Pivotal Tracker, "Grades buttons change position/height when pressed"](https://www.pivotaltracker.com/story/show/178325503) -- wanted to check if this bug would appear in the style guide (just ruled it out), and thought this would be a nice small improvement.

Now you can click on the toggle buttons under "Button toggle" to preview how they change state visually: 
image

Thank you @tesk9 for pairing with me on this!

### DIFF
--- a/styleguide-app/Examples/Button.elm
+++ b/styleguide-app/Examples/Button.elm
@@ -80,12 +80,14 @@ update msg state =
             ( state, Cmd.none )
 
         ToggleToggleButton id ->
-            ( { state | pressedToggleButtons =
-                if Set.member id state.pressedToggleButtons then
-                    Set.remove id state.pressedToggleButtons
-                else
-                    Set.insert id state.pressedToggleButtons
-                }
+            ( { state
+                | pressedToggleButtons =
+                    if Set.member id state.pressedToggleButtons then
+                        Set.remove id state.pressedToggleButtons
+
+                    else
+                        Set.insert id state.pressedToggleButtons
+              }
             , Cmd.none
             )
 
--- a/styleguide-app/Examples/Button.elm
+++ b/styleguide-app/Examples/Button.elm
@@ -59,6 +59,7 @@ type ButtonType
 type Msg
     = SetDebugControlsState (Control Model)
     | ShowItWorked String String
+    | ToggleToggleButton Int
     | NoOp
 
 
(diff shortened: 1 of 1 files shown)


bendansby (approved):
Lovely!

LearningNerd:
I take it back about ruling this out -- I was able to replicate the bug inside the style guide by changing the containing `div` to `display: block` (it was currently using `display:flexbox`):

image
image

I assume we want these toggle buttons to look right regardless of what kind of container they're used inside, so here's my proposed solution to fix the on-click visual jitter: set the toggle button's `vertical-align` property to _anything other than_ `baseline`. That fixes it.

@bendansby any thoughts on this? Specifically, is `vertical-align: bottom` a safe default to use instead of `baseline`? Or maybe `middle` is better?

bendansby:
`middle` seems like a sensible choice! I also have an alternate idea for your consideration:

I see that our buttons are set to `display: inline-flex` and setting them to `display: flex` fixes the problem. To me that seems too opinionated for a style guide primitive. I would opt for just having them be `display: flex` and let their containers out in the wild determine their layout. 

But I'm assuming we have a lot of buttons out there relying on their inline-flexibility, so what do you think about making just the toggle buttons `display: flex` and adjusting their containers as needed, since as you mentioned it's not a heavily used subcomponent?

Either solution works for me!

bendansby (approved):
Actually thinking on it more, I like your solution better, since, if we want any kind of button to be inline-flex, it's probably the toggle buttons.

styleguide-app/Examples/Button.elm
```diff
@@ -78,6 +79,16 @@ update msg state =
             in
             ( state, Cmd.none )
 
+        ToggleToggleButton id ->
+            ( { state | pressedToggleButtons =
+                if Set.member id state.pressedToggleButtons then
+                    Set.remove id state.pressedToggleButtons
+                else
+                    Set.insert id state.pressedToggleButtons
```
ap-nri: I could have sworn there was a helper for this in Set.Extra or Sort.Set but nope.

ap-nri: Oh wait there is! this is totally fine as is though. https://package.elm-lang.org/packages/stoeffel/set-extra/latest/Set-Extra#toggle

ap-nri (approved):
'88 jordan leapin' from the free throw

PR #727, opened 2021-09-04
Input padding

bendansby:
I noticed that a checkbox was not aligned on the assignment form, and then I checked some radio buttons, and same thing, and then once you see it, you can't unsee it everywhere, so, gotta fix it.

Napkin 2 09-04-21, 11 12 08 AM


styleguide-app/Examples/RadioButton.elm
```diff
@@ -35,7 +35,7 @@ example =
     , update = update
     , subscriptions = subscriptions
     , view = view
-    , categories = [ Layout ]
+    , categories = [ Inputs ]
```
tesk9, line 38: 😍

styleguide-app/Examples/RadioButton.elm
```diff
@@ -85,7 +85,7 @@ view model =
 
 viewVanilla : State -> Html Msg
 viewVanilla state =
-    div [ css [ Css.margin (Css.px 8) ] ]
+    div []
```
tesk9, line 88: Should the `viewPremium` margin also be cleared out?

src/Nri/Ui/Checkbox/V5.elm
```diff
@@ -300,9 +300,8 @@ viewIcon styles icon =
         [ css
             [ position absolute
             , left zero
+            , paddingTop (px 4)
             , top (calc (pct 50) minus (px 18))
-            , border3 (px 2) solid transparent
```
tesk9, line 304: These styles were to account for the faked focus outline for keyboard users. Without the padding being even and without the transparent border, the focus styles looks kinda funky:

image

src/Nri/Ui/RadioButton/V2.elm
```diff
@@ -177,7 +178,7 @@ internalView config =
                 , ( "Nri-RadioButton-RadioButtonChecked", isChecked )
                 ]
             , css
-                [ padding4 (px 4) zero (px 4) (px 40)
+                [ padding4 zero zero zero (px 35)
```
tesk9, line 181: We're seeing a similar problem with the focus ring as the checkbox. Now, you can't see the focus ring for radio buttons at all.

tesk9 (commented):
@LearningNerd and I paired on reviewing and we noticed that the style changes cause regressions to the focus rings. We'd offer you a suggestion but it definitely seems like a tricky puzzle!

bendansby:
[Chesterton's fence](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence) strikes again!

bendansby:
What do you think of this solution on checkbox?

bendansby:
Ugh, I just realized my solution only shows for keyboard input in Safari but shows on click in Chrome.

bendansby:
@LearningNerd @tesk9 OK, trying a different tack! How's that look?

tesk9:
Looks really nice! 😻 

When we use the radio buttons and checkboxes, we'll definitely want to be careful about spacing them out so the focus doesn't overlap:
Screen Shot 2021-09-09 at 5 09 56 PM
But I doubt we're smooshing the inputs close enough together that we'll end up with any problems. Looks like it'll work great!

I'm out of time to review the code part today. Liz, feel free to review, approve, & merge without me! If you don't get to it, I'll come back to it when I can.

LearningNerd:
> "When we use the radio buttons and checkboxes, we'll definitely want to be careful about spacing them out so the focus doesn't overlap [...] But I doubt we're smooshing the inputs close enough together that we'll end up with any problems"
-- @tesk9 

Hmm, I'm not clear on the convention for how things appear in the style guide versus how they appear "in the wild" inside NRI. For example, do we often use radio buttons side by side? If I were to add radio buttons to a page, would it then just be up to me to add some additional CSS to them or their parent container to space them out horizontally in a way I think looks good? 

@bendansby I don't know if it's easy to add more default spacing to the radio buttons if they're side by side, but that'd be my only other suggestion -- _or_ some guidance from the style guide on how I should use them / space them out would be super helpful! :)

bendansby:
Good question! We do have some side by side radio buttons on the site. In this example, padding has been added to separate each one. So that need was already present. The style guide could definitely be a lot more helpful in terms of recommending real world usage, though, I hear that!

image

LearningNerd:
Cool, thanks for clarifying @bendansby! Sounds like a topic for future updates / discussion though, I don't want to derail this thread lol. I think we can call this one done, calling in the bot 🤖 🎉   @NoRedInkBot +r

bendansby:
I think r+ only works on the monolith? Let's see:

@NoRedInkBot r+

bendansby:
Yup looks like. Thanks for reviewing!

PR #728, opened 2021-09-07
Delete the now unused callout component

bendansby:


BrianHicks:
this will create a major version bump. That's probably OK, but we should talk with folks in #eng-frontend to see if there's anything else we need to batch with that!

This change, by itself, looks fine! We just need to remove the module name from `elm.json` before it will pass.

bendansby:
When a new version is created, don't we/doesn't that just bring along all the stuff since the previous version?

BrianHicks:
yep, but we try to batch major version releases because they can create situations where we need to remove code. I know you removed all the relevant code first this time, but there may be other stuff and syncing on it is probably wise! I'll ask around.

bendansby:
OK, thanks!

LearningNerd:
Thanks for posting on Slack @BrianHicks! Just posted my tiny tiny update to the toggle button UI in https://github.com/NoRedInk/noredink-ui/pull/729, maybe we can include that in the update.

PR #733, opened 2021-09-28
Tessa/add balloon v1

tesk9:
Adds Balloon component to noredink-ui.

image

Made some slight changes to the API to make the `Attribute`s of the Balloon concrete by removing type holes..

There are some weird behaviors when using a percentage width. These were pre-existing. The next version of the Balloon should fix.


Arkham:
LGTM :)

PR #734, opened 2021-09-28
Tessa/add search input

tesk9:
I started looking into adding a new SearchBar component (based off of Nri/SearchBar), but it seemed the existing monolith component was not accessible. So instead, I've extended the TextInput module to include a `search` type.

Visually, the TextInput.search should match the Nri.SearchBar pretty closely. The TextInput version will _not_ keep track of searchIterations, as the Nri.SearchBar version did.

search text input

cc @NoRedInk/design


bendansby:
> The TextInput version will not keep track of searchIterations

what does this mean?

tesk9:
> > The TextInput version will not keep track of searchIterations

> what does this mean?

For a reason that's unclear to me, Nri.SearchBar keeps track of how may times the search has been reset, and calls this number `searchIter`. I think there may have been a bug in the reset behavior at some time, since the searchIter value is being used for a keyed node?

In any event -- I mentioned this here in the interest of explicitly noting differences between TextInput.search and the monolith equivalent. I don't think there's any kind of user-facing impact or metrics impact.

PR #735, opened 2021-09-29
Dansby/mobile modal

bendansby:
Some tweaks for modals on mobile devices.

IMG_2896
IMG_2897

### DIFF
--- a/src/Nri/Ui/Modal/V11.elm
+++ b/src/Nri/Ui/Modal/V11.elm
@@ -438,7 +438,7 @@ modalStyles =
     , width (px 600)
     , backgroundColor Colors.white
 
-    -- the modal should grow up to the viewport minus a 50px margin
+    -- the modal should grow up to the viewport minus a 20px margin
     , maxHeight (calc (pct 100) minus (px 40))
     ]
 
--- a/src/Nri/Ui/Modal/V11.elm
+++ b/src/Nri/Ui/Modal/V11.elm
@@ -166,6 +166,7 @@ import Browser
 import Browser.Dom as Dom
 import Browser.Events
 import Css exposing (..)
+import Css.Media
 import Css.Transitions
 import Html.Styled as Root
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/Modal/V11.elm
```diff
@@ -425,16 +426,20 @@ modalStyles =
...
     , borderRadius (px 20)
     , boxShadow5 zero (px 1) (px 10) zero (rgba 0 0 0 0.35)
+    , Css.Media.withMedia
+        [ Css.Media.all [ Css.Media.maxWidth (px 1000) ] ]
+        [ borderRadius zero
+        ]
 
     -- Spacing
-    , margin2 (px 50) auto
+    , margin2 (px 20) auto
```
bendansby, line 435: This is intended for both mobile and desktop.

src/Nri/Ui/Modal/V11.elm
```diff
@@ -425,16 +426,20 @@ modalStyles =
...
 
     -- Size
     , width (px 600)
     , backgroundColor Colors.white
 
     -- the modal should grow up to the viewport minus a 50px margin
```
tesk9, line 441: Maybe we should update this comment too?

tesk9 (commented):
Looks good! The chrome mobile previewer was not very helpful but I looked at it on my actual phone (thanks, netlify previews!) and it's definitely a big improvement!

One thing on the margin2 mobile & desktop change -- I think this allows for some content overlap:

### Before
Screen Shot 2021-09-29 at 11 40 45 AM

### On this PR
Screen Shot 2021-09-29 at 11 40 20 AM

src/Nri/Ui/Modal/V11.elm
bendansby: Good catch, updated.

bendansby:
Thanks! I think I'm OK with that overlap. Will see how it feels once it's in the site and make adjustments as necessary.

PR #736, opened 2021-10-03
Message tweaks

bendansby:
Especially mobile styles for the banner size.

Screen Shot 2021-10-03 at 11 34 52 AM
Screen Shot 2021-10-03 at 11 35 03 AM
Napkin 2 10-03-21, 12 17 26 PM

### DIFF
--- a/src/Nri/Ui/Message/V3.elm
+++ b/src/Nri/Ui/Message/V3.elm
@@ -162,6 +162,7 @@ view attributes_ =
                              borderRadius (px 8)
                            , padding (px 20)
                            , backgroundColor_
+                           , position relative
                            ]
                     )
                  ]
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/Message/V3.elm
```diff
@@ -775,33 +800,6 @@ getRoleAttribute role =
 
 
 
--- Style overrides
```
bendansby, line 778: I think this was only added for the login page so I will mop up the spill.

tesk9 (commented):
Should the "large" banner dismiss x also have less right margin when the viewport is small?

Screen Shot 2021-10-04 at 10 02 19 AM

---

This looks good to me! I think I want to do a noredink-ui release before this merges, and right after it merges, so that some of the other changes can go out without needing to worry about Message style regressions (since this PR kindly removes the styleOverrides business 🙏 ).

bendansby:
> Should the "large" banner dismiss x also have less right margin when the viewport is small?

Yes, because I positioned it absolutely so the text would have more room so I scooted it into the padding area to eliminate text overlap.

bendansby:
>I think I want to do a noredink-ui release before this merges
👍

tesk9:
Hmm, it looks to me like the "large" Message's x _isn't_ scooted and the "banner" Message's x _is_ scooted?

At any rate, feel free to merge when you're ready! (I released a noredink-ui version a few minutes ago)

bendansby:
Good catch, unified the scoots.

PR #738, opened 2021-10-04
Add mobile query

tesk9:
Move Nri.MediaQuery from the monolith.

### DIFF
--- a/src/Nri/Ui/MediaQuery/V1.elm
+++ b/src/Nri/Ui/MediaQuery/V1.elm
@@ -1,13 +1,13 @@
 module Nri.Ui.MediaQuery.V1 exposing
     ( mobile, notMobile
-    , teacherStudentPage
+    , quizEngineMobile
     )
 
 {-| Standard media queries for NRI responsive pages.
 Can be used to create a Snippet using Css.Global.media
 
 @docs mobile, notMobile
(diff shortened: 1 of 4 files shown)


src/Nri/Ui/MediaQuery/V1.elm
```diff
@@ -0,0 +1,52 @@
...
+`minWidth (px 1)` is for a bug in IE which causes the media query to initially trigger regardless of window size
+
+See: <http://stackoverflow.com/questions/25673707/ie11-triggers-css-transition-on-page-load-when-non-applied-media-query-exists/25850649#25850649>
+
+-}
+teacherStudentPage : MediaQuery
+teacherStudentPage =
+    only screen
+        [ minWidth (px 1)
+        , maxWidth (px 700)
```
tesk9, line 51: @bendansby is the 700px breakpoint deprecated, or do we still want it sometimes..?

bendansby: We still want it, but it should be renamed to quizEngineAssignment or something

lindsaykwardell (approved):
Looks good to me!

Not sure where to find Gifs to put here, so image the perfect Gif is playing below 🌠

PR #739, opened 2021-10-05
Dismiss button tweaks

bendansby:
Some more tweaks to Modal and Message to improve the dismiss experience

Modal:
- Touch target extends all the way around x
- Translucent background and z-index ensure x won't be covered up or obscured by lesson content on mobile

Message
- Dismiss button no longer increases height of large and banner
- Padding is now even on large and banner on mobile


tesk9 (approved):
Looks good to me, thank you!

PR #761, opened 2021-10-28
Tessa/use control extras

tesk9:
Use `Debug.Control.Extra` list in many of the styleguide examples to make adding new attributes/settings less involved.

Updates styleguide example for:
- Balloon
- Message
- Modal
- Tooltip

### DIFF
--- a/styleguide-app/Examples/Balloon.elm
+++ b/styleguide-app/Examples/Balloon.elm
@@ -35,26 +35,25 @@ example =
 
 {-| -}
 type alias State =
-    Control Settings
-
-
-type alias Settings =
-    { copy : String
-    , theme : Maybe ( String, Balloon.Attribute )
-    , position : Maybe ( String, Balloon.Attribute )
-    , width : Maybe ( String, Balloon.Attribute )
-    , padding : Maybe ( String, Balloon.Attribute )
+    { copy : Control String
+    , attributes : Control (List ( String, Balloon.Attribute ))
     }
 
 
-init : Control Settings
+init : State
 init =
-    Control.record Settings
-        |> Control.field "copy" (Control.string "Hello, world!")
-        |> Control.field "theme" (Control.maybe False themeOptions)
-        |> Control.field "position" (Control.maybe False positionOptions)
-        |> Control.field "width" (Control.maybe False widthOptions)
-        |> Control.field "padding" (Control.maybe False paddingOptions)
+    { copy = Control.string "Hello, world!"
+    , attributes = controlAttributes
+    }
+
+
+controlAttributes : Control (List ( String, Balloon.Attribute ))
+controlAttributes =
+    ControlExtra.list
+        |> ControlExtra.optionalListItem "theme" themeOptions
+        |> ControlExtra.optionalListItem "position" positionOptions
+        |> ControlExtra.optionalListItem "width" widthOptions
+        |> ControlExtra.optionalListItem "padding" paddingOptions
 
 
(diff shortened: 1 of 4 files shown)


styleguide-app/Examples/Tooltip.elm
```diff
@@ -206,9 +197,19 @@ controlContent =
...
+        (\bool ->
+            if True then
+                Tooltip.withoutTail
+
+            else
+                -- TODO: change `withoutTail` to take
+                -- a bool or expose a `withTail` from Tooltip.
+                Tooltip.css []
+        )
+        (Control.bool False)
```
LearningNerd, line 212: Not sure I follow this part... Why do we need to pass in `Never`?  And Why was it hard-coded to `Control.bool False` beforehand? 🤔 

Testing out the styleguide example tooltip, I can no longer toggle the "tail" by checking `withoutTail`. (But on the master branch, toggling it does work.)

LearningNerd (commented):
Only question / behavior I wanted to confirm before merging is that the tooltip styleguide example no longer lets me toggle "withoutTail". It just always has a tail now:
image

I posted a couple inline comments as well for parts of the code that I didn't follow. The tooltip one really confused me lol 😅

styleguide-app/Examples/Balloon.elm
```diff
@@ -103,41 +102,43 @@ paddingOptions =
...
-            List.filterMap identity
-                [ settings.theme
-                , settings.position
-                , settings.width
-                , settings.padding
-                ]
+            Control.currentValue state.attributes
     in
-    [ Control.view SetDebugControlsState state |> fromUnstyled
+    [ Control.view SetCopy state.copy |> fromUnstyled
```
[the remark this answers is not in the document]
tesk9: elm-css pretty much copies _all_ of `elm/html`, and has its own version of the `elm/html` node types! See https://github.com/rtfeldman/elm-css/blob/17.0.1/src/VirtualDom/Styled.elm#L42 and https://github.com/elm/virtual-dom/blob/master/src/VirtualDom.elm#L40.

We need to use `fromUnstyled` here because `Control.view` produces the `elm/html` `Html` type and the rest of our `Html` is the elm-css version.

It's definitely confusing! There was a period when much of our monolith codebase was a mix of elm-css html and elm html and folks exposed to it for the first time were often very confounded by the whole thing 😖

styleguide-app/Examples/Tooltip.elm
tesk9: OOPS I see what I did 🙈 `if True then` Thanks for noticing!!!

---

As to how this is _supposed_ to work when I don't hardcode bools:

`Control.bool` takes a bool argument so that it can create a checkbox that's initialized in the state you want.

So, if I wanted the tooltip example to show the without-tail behavior by default, I would change `Control.bool False` to `Control.bool True`

tesk9: Oh there's a confounding problem -- `withAlignment` in the Tooltip component wipes the no-tail setting. That's not good! Alignment and tail-less-ness should be independent properties (you can be an evil hamster OR an evil mouse, after all!)

I'm not going to mess with the internals of tooltip in this PR. But definitely good to know!

tesk9:
@LearningNerd what do you think, okay to merge?

LearningNerd:
Yes, I was just catching up on your updates and replies here. :) Looks great! 

@NoRedInkBot r+
sheepdance

LearningNerd (approved):
Oh right, no bot here! I'll click the button :)

Issue #766, opened 2021-10-30
Can't set alignment and remove tail on Tooltip at the same time


tesk9:
Example: https://ellie-app.com/fHSRvLJ5Ktsa1

First reported by @LearningNerd here: https://github.com/NoRedInk/noredink-ui/pull/761#discussion_r739535231

PR #768, opened 2021-11-01
Split Tooltip alignment and tail

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/766

`withoutTail` should always remove the Tooltip tail, even if the attributes later includes a shift in alignment, so that the tooltip isn't centered.


micahhahn (approved):
Looks good!

Image

PR #775, opened 2021-11-05
Elm style helpers for working with conditional styles

micahhahn:
Working with conditional styles can be awkward sometimes.  For example I recently needed to make the `backgroundColor` conditional in the following code: 

```
[ flexGrow (num 1)
, flexShrink zero
, displayFlex
, flexDirection column
, bottom zero
, Css.width (pct 100)
, padding2 (px 12) (px 20)
, borderTop3 (px 1) solid Colors.gray92
, backgroundColor Colors.frost
, Css.Media.withMedia
  [ Css.Media.only Css.Media.print [] ]
  [ backgroundColor transparent ]
]
```

Typical approach 1 is handling optional styles separately: 

```
([ flexGrow (num 1)
 , flexShrink zero
 , displayFlex
 , flexDirection column
 , bottom zero
 , Css.width (pct 100)
 , padding2 (px 12) (px 20)
 , borderTop3 (px 1) solid Colors.gray92
 , Css.Media.withMedia
   [ Css.Media.only Css.Media.print [] ]
   [ backgroundColor transparent ]
 ]
   ++ List.filterMap identity
     [ if isEmphasized then
         Just <| backgroundColor Colors.frost

       else
         Nothing
     ]
)
```

Typical approach 2 is making all styles optional: 

```
([ Just <| flexGrow (num 1)
 , Just <| flexShrink zero
 , Just <| displayFlex
 , Just <| flexDirection column
 , Just <| bottom zero
 , Just <| Css.width (pct 100)
 , Just <| padding2 (px 12) (px 20)
 , Just <| borderTop3 (px 1) solid Colors.gray92
 , if isEmphasized then
    Just <| backgroundColor Colors.frost
    
   else
    Nothing
 , Just <|
    Css.Media.withMedia
        [ Css.Media.only Css.Media.print [] ]
        [ backgroundColor transparent ]
 ]
    |> List.filterMap identity
)
```

Both of these feel rather heavy to me and not ideal.  

This PR adds two new functions `styleIf` and `styleJust` which are modeled after `viewIf` and `viewJust`. 

```
[ flexGrow (num 1)
, flexShrink zero
, displayFlex
, flexDirection column
, bottom zero
, Css.width (pct 100)
, padding2 (px 12) (px 20)
, borderTop3 (px 1) solid Colors.gray92
, styleIf isEmphasized [ backgroundColor Colors.frost ]
, Css.Media.withMedia
    [ Css.Media.only Css.Media.print [] ]
    [ backgroundColor transparent ]
]
```

~This is accomplished by using an unrecognized vendor-prefix property as a css no-op (-nri: noop;).~

~The downside, of course, with this approach is that we will be injecting extra css into our application.~

Curious for thoughts on if this is a good idea or not 😃

### DIFF
--- a/src/Nri/Ui/Css/V1.elm
+++ b/src/Nri/Ui/Css/V1.elm
@@ -6,7 +6,7 @@ module Nri.Ui.Css.V1 exposing (styleIf, styleJust)
 
 -}
 
-import Css exposing (Style, property)
+import Css exposing (Style, batch)
 
 
 {-| Apply a style only if a condition is `True`.
@@ -17,7 +17,7 @@ styleIf view condition =
         view ()
 
     else
-        property "-nri" "noop"
+        batch []
 
 
 {-| Apply a style if the `Maybe` is a `Just`, otherwise apply nothing.
@@ -29,4 +29,4 @@ styleJust view maybe =
             view whatever
 
         Nothing ->
-            property "-nri" "noop"
+            batch []
--- a/src/Nri/Ui/Css/V1.elm
+++ b/src/Nri/Ui/Css/V1.elm
@@ -11,10 +11,10 @@ import Css exposing (Style, batch)
 
 {-| Apply a style only if a condition is `True`.
 -}
-styleIf : (() -> Style) -> Bool -> Style
-styleIf view condition =
+styleIf : Bool -> Style -> Style
+styleIf condition view =
     if condition then
-        view ()
+        view
 
     else
         batch []
--- a/src/Nri/Ui/Css/V1.elm
+++ b/src/Nri/Ui/Css/V1.elm
@@ -1,0 +1,26 @@
+module Nri.Ui.Css.V1 exposing (..)
+
+import Css exposing (Style, property)
+
+
+{-| Apply a style only if a condition is `True`.
+-}
+styleIf : (() -> Style) -> Bool -> Style
+styleIf view condition =
+    if condition then
+        view ()
+
+    else
+        property "-nri" "noop"
+
+
+{-| View value of if `Maybe` is a `Just`, otherwise show nothing.
+-}
+styleJust : (a -> Style) -> Maybe a -> Style
+styleJust view maybe =
+    case maybe of
+        Just whatever ->
+            view whatever
+
+        Nothing ->
+            property "-nri" "noop"
--- a/src/Nri/Ui/Css/V1.elm
+++ b/src/Nri/Ui/Css/V1.elm
@@ -1,5 +1,11 @@
 module Nri.Ui.Css.V1 exposing (styleIf, styleJust)
 
+{-|
+
+@docs styleIf, styleJust
+
+-}
+
 import Css exposing (Style, property)
 
 
@@ -14,7 +20,7 @@ styleIf view condition =
         property "-nri" "noop"
 
 
-{-| View value of if `Maybe` is a `Just`, otherwise show nothing.
+{-| Apply a style if the `Maybe` is a `Just`, otherwise apply nothing.
 -}
 styleJust : (a -> Style) -> Maybe a -> Style
 styleJust view maybe =
--- a/src/Nri/Ui/Css/V1.elm
+++ b/src/Nri/Ui/Css/V1.elm
@@ -1,4 +1,4 @@
-module Nri.Ui.Css.V1 exposing (..)
+module Nri.Ui.Css.V1 exposing (styleIf, styleJust)
 
 import Css exposing (Style, property)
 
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/Css/V1.elm
```diff
@@ -0,0 +1,32 @@
...
+
+{-| Apply a style only if a condition is `True`.
+-}
+styleIf : (() -> Style) -> Bool -> Style
+styleIf view condition =
+    if condition then
+        view ()
+
+    else
+        property "-nri" "noop"
```
BrianHicks, line 20: this can be unhacked like so:

```suggestion
        Css.batch []
```

BrianHicks (changes requested):

src/Nri/Ui/Css/V1.elm
micahhahn: Oh man I searched for this forever.  Amazing

micahhahn:
Thanks to Brian for pointing out the `Css.batch []` way of generating an empty style.  It does not appear like we use at all in our NoRedInk repo as a way of collapsing optional styles.  So maybe adding these functions and letting folks know they exist is still wise?

src/Nri/Ui/Css/V1.elm
micahhahn: Or rather an approach that wasn't hacky.

BrianHicks:
stands to reason. I'm not sure why we don't use `batch []` in our styles… I feel like I've used it a lot! But maybe just for prototyping and that code all got removed or refactored away? 🤷

micahhahn:
I like that it give us a nice tight one-liner for the `styleIf` case.  elm-format wants to always split out lines were I to try and use an `if` statement with `Css.batch []`

src/Nri/Ui/Css/V1.elm
```diff
@@ -11,10 +11,10 @@ import Css exposing (Style, batch)
 
 {-| Apply a style only if a condition is `True`.
 -}
-styleIf : (() -> Style) -> Bool -> Style
-styleIf view condition =
+styleIf : Bool -> Style -> Style
+styleIf condition view =
```
BrianHicks, line 15: how would you feel about this instead:

```elm
styleIf : Bool -> List Style -> Style
styleIf cond stuff = batch (if cond then stuff else [])
```

The result would look like:

```elm
styleIf (Flags.isEnabled model.someFeatureFlag)
    [ backgroundColor red
    , float left
    , position absolute
    , blink True
    ]
```

(with, uh, better styles but you get the idea)

It might be better to flip the arguments, but I don't see us using this a lot in pipelines so I don't think it matters and we can define a local `flip` for it if we need it.

micahhahn: Yeah, I like that better.  More than likely there will be a handful of associated styles you'd wish to set.

src/Nri/Ui/Css/V1.elm
```diff
@@ -0,0 +1,34 @@
...
+            stuff
+
+         else
+            []
+        )
+
+
+{-| Apply a style if the `Maybe` is a `Just`, otherwise apply nothing.
+-}
+styleJust : (a -> Style) -> Maybe a -> Style
```
BrianHicks, line 27: should this one switch too then?

src/Nri/Ui/Css/V1.elm
stoeffel: This is not blocking, but I prefer adding less DSL like constructs and use tools that new people are already familiar.
```elm
Css.batch <| if (Flags.isEnabled model.someFeatureFlag) then
    [ backgroundColor red
    , float left
    , position absolute
    , blink True
    ]
else
  []
```
Or if we really are annoyed by this, I could see:
```elm
if (Flags.isEnabled model.someFeatureFlag) then
   Css.batch [ backgroundColor red
    , float left
    , position absolute
    , blink True
    ]
else
  Nri.Css.noStyles
```

stoeffel: Adding `styleIf` opens up the door for adding more and more DSL-thingies. See https://package.elm-lang.org/packages/elm-community/html-extra/latest/Html-Extra
Like later people need to learn `styleJust` `styleNothing` `styleUnless` and so on. 
Where as adding `noStyles` can be used in all these cases with "normal" elm syntax.

micahhahn:
@stoeffel @BrianHicks 

Wow, wait, so we already have these functions defined?  ur/src/Css.Extra defines `empty : Style`, `styleIf : Bool -> List Style -> Style` and `styleMaybe : Maybe Style -> Style`.  

I'm going to to close this PR without merging then.  Perhaps I'll do a D&D on these functions as a reminder that they exist - at the very least I don't feel like I've seen them in code that Raven regularly works in.

PR #780, opened 2021-11-10
add describedBy and disclosure options to radio buttons

ap-nri:
This is not styled. This is not a real PR because if it were I'd be making `Nri.Ui.RadioButton.V3` and incorporating all the TODOs in Examples/RadioButton.elm. Rather, this is what could be. Do we like a disclosure property? Instead of `disclosure : Maybe (Html msg)` should it be `disclosure : Maybe ({ isChecked : Bool, label : String } -> Html msg)` because I went back and forth on that a few times. 

MOV demo video

https://user-images.githubusercontent.com/75996561/141030888-9381ce30-25e6-438c-954e-913d71c35db6.mov

### DIFF
--- a/styleguide-app/Examples/RadioButton.elm
+++ b/styleguide-app/Examples/RadioButton.elm
@@ -10,13 +10,15 @@ module Examples.RadioButton exposing
 
 -}
 
+import Accessibility.Styled.Aria as Aria
+import Accessibility.Styled.Style exposing (invisible)
 import Category exposing (Category(..))
 import Css exposing (..)
 import Debug.Control as Control exposing (Control)
 import Dict exposing (Dict)
 import Example exposing (Example)
 import Html.Styled as Html exposing (..)
-import Html.Styled.Attributes exposing (css)
+import Html.Styled.Attributes as Attributes exposing (css)
 import KeyboardSupport exposing (Direction(..), Key(..))
 import Nri.Ui.Button.V10 as Button
 import Nri.Ui.Data.PremiumLevel as PremiumLevel exposing (PremiumLevel)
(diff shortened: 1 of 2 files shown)


andrewMacmurray:
Makes sense to me! The only thing that took me a moment was the name `disclosure`, what about `labelContent`?

I like `disclosure: Maybe (Html msg)` over the fancier version 👍

andrewMacmurray (approved):
Makes sense to me! The only thing that took me a moment was the name `disclosure`, what about `labelContent`?

I like `disclosure: Maybe (Html msg)` over the fancier version 👍

ap-nri:
TODO: use `aria-controls` on the radio input to link it to the disclosure it's controlling, https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

ap-nri:
@andrewMacmurray I went with `disclosure` since that's the [name of the pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) when you have `<details><summary></summary></details>`. But it's not a perfect fit for that field on the record so I'm not wed to the idea.

andrewMacmurray:
> @andrewMacmurray I went with `disclosure` since that's the [name of the pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) when you have `<details><summary></summary></details>`. But it's not a perfect fit for that field on the record so I'm not wed to the idea.

If it's in the web standards I'm sold!

styleguide-app/Examples/RadioButton.elm
```diff
@@ -102,6 +110,62 @@ viewVanilla state =
...
+viewDisclosures state =
+    div []
+        [ RadioButton.view
+            { label = "Reset to default class password"
+            , value = "DefaultPassword"
+            , name = "radio-button-examples"
+            , selectedValue = state.selectedValue
+            , onSelect = Select
+            , valueToString = identity
+            , describedBy = [ "default-class-password" ]
```
tesk9, line 129: I'm not sure that, as a developer using the RadioButton, I would understand what values to pass in to `describedBy` 🤔

For instance, why do I/how would I know to pass in the id for the "The default class password is" element but not the id for the "Copy password" button?

tesk9:
@ap-nri are you still using this PR?

PR #781, opened 2021-11-11
Adjust inputs' margin top

tesk9:
<details>
<summary> Previous description

~~Changes the amount of space on top of an input from 9px to 5px. See https://www.pivotaltracker.com/story/show/180137785~~

</summary>

cc @NoRedInk/design 

### Before

Screen Shot 2021-11-11 at 11 06 07 AM

### After

Screen Shot 2021-11-11 at 11 05 42 AM

</details>

Adjust error/guidance positioning


bendansby:
Oh I'm sorry, I made an assumption here that was wrong. I thought the defaultMarginTop was for the space between input and error messages. We don't need to change anything on the inputs themselves, so I think this can be closed.

tesk9:
Oh! Sweet, makes sense. Would you like the error message to be closer to the text input/select? That should be 5px?

bendansby:
Yes, tiny error messages and captions should always be 5 px below inputs

tesk9:
Okay, so I've got the error message 5px down from the select (same styles apply for the textinput):
Screen Shot 2021-11-11 at 2 13 27 PM


Which looks like so:
Screen Shot 2021-11-11 at 2 12 06 PM


But which leads me to a question. Do we want guidance messages to (a) be compact or (b) minimize layout changes when there's an error?

### a 
Screen Shot 2021-11-11 at 2 12 38 PM

### b
Screen Shot 2021-11-11 at 2 12 10 PM


### c
LMK what you'd prefer!

bendansby:
Ooh that's a tough one. I'm going to delay my answer by noting if I look at your Netlify preview and change it to 5px, it looks less tight than your A screenshot above. Is that screenshot in fact using 5px?

image

tesk9:
(a) was using whatever the default Text.caption top margin is (I removed the custom styles entirely), so yeah I think 0 top margin

(b) was using 8px top margin (I think the text of the tiny message is shifted down 3px by the icon)

I'll add (c), with 5px margin:

image

But now I'm noticing that the line height is definitely playing a non-trivial role here 🙈 

image

The orange bit is the margin...

bendansby:
Great, thank you. Yeah if we can, I think we should normalize caption line-height when used as guidance.

tesk9:
How's this?

Screen Shot 2021-11-12 at 9 19 49 AM
Screen Shot 2021-11-12 at 9 26 17 AM
Screen Shot 2021-11-12 at 9 26 12 AM

bendansby:
Beautimous!

ap-nri (approved):
@NoRedInkBot r+

PR #782, opened 2021-11-11
Tessa/re org examples slightly

tesk9:
Condense categories so that it's easier to browse the examples

## Before

image

## After

image


andrewMacmurray (approved):
Nice, love a design atom :atom:

PR #783, opened 2021-11-12
Add UiIcon.mail

macoca:
Adding a new mail icon and including it into the list of actions

Screen Shot 2021-11-12 at 12 42 15 PM


tesk9 (approved):
Looks great!



I would mayyyybe move it to the Communication section from the Actions section, but not a big deal and not blocking.

PR #787, opened 2021-11-19
Create RadioButton V3

ap-nri:
* [x] list based API instead of Record based
* [ ] disclosure content that automatically does all the aria stuff detailed [here](https://paper.dropbox.com/doc/A11y-radio-buttons-for-new-password-modal--BWXjNkCejrq5lzatMSuHtOfHAg-nTc7EdB4OcR2r9e1njMmb) 
  * [x] functions
  * [ ] looks cool
* [x] describedBy handler
  * add reasoning here for why not make that internal to the radiobutton 
* [ ] handle focus when closing modals

https://www.pivotaltracker.com/story/show/180083800

### DIFF
--- a/src/Nri/Ui/RadioButton/V3.elm
+++ b/src/Nri/Ui/RadioButton/V3.elm
@@ -63,7 +63,7 @@ enabled =
 name : String -> Attribute value msg
 name name_ =
     Attribute emptyEventsAndValues <|
-        \config -> { config | name = name_ }
+        \config -> { config | name = Just name_ }
 
 
 {-| Sets the value of one radio button
@@ -115,8 +115,8 @@ emptyEventsAndValues =
 {-| This is private. The public API only exposes `Attribute`.
 -}
 type alias Config =
-    { label : String
-    , name : String
+    { label : Maybe String
+    , name : Maybe String
     , isLocked : Bool
     , isDisabled : Bool
     , showPennant : Bool
@@ -125,8 +125,8 @@ type alias Config =
 
 emptyConfig : Config
 emptyConfig =
-    { label = ""
-    , name = ""
+    { label = Nothing
+    , name = Nothing
     , isLocked = False
     , isDisabled = False
     , showPennant = False
@@ -209,15 +209,15 @@ internalView config attributes =
         config_ =
             applyConfig attributes
 
-        unvalidatedRadioConfig : Maybe value
+        unvalidatedRadioConfig : ( Maybe value, Maybe String )
         unvalidatedRadioConfig =
-            eventsAndValues.value
+            ( eventsAndValues.value, config_.name )
     in
     case unvalidatedRadioConfig of
-        Just realValue ->
+        ( Just value_, Just name_ ) ->
             let
                 id_ =
-                    config_.name ++ "-" ++ dasherize (toLower (config.valueToString realValue))
+                    name_ ++ "-" ++ dasherize (toLower (config.valueToString value_))
             in
             Html.span
                 [ id (id_ ++ "-container")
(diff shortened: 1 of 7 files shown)


src/Nri/Ui/RadioButton/V3.elm
```diff
@@ -0,0 +1,561 @@
...
+
+maybeAttr : (a -> Html.Attribute msg) -> Maybe a -> Html.Attribute msg
+maybeAttr attr maybeValue =
+    maybeValue
+        |> Maybe.map attr
+        |> Maybe.withDefault Attributes.none
+
+
+{-| View a single radio button.
+Renders nothing if the attributes list does not contain value, name, and valueToString.
```
ap-nri, line 217: on the fence here where maybe instead of `String -> List (Attribute value msg) -> Html msg` we should go full circle back to `{label : String, name : String, value : value, valueToString : value -> String } -> List (Attribute value msg) -> Html msg` and only use the attributes list for non-critical fields.

ap-nri:
OK my next commit takes some explaining of what I'm driving at. Why hide the label? Why are you like this Alex? I've marked up the screenshot of the expanded state from [the antetype](https://cloud.antetype.com/exports/link/e9686220-bd56-4da3-8561-e57190d12481/) with my thinking: the red box around the grey area is a wrapping `<button>` with `display: flex` (or maybe `display: grid`, depends on how the next few commits go) to have its children display as a row filling it from left to right. The purple boxes are the children of the red boxed parent. In particular the middle box needs to contain the two yellow boxes and show them vertically. This means that what is semantically the label (top yellow box) is getting rendered as the child of the `disclosure` content. In that case we would want the `label` content rendered by the `RadioButton.view` to not appear visually and then render it redundantly inside of the disclosure content when that content is expanded. This may also involve making `disclosure : { label: Html never} -> List (Html msg) -> Attribute value msg` so that the label with the proper `for` and styling is passed in to be rendered by the disclosure helper. 

I may be able to avoid all of this mishegas by using CSS Grid and passing custom styling to the label. We shall see. 

marked up screenshot of antetype showing box model of how the belemnts of the radio button fit together in a flexbox model

ap-nri:
copied from my comment to the same effect in the standup doc.

Getting the disclosure content to look how it looks in the design and keep the other features in the API (inline rendering, premium flag) has been a hassle. Really I think I’ve kind of gone too far down the path of cramming things into the radio button API. We could instead only change the API to have the following added options

hide/show label
aria-controls
an exposed helper function that returns the radio button’s ID
Then we just have all of the show/hide disclosure stuff in our app code and use the hidden-label version of the radio button. But the list based API was worth doing so it’s not like this is a loss.

tesk9:
Thanks for the great work on this, Alex! And for the clear notes on where you were considering going next!

I did the following:
- leveraged Debug.Controls on the example code (to my mind, this really illustrates how helpful a RadioGroup component might be in the future!)
- went with the require-required-fields option from https://github.com/NoRedInk/noredink-ui/pull/787#discussion_r753280949
- added an `id` attribute that overrides the default generated radio button id 
- removed the inline/block rendering distinction (I think `containerCss` might help to do the same thing?)
- added the typical attributes (`custom`, `containerCss`, etc.)
- added the shared Input error message and guidance helpers (maybe guidance and error messages only make sense for the radio group, and not for the individual inputs? I could see arguments either way.)


Screen Shot 2021-11-22 at 4 35 14 PM
Screen Shot 2021-11-22 at 4 35 34 PM

andrewMacmurray (approved):
This looks great, thanks for your hard work on this Alex and Tessa! If @bendansby is ok with the disclosure appearing below the label then we shouldn't need to make any changes to this

ap-nri:
This has my strongest stamp of approval! I suspect we won't end up using `disclosure` and will take that out in a sec for reasons discussed previously here: https://github.com/NoRedInk/noredink-ui/pull/787#issuecomment-974838245, but if it's in the API I have ZERO issue with that. Not even a small one.

ap-nri:
So in conclusion: I like it, Tessa likes it, Andrew likes it, and Ben likes it (by the transitive property https://github.com/NoRedInk/NoRedInk/pull/36437) so I'm going to merge this in 30 minutes unless someone comes in screaming "WAAAIIIITTTT"

ap-nri:
I checked something that's been a splinter in my mind and this is good for it: if we specify the ID and hide the label in order to have a custom styled label, are there any bad outcomes from having another `<label for="our-custom-id">` in the DOM? Not according to [the spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attributes) and not according to [the experiment I just made](https://codepen.io/ap-nri/full/PoJoKYv). Both labels get read which is redundant if we have both match each other, but we can make the mandatory label `""` and add a label and be all ok from an a11y perspective.

PR #792, opened 2021-12-02
RadioButton preview

tesk9:
Adds preview for RadioButtons, now that RadioButton.V3 supports it!

image

@LearningNerd not sure if you're planning to follow Ben's suggestion for hack day, but this is an example of how to set up a preview. Last hack day, I did most of the straightforward previews -- some of the remaining ones may not let you use the real component to make the example. See the modal example for one way around not being able to use the real comopnent.

### DIFF
--- a/styleguide-app/Examples/RadioButton.elm
+++ b/styleguide-app/Examples/RadioButton.elm
@@ -10,6 +10,7 @@ module Examples.RadioButton exposing
 
 -}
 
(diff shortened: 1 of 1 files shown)


LearningNerd (approved):
Cool!! I didn't end up working on this today but will bookmark this for next hack day :) Thanks, Tessa! 

Looks great to me, here's my _seal_ of approval! 
sealbounce

PR #794, opened 2021-12-03
Adds general sidenav component

tesk9:
There are definitely accessibility improvements and API improvements left to make here, but this is a start!

Step 1 of https://trello.com/c/SPSSTOZq/487-style-guide-consolidate-and-add-sidebar-nav See https://noredink.slack.com/archives/C9MEFL3KL/p1638548734064000 for some spec clarification.

<details><summary>Old screenshots</summary>

Screen Shot 2021-12-03 at 12 04 37 PM

---

Screen Shot 2021-12-03 at 12 11 40 PM

</details>

cc @NoRedInk/design

### DIFF
--- a/src/ClickableAttributes.elm
+++ b/src/ClickableAttributes.elm
@@ -23,9 +23,9 @@ import Nri.Ui.Html.Attributes.V2 as AttributeExtras exposing (targetBlank)
 
 
 {-| -}
-type alias ClickableAttributes msg =
+type alias ClickableAttributes route msg =
     { linkType : Link
-    , url : String
+    , url : Maybe route
     , onClick : Maybe msg
     }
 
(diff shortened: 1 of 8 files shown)


bendansby:
Cool! Ready for style notes yet or nah?

tesk9:
Definitely would love style notes!!

bendansby:
Kinda looks like the styles of the first and second screenshots need to merge. The first has the right container padding, while the second has the right distance between items.

tesk9:
How do these look?

Screen Shot 2021-12-03 at 1 09 42 PM
Screen Shot 2021-12-03 at 1 09 48 PM

bendansby:
Perfect!

src/ClickableAttributes.elm
```diff
@@ -102,12 +112,24 @@ toButtonAttributes clickableAttributes =
 
 
 {-| -}
-toLinkAttributes : ClickableAttributes msg -> ( String, List (Attribute msg) )
-toLinkAttributes clickableAttributes =
+toLinkAttributes : (route -> String) -> ClickableAttributes route msg -> ( String, List (Attribute msg) )
+toLinkAttributes routeToString clickableAttributes =
```
juanedi, line 116: 🎨  It feels like there are some implicit rules here about what the valid combinations of `url`/`urlString` and `linkType` are, and I wonder if modelling those would simplify this function and make this easier to understand overall.

I'm thinking something like this:

```elm
type alias ClickableAttributes route msg =
    { linkType : Link route
    , onClick : Maybe msg
    }


type Link route
    = Default
    | WithTracking route
    | SinglePageApp route
    | WithMethod String route
    | External String
    | ExternalWithTracking String
```

It seems like:
  - `Default` is the case in which there isn't any target `url`/`urlString`
  - We only use `urlString` with external links
  - For all other cases we use `url`.

If this made sense, we could implement this with a single case, grabbing the target link from the `Link`'s constructor params.

juanedi (commented):
This looks awesome! 😻 😻 😻 

Left just one suggestion with a potential refactor in `ClickableAttributes`. If that doesn't make sense for any reason feel free to merge!

src/ClickableAttributes.elm
tesk9: Thanks for the suggestion here! I'm going to link this in an issue, and come back to it when I've got a little more bandwidth. 🙏

Issue #795, opened 2021-12-03
Reposition Premium Pennants on Premium checkbox


tesk9:
See https://github.com/NoRedInk/NoRedInk/pull/36480#issuecomment-985829602 for context

PR #797, opened 2021-12-07
Kraken/input positioning

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/795

image

Pennants showing to the left of the checkbox.

---

@NoRedInk/design I have a couple of questions/concerns that don't necessarily block this PR, but that I thought I'd best mention:
- Should the Pennant flip to look like `>=|` instead of like `|=<`? It doesn't look like a sticky that's attached to the checkbox anymore
- Should the Pennant do anything on click? The RadioButton component launches the Premium modal when you click the Premium Pennant, not if you click the label or input.
- Should the RadioButton pennant also go on the other side?

I also want to flag (ha) that there are accessibility issues with the way the Premium icon is currently shown -- it's not attached to the input in a screenreader-friendly way. I think fixing this relationship will require a new Checkbox version to do properly, so I'm not going to do it yet.

### DIFF
--- a/deprecated-modules.csv
+++ b/deprecated-modules.csv
@@ -1,6 +1,7 @@
 Nri.Ui.Accordion.V1,upgrade to V3
 Nri.Ui.Menu.V1,upgrade to V3
 Nri.Ui.Menu.V2,upgrade to V3
+Nri.Ui.PremiumCheckbox.V6,upgrade to V7
 Nri.Ui.RadioButton.V2,upgrade to V3
 Nri.Ui.Table.V4,upgrade to V5
 Nri.Ui.Tabs.V6,upgrade to V7
--- a/elm.json
+++ b/elm.json
@@ -43,6 +43,7 @@
         "Nri.Ui.Palette.V1",
         "Nri.Ui.Pennant.V2",
         "Nri.Ui.PremiumCheckbox.V6",
+        "Nri.Ui.PremiumCheckbox.V7",
         "Nri.Ui.RadioButton.V2",
         "Nri.Ui.RadioButton.V3",
         "Nri.Ui.SegmentedControl.V14",
--- a/forbidden-imports.toml
+++ b/forbidden-imports.toml
@@ -78,6 +78,9 @@ usages = ['styleguide-app/Examples/RadioButton.elm']
 [forbidden."Nri.Ui.Modal.V3"]
 hint = 'upgrade to V11'
 
+[forbidden."Nri.Ui.PremiumCheckbox.V6"]
+hint = 'upgrade to V7'
+
 [forbidden."Nri.Ui.RadioButton.V1"]
 hint = 'upgrade to V2'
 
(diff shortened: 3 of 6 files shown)


bendansby:
Should the pennant even be in the checkbox and radio button components? Showing it on the left works in the assignment library, but maybe not elsewhere. And I don't even know if we use/should use it elsewhere, so maybe it should just be a thing that's implemented on the page?

tesk9:
Hmm! Well, I think in order for us to make the relationship between the pennant and the content that it's labelling clear & consistent & programmatic, and keep the premium upsell behavior consistent and predictable, it makes sense to have some extracted logic around what the pennant does/should do. 

Like -- the pennant showing on the left or the right is a visual choice that might depend contextually. But I believe we are going to need (at some point) to add an `aria-describedby` tying the "Premium" text in the SVG with the checkbox input. I'd like to keep this as part of the API so that fixing the a11y problems in many places can be done simply.

I could add attributes for Left/Right pennant alignment?

bendansby:
OK, let me take a look this week where we're using the pennant and see if I can cohere a viewpoint

bendansby:
As far as I can tell and recall, the only other time we use pennants with checkboxes and radio buttons is with Peer and Self Review on the assignment form. Does that jibe with your understanding?

tesk9:
I think that's mostly right!

I'm seeing that the assignment form "Assign to" section also uses the premium pennant with checkboxes:
image

And you can still get to Step 2 of the assignment form (😭) to see the pennants on pathways:
image

There are other places where features are gated by premiumness for the course (e.g., the Mastery page "See all pathways" dropdown will show locked standards for non-premium courses), but it looks like we're not using checkboxes or radios for these.

bendansby:
> And you can still get to Step 2 of the assignment form 

By entering a URL?

tesk9:
> By entering a URL?

yeeeeees? I feel like I'm missing something because it seems like we would have removed the view if there wasn't a way to get to it _not_ by URL?

But yeah -- it seems like editing doesn't get you there and shareable blueprints don't get you there... 

You end up on Step 2 for planning diagnostics if you use the "Create your Own" button, but categories aren't ever premium... 🤔

bendansby:
Yeah I'm surprised we haven't heart with skull eyesed those views yet! At the least I think you should get redirected to the AL.

Regardless, it seems like doing a universal left pennant alignment should be fine, except for the classes in the "Assign to" section on mobile; they'll likely get cut off there. I can make a style adjustment there though, maybe put that section in a container with padding on mobile.

And since this seems like kind of a dormant pattern, I think we can get away with punting on a technophilosophical revelation as to the locked/unlocked/checkbox/radio/pennnant relationship and just go with your cleanup! Something I'll keep an eye on though if we start to use checkboxes in this way again, though.

LearningNerd (approved):
Looks great! Seal of approval. First step, PremiumCheckbox.V7. Next, _the world~~~_
sealbounceworld

PR #799, opened 2021-12-09
Kraken/radio styling

tesk9:
Change the radio icon SVG to align vertically at the center of the label & disclosure content, rather than at the top. 

See https://github.com/NoRedInk/NoRedInk/pull/36500 for the context in which this will be useful.

image

Screen Shot 2021-12-09 at 1 32 21 PM

This should also improve the screenreader experience described in https://github.com/NoRedInk/NoRedInk/pull/36492#issuecomment-990259528 

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/RadioButton/V3.elm
+++ b/src/Nri/Ui/RadioButton/V3.elm
@@ -327,6 +327,7 @@ view { label, name, value, valueToString, selectedValue } attributes =
         , css
             [ position relative
             , marginLeft (px -4)
+            , Css.paddingLeft (Css.px 40)
             , display inlineBlock
             , Css.height (px 34)
             , pseudoClass "focus-within"
@@ -380,8 +381,7 @@ view { label, name, value, valueToString, selectedValue } attributes =
                 , ( "Nri-RadioButton-RadioButtonChecked", isChecked )
                 ]
             , css
-                [ position relative
-                , outline Css.none
+                [ outline Css.none
                 , margin zero
                 , Fonts.baseFont
                 , Css.batch
@@ -400,7 +400,7 @@ view { label, name, value, valueToString, selectedValue } attributes =
                         , cursor pointer
                         ]
                     )
-                , padding4 (px 6) zero (px 4) (px 40)
+                , padding4 (px 6) zero (px 4) zero
                 , fontSize (px 15)
                 , Css.property "font-weight" "600"
                 , display inlineBlock
--- a/src/Nri/Ui/RadioButton/V3.elm
+++ b/src/Nri/Ui/RadioButton/V3.elm
@@ -301,20 +301,17 @@ view { label, name, value, valueToString, selectedValue } attributes =
                 |> Maybe.withDefault True
                 |> not
 
-        ( disclosureIds, disclosureElements ) =
-            case ( config.disclosedContent, isChecked ) of
-                ( [], _ ) ->
-                    ( [], [] )
-
-                ( _, False ) ->
-                    ( [], [] )
-
-                ( _, True ) ->
-                    ( List.indexedMap
-                        (\index _ -> (idValue ++ "-disclosure-content-") ++ String.fromInt index)
-                        config.disclosedContent
-                    , config.disclosedContent
+        ( disclosureIds, disclosedElements ) =
+            config.disclosedContent
+                |> List.indexedMap
+                    (\index element ->
+                        let
+                            id_ =
+                                (idValue ++ "-disclosure-content-") ++ String.fromInt index
+                        in
+                        ( id_, span [ Attributes.id id_ ] [ element ] )
                     )
+                |> List.unzip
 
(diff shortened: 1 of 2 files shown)


bendansby:
Looks not quite centered
Napkin 7 12-09-21, 2 24 50 PM

tesk9:
Ahh oops! Clarification on what's desired ---

I've got it like this currently:
image

Should the bottom of the "D" or the bottom of the "g" be aligned with the bottom of the inner circle? (Or should something else happen?)

bendansby:
I'm happy to have CSS center things as it will, except I think our containers and padding and positioning and such should be tidy to set CSS up for success, versus this:

Screen Shot 2021-12-09 at 2 45 37 PM
Screen Shot 2021-12-09 at 2 45 28 PM

Make sense?

tesk9:
So I think there's a couple of concerns here that require us to do funky CSS instead of sensible CSS -- I think I can get the resulting styles to look nicer, but I don't think there's a way around having complicated CSS styles inside the component.

- We should use an "input type='radio'" element to back our RadioButton component because this is the most robust/accessible approach
- We want custom SVGs for the Radio, instead of the default browser Radio
- We want clicking the Radio SVG icon to select the radio (meaning that the Radio svg icon needs to be _inside_ the input's label)
- We want the radio SVG icon to be centered vertically against the label text, the guidance, and the disclosed content (note that the guidance and disclosed content cannot be inside the label)
- When a "input type='radio'" is selected, the browser will scroll to that element's vertical position (which means that we need to put the real radio in-line with fancy styled radio, or we'll get bugs like https://linear.app/noredink/issue/ZEN-144/schoology-signup-flow-scroll-position-jumps-around-in-school-selection)

These requirements lead to our 😱 -inducing HTML structure (pseudocode ahoy):

```
<radiocontainer>
    <hiddenRealRadioButton></hiddenRealRadioButton>
    <label>
        <icon />
        <labelText>My Radio</labelText>
        <premiumPennant />
    </label>
   < guidance or error />
   < disclosedContent />
</radiocontainer>
```

This HTML structure makes positioning the icon against the labeltext, premiumpennant, guidance, and disclosed content tricky. If we could move the icon out of the label, things would be simpler, since then we could use flexbox to get perfect styles. Moving the icon out is likely to lead to bugs around the hit area for the input, even if we were to re-implement the label click behavior by hand.

---

All that said... I am going to do my best to make the alignment nice! (And someday when we're allowed to directly style radio buttons, instead of dancing our way around it, I shall breath a heavy sigh of relief!!!)

tesk9:
I see why my positioning was off -- I failed to account for a border and I forgot to multiply by 2 to account for top _and_ bottom in a couple of places. Should be fixed in 2 shakes!

tesk9:
Screen Shot 2021-12-09 at 5 05 13 PM
Screen Shot 2021-12-09 at 5 05 02 PM
Screen Shot 2021-12-09 at 5 05 28 PM
Screen Shot 2021-12-09 at 5 05 21 PM

bendansby:
Hooray! Thanks for talking it through for me and whipping it into shape!

ap-nri:
note to future viewers: checking the "disclosure" checkbox on the demo does move the radio buttons to the right because it reveals the disclosure input box, not because the radio buttons changed shape.

ap-nri (commented):
yes. YES

src/Nri/Ui/RadioButton/V3.elm
```diff
@@ -301,19 +301,17 @@ view { label, name, value, valueToString, selectedValue } attributes =
...
 
-        ( disclosureId, disclosureElement ) =
-            case ( config.disclosedContent, isChecked ) of
-                ( [], _ ) ->
-                    ( Nothing, text "" )
-
-                ( _, False ) ->
-                    ( Nothing, text "" )
-
-                ( (_ :: _) as childNodes, True ) ->
```
ap-nri, line 312: RIP truckzilla

micahhahn:
@tesk9 

This changed the way that the revision requests form looks as we are using radio button disclosures: 





I'm not sure I entirely like it, as the button label is really the primary action and the disclosure is secondary.  It's also a little jarring when it's position feels like it changes upon selection.  (That could be an artifact of using such a long disclosure though). 

This isn't a huge problem for Raven as we'll be (hopefully) be changing this as Quick Writes gets to full feature parity with Guided Drafts.  Just wanted to raise the concern

staceyadams:
Sorry I haven't been following this thread.

I think the radio svg vertically centered is visually more confusing than aligned to the vertical top, next to the text that first describes the option being selected, especially in the screenshots in Tessa's Dec 9th comment.

Agreed with Micah that the QW revision request case isn't a huge problem since it's temporary.

tesk9:
Sorry about the regression, Ravens!!! Feel free to make an issue if there are changes you want made 🙏

PR #801, opened 2022-01-06
adds role img and hide decorative svgs

tesk9:
Per audit (line one, https://docs.google.com/spreadsheets/d/1bRr62TZB0I3JZzLm-UDBIoAPFB3Dzl-B/edit#gid=14356013) -- our SVGs were missing role img and were being displayed even when they didn't have a title.

This PR updates the Svg behavior such that if there's a label, it's read out, and if there's _not_ a label, the SVG is assumed to be decorative.

This is a pretty safe default behavior, in my opinion, since our SVGs are generally used directly with other components (like ClickableSvg, ClickableText, and Button).

Prior to this PR, you could get to a state like this with VO:
Screen Shot 2022-01-05 at 4 10 11 PM

Now, you won't be able to use VO-arrow to navigate to the search icon (as an example).

cc @krinhorn

### DIFF
--- a/src/Nri/Ui/Svg/V1.elm
+++ b/src/Nri/Ui/Svg/V1.elm
@@ -12,6 +12,7 @@ module Nri.Ui.Svg.V1 exposing
 
 -}
 
+import Accessibility.Styled.Role as Role
 import Accessibility.Styled.Widget as Widget
 import Css exposing (Color)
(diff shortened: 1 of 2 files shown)


micahhahn (approved):
Makes sense to me! 

Image

Issue #804, opened 2022-01-07
Expose breakpoint value from MediaQuery


tesk9:
We are currently sometimes hardcoded 1000px in the monolith, where we should be guaranteeing seamless transitions.

PR #805, opened 2022-01-10
Add MediaQuery breakpoint px values

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/804. Also makes some minor improvements to the docs.

```
This is a MINOR change.

---- Nri.Ui.MediaQuery.V1 - MINOR ----

    Added:
        mobileBreakpoint : Css.Px
        quizEngineBreakpoint : Css.Px
```

### DIFF
--- a/src/Nri/Ui/MediaQuery/V1.elm
+++ b/src/Nri/Ui/MediaQuery/V1.elm
@@ -1,13 +1,18 @@
 module Nri.Ui.MediaQuery.V1 exposing
     ( mobile, notMobile
+    , mobileBreakpoint
     , quizEngineMobile
+    , quizEngineBreakpoint
     )
 
 {-| Standard media queries for NRI responsive pages.
 Can be used to create a Snippet using Css.Global.media
 
 @docs mobile, notMobile
+@docs mobileBreakpoint
+
 @docs quizEngineMobile
+@docs quizEngineBreakpoint
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/MediaQuery/V1.elm
```diff
@@ -1,52 +1,75 @@
 module Nri.Ui.MediaQuery.V1 exposing
     ( mobile, notMobile
+    , mobileBreakpoint
```
JSuder-xx, line 3: No big deal but I am curious about the need to expose the breakpoints. Do we expect other styling to be dependent upon the breakpoint pixel sizes?

JSuder-xx (approved):

src/Nri/Ui/MediaQuery/V1.elm
tesk9: I expect other styling to be dependent on the breakpoint, yes!

Specifically, I think we often have layouts like this:

image

Where there's an element that has a max-width or width equal to the page breakpoint. I think having the breakpoint and the max width be identical makes the resizing experience nicer. I could be wrong -- I just like the idea of the margin-auto decreasing until it can't decrease further, and then immediately transitioning into the mobile view

JSuder-xx: I agree that is a usability win! I don't think there is anything else we should be doing different here. 

I was probing to see if we could figure out a way to encapsulate the pixel size as an implementation detail; exposing more narrowly typed and specifically named abstractions instead. I agree it is too early for that. Attempts at abstraction before we understand the pattern lead to over-engineering and all. I just thought I'd ask (for grins). 

The idea was that if we wanted to make an element max-width equal to a mobile breakpoint then we might offer 
```elm
maximizedMobileContainer : List (Attribute msg) -> List (Html msg) -> Html msg
```
where the function includes the attributes necessary (setting the max width). However, that looks clumsy to me. If you see a bunch of repeated patterns in how people consume the breakpoints then maybe there will come a time for a new abstraction. Or maybe not. 😄

PR #807, opened 2022-01-10
Tessa/accessibility invisible style

tesk9:
Use https://package.elm-lang.org/packages/tesk9/accessible-html-with-css/latest/Accessibility-Styled-Style#invisibleStyle to hide navbar skiplink.

Also updates to elm-css 17.0.4, per https://noredink.slack.com/archives/C0VVDLEES/p1641739931013700


lindsaykwardell:
Looks good! Do we need to update elm-css in the monolith as well?

lindsaykwardell (approved):

tesk9:
Yes, we'll want to make similar changes in the monolith! I'll do it along with the update to the nxt version of noredink-ui

PR #811, opened 2022-01-14
Expand mobile modal height

bendansby:
Paired with @tesk9 

Allows mobile modals to reach max viewport height to maximize screen real estate.

Screen Shot 2022-01-14 at 2 21 39 PM

### DIFF
--- a/src/Nri/Ui/Modal/V11.elm
+++ b/src/Nri/Ui/Modal/V11.elm
@@ -610,19 +610,19 @@ viewInnerContent ({ visibleTitle } as config) =
         visibleFooter =
             not (List.isEmpty config.footer)
 
-        titleHeight =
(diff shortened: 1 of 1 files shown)


tesk9 (approved):
Woo, so excited!

Screen Shot 2022-01-14 at 3 13 29 PM

I noticed while reviewing that if we don't have the usual footer buttons, then the footer height will be slightly different, and that will cause the positioning to be slightly wrong.

I think this is still enough of an improvement that we should merge, but I'll leave the final call to you! (I'm thinking in the next version of Modal, we could try to change how the modal content is constrained so that its height doesn't depend on the height of the title and footer so explicitly)

bendansby:
Yeah I realized similar, but agree it's worth merging for the improvement. V12 will be great!

PR #812, opened 2022-01-15
Remove `aria-checked` from Checkbox's label

tesk9:
https://www.w3.org/TR/wai-aria-1.1/#aria-checked is not meant for items with a labelling role.

See row 379 https://docs.google.com/spreadsheets/d/1bRr62TZB0I3JZzLm-UDBIoAPFB3Dzl-B/edit#gid=14356013

https://linear.app/noredink/issue/A11-134/checkbox-incorrectly-uses-aria-label-and-aria-controls

### DIFF
--- a/src/Nri/Ui/Checkbox/V5.elm
+++ b/src/Nri/Ui/Checkbox/V5.elm
@@ -213,8 +213,6 @@ viewEnabledLabel :
 viewEnabledLabel model labelView icon =
     Html.Styled.label
         [ Attributes.for model.identifier
-        , Aria.controls model.identifier
-        , Widget.checked (selectedToMaybe model.selected)
         , labelClass model.selected
         , css
             [ positioning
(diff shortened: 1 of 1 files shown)


krinhorn:
Can we review live? I understand the issue but here's what I'm blocked on:
- I'm not sure how to actually test something with a screenreader from NRI-UI. (I tried to mock it by just using staging and ripping out `aria-checked`, but it comes back when I interact with the element and I'm not confident that's an accurate test.)
- I don't fully understand the code here, though I understand the jist of what it's doing.

Other things I'm unsure about:
- Do we still need `aria-controls="readaloud-checkbox"` on this element? (I'm a little confused about whether there are two focusable elements here or just one.) That might be outside the scope of this PR, though.

krinhorn (approved):
giphy

Do you have an easy way to go back and find audit-related PRs? If not, maybe for now we should go with the convention of adding PR links a rightmost new column in the audit doc? Or if you wanna try to set up the Linear+GH integration, you could make a story per audit item. That seems better if you have time, but if not, I at least want to make sure these are tracked _somehow_.

PR #815, opened 2022-01-18
Tessa/move lms logos

tesk9:
Extract logos from an Extra file in the monolith.

image

Zen -- I think that the schoology logo maybe has too much whitespace on top and on bottom due to the viewBox. Is it supposed to be shorter than the Canvas icon?


asterite:
Looks good!

> I think that the schoology logo maybe has too much whitespace on top and on bottom due to the viewBox. Is it supposed to be shorter than the Canvas icon?

Maybe this is a question for Ben, but I guess they should have similar sizes. I think they show up at the top-right of a course box in the dashboard, so maybe if they don't have similar sizes it's not noticeable, but I don't know.

tesk9:
@bendansby any thoughts on the viewbox here?

bendansby:
As a rule, SVG viewboxes should be cropped to the content. I don't think it's a problem in this case since as Ary noted we use it in just the one place, but if you want to do a quick fix or add a todo that'd be 👍

PR #816, opened 2022-01-22
Tessa/percy 2

tesk9:
Get Percy snapshots on https://percy.io/NoRedInk/NoRedInk-UI/ working again.


script/percy-tests.js
```diff
@@ -1,38 +1,81 @@
-const PercyScript = require('@percy/script')
+const expect = require('expect');
```
tesk9, line 1: Lots of this is from the example code for using percy/cli and puppeteer: https://github.com/percy/example-percy-puppeteer/blob/master/tests/todomvc_spec.js#L1

PR #820, opened 2022-01-25
Adds percy token

tesk9:
- Adds percy token as an env variable for github actions
- uses `--keep` flag so that `--pure` doesn't wipe out the percy token


3kh0 (approved):
Looks good

PR #838, opened 2022-02-18
Bat/button icon alignment

tesk9:
Changes to SVG rendered caused about 2px of icon shift on Button (which I didn't notice because we didn't have percy snapshots on noredink-ui including buttons with icons -- this PR changes the default view so that the snapshots should include some icons). This PR changes how styles are applied to the button icon so that the icons render in (what I hope is) the right spot.

We _did_ [have percy snapshots on monolith buttons with icons](https://percy.io/NoRedInk/NRI-Cypress/builds/15873368/changed/895582665), which is how we found this minor misalignment issue:
image

### Before - Preview

image

### After - Preview

Screen Shot 2022-02-18 at 11 50 36 AM

### Before - Left Arrow

image

### After - Left Arrow

Screen Shot 2022-02-18 at 11 50 54 AM

### Before - Download

image

### After - Download

Screen Shot 2022-02-18 at 11 50 41 AM

---

@NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -640,7 +640,7 @@ buttonStyles size width colors customStyles =
 viewLabel : ButtonSize -> Maybe Svg -> String -> Html msg
 viewLabel size maybeSvg label_ =
     let
-        { imageHeight } =
+        { fontAndIconSize } =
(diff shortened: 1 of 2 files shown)


bendansby:
Thanks!!

PR #840, opened 2022-02-21
Prevent scroll on arrow down

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/832

The Menu component should not scroll the page when the up or down arrows are used to navigate within the options.

https://user-images.githubusercontent.com/8811312/155007201-32a39043-0a6f-49b5-af39-f8df5c3f0c50.mov


waj (approved):
It works and look great!

PR #842, opened 2022-02-24
Shadows

bendansby:
Adjusts the shadows on Container, Modal, and Menu and adds them to Message.Large. Maybe we should have a Shadow.elm? Also, my shadows are semi-copy pasted so I'm just using the property property instead of trying to figure out how to make their syntax work with elm-css.

Screen Shot 2022-02-24 at 8 36 03 AM
Screen Shot 2022-02-24 at 8 35 49 AM
Screen Shot 2022-02-24 at 8 35 42 AM
image

### DIFF
--- a/src/Nri/Ui/Container/V2.elm
+++ b/src/Nri/Ui/Container/V2.elm
@@ -153,7 +153,7 @@ defaultStyles : List Css.Style
 defaultStyles =
     [ borderRadius (px 8)
     , border3 (px 1) solid Colors.gray92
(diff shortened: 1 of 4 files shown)


tesk9 (approved):
You beat Percy! It seems like it only caught the change to Message.

I like the new shadows! I agree that a Shadow module could be great to have (I'll add an issue to noredink-ui). And I think it's _very_ reasonable not to try to write these styles in the elm-css way (so many arguments 💀 )

Merge away!

bendansby:
Thanks! Also just pushed a shadow change for Menu as well.

PR #846, opened 2022-03-01
Add PremiumPromo PremiumLevel

caseyWebb:
Adds `PremiumPromo` PremiumLevel for upcoming marketing initiative.

There are no visible changes introduced here, and fwiw as far as I can tell RadioButton.V2 isn't being used anywhere.

### DIFF
--- a/src/Nri/Ui/Data/PremiumLevel.elm
+++ b/src/Nri/Ui/Data/PremiumLevel.elm
@@ -11,6 +11,7 @@ module Nri.Ui.Data.PremiumLevel exposing (PremiumLevel(..), allowedFor, highest,
 type PremiumLevel
     = Free
     | Premium
+    | PremiumPromo
     | PremiumWithWriting
 
 
@@ -44,9 +45,12 @@ order : PremiumLevel -> Int
 order privileges =
     case privileges of
         PremiumWithWriting ->
-            2
+            3
 
         Premium ->
+            2
+
+        PremiumPromo ->
             1
 
         Free ->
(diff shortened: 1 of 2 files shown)


celsobonutti (dismissed):
Compiles locally and everything seems fine.
Looks like we're having a problem with CI+Percy though, but it seems like it's not only with this PR 😅 

Update: I initially approved it, but after getting a bit more of context, suggested some changes

src/Nri/Ui/Data/PremiumLevel.elm
```diff
@@ -44,9 +45,12 @@ order : PremiumLevel -> Int
...
     case privileges of
         PremiumWithWriting ->
-            2
+            3
 
         Premium ->
+            2
+
+        PremiumPromo ->
             1
```
celsobonutti, line 54: I think it's better to use `3` for PremiumPromo and leave the rest of the values as-is, since that's [what we're going for in the backend](https://linear.app/noredink/issue/GRO-173#comment-f559d70e)

```suggestion
        PremiumPromo ->
            3
            
        PremiumWithWriting ->
            2
            
        Premium ->
            1
```

celsobonutti (approved):

bcardiff (approved):
I rerun the CI since it failed, but now it's happy :-)

I think the order implementation is fine since it's not exported, and is used to implement `allowedFor`, `highest`, `lowest`. It doesn't seem to be the case that that number is used.

On the app that is another story. It would be nice to keep the values aligned, but it is not mandatory.

bcardiff:
That being said, I think it would make sense to remove PremiumLevel in the long run from this repo, is more about our app rather than the ui components. Would be great to come with some other abstraction, eventually.

PR #847, opened 2022-03-02
Changes 'PremiumLevel.allowedFor' logic to reflect Rails' changes

celsobonutti:
Changes `PremiumLevel.allowedFor` logic to reflect changes in Rails.

Related to https://github.com/NoRedInk/NoRedInk/pull/37316.

### DIFF
--- a/src/Nri/Ui/Data/PremiumLevel.elm
+++ b/src/Nri/Ui/Data/PremiumLevel.elm
@@ -19,7 +19,18 @@ type PremiumLevel
 -}
 allowedFor : PremiumLevel -> PremiumLevel -> Bool
 allowedFor requirement actor =
-    order requirement <= order actor
+    case requirement of
+        Free ->
+            True
+
+        PremiumPromo ->
+            True
+
+        Premium ->
+            actor /= Free
+
+        PremiumWithWriting ->
+            actor == PremiumWithWriting
 
 
 {-| The highest premium level in a list


tesk9 (commented):
Hiya! I happened to see this PR -- I am soo excited to see changes to premium modeling! There's so much potential here in terms of making premium states easier to understand!

I wanted to share a little bit of context on noredink-ui versioning and on historical premium modeling.

### noredink-ui versioning

There's some more info on the approach [we're taking in the readme](https://github.com/NoRedInk/noredink-ui#versioning-policy), but essentially we aim to avoid ever making breaking or major changes to an existing component. (I do realize a previous PR added `PremiumPromo`, rather than this one).

If this change merges, and is released, and you try to upgrade noredink-ui to the next major version in the monolith, you'll have to do _all_ of the changes at once. For instance, modules like [this one](https://github.com/NoRedInk/NoRedInk/blob/master/monolith/ui/src/Data/PremiumLevel.elm) will need to be updated in the same PR as the elm.json change.

Using the same Nri.Ui.Data.PremiumLevel also keeps code like this that doesn't fully make sense with `PremiumPromo`:

```
{-| The highest premium level in a list
-}
highest : List PremiumLevel -> Maybe PremiumLevel
highest privileges =
    privileges
        |> List.sortBy order
        |> List.reverse
        |> List.head


{-| The lowest premium level in a list
-}
lowest : List PremiumLevel -> Maybe PremiumLevel
lowest privileges =
    privileges
        |> List.sortBy order
        |> List.head
```

What does highest or lowest privilege _mean_ when comparing PremiumPromo and Premium?

I highly recommend adding a new module (maybe `Nri.Ui.Data.PremiumLevel.V2`?) and redoing the premium modeling for the frontend from scratch.

This will require adding helpers or new versions to some of the other noredink-ui components in order to work with the new PremiumLevel, but you will be able to upgrade to them one at a time instead on en masse.

### Premium vs PremiumWithWriting

When NRI first released WritingCycles (the precursor to Peer Reviews, and the first of the writing assignments), they were only available to Premium users who had Writing access as an add-on.

There is no longer a non-writing Premium level. If you have Premium access, you have access to all features in writing assignments.

I would love for an updated Premium modeling to no longer describe the world of 5 years ago 😅 !

### User Premium Level, Course Premium Level, and Content Premium Level

Right now, on the Elm side, we model user level, course level, and content level all with the same types.

I think this has been confusing for some time (I always have to check the argument order when using a helper like `allowedFor` -- is it the content or the actor who comes first???), but I think adding PremiumPromo on to this may compound the issue.

I think in the context of a _user_, a "PremiumPromo" would imply that the user has access to Premium temporarily! 🎉 

But in the context of _content_, a "PremiumPromo" would imply that free users have access to usually-Premium content!

😬 I'm worried continuing to share the same types will get more confusing!

-- 

Feel free to do what you wish with this info/feedback. I don't want to be a roadblock and a drive-by commenter at the same time!

I'm also happy to talk more about this sync with you or with the team!

tesk9:
> I would even like to get rid of PremiumLevel on noredink-ui actually. It doesn't seems to be a ui concern. I would expect adding locks, disabling elements to be information used by noredink-ui. PremiumLevel seems more of an app modelling concern.

Yeah I buy this! I think I added it because I wanted to avoid APIs like:
```
view : {  showPremiumIcon : Bool, isLocked: Bool ....}
```

in favor of:

```
view : { contentPremiumLevel: PremiumLevel, userPremiumLevel: Bool }
```

There's some additional rules around _when_ premium icons show: currently, we want to show premium icons to Free and Non-Site License Premium users, so that teachers with a mix of premium and non premium content don't get to the form and then discover they can't assign the content they selected to the course they want...

So maybe a `PremiumDisplay` type would be appropriate for noredink-ui?

```

It seems like there's a decision point here around 
1. keep some premium data modeling in noredink-ui, and add a new PremiumLevel file 
2. consider Nri.Ui.Data.PremiumLevel as just an API concern for the existing components, move the real modeling back to the monolith, and aim to clean up the component APIs such that data premium level isn't necessary in the future.

Does that sound right?

**Comment by bcardiff** [comment]
I agree with the decision point. I think in the long run I want 2, `PremiumDisplay` sounds good to me.
The effort of making a PremiumLevel.V2, SideNav.V2, etc makes me think to make the long run now and not wait :-). 

@caseyWebb, what do you think?

**Comment by caseyWebb** [comment]
> The effort of making a PremiumLevel.V2, SideNav.V2, etc makes me think to make the long run now and not wait

Agreed; if there is going to be friction and extra work required either way then I think it's best to make the jump now.

Regarding `PremiumDisplay`, this would be something like

```
type PremiumLevel
  = Free
  | PremiumUnlocked
  | PremiumLocked
```

?

And if so, to recap:

- Move `PremiumLevel` modeling to monolith and split into `UserPremiumLevel`, `ContentPremiumLevel`, and `CoursePremiumLevel`
- Add `PremiumDisplay` type to noredink-ui
- Add helpers to monolith to turn `XPremiumLevels` into `PremiumDisplay`
- Create versioned UI components that use `PremiumDisplay`
- Plumbing 🚽 👨🏼‍🔧 

?

bcardiff:
@caseyWebb 

> Move PremiumLevel modeling to monolith and split into UserPremiumLevel, ContentPremiumLevel, and CoursePremiumLevel

I think this is optional / nice-to-have and is entirely on the monolith.

All the rest is good! Let's start by merging #848 and introduce `PremiumDisplay` with the new versioned components 🚀

celsobonutti:
Should we close this PR and start the work on PremiumLevel.V2 on another one?

PR #850, opened 2022-03-07
Introduces PremiumDisplay

celsobonutti:
This PR introduces the `PremiumDisplay` module/type.
It also makes new versions for all the elements that previously used `PremiumLevel` in some way, which are:
* SideNav: V1 -> V2
* RadioButton: V3 -> V4
* PremiumCheckbox: V7 -> V8

Examples were also updated, to use the latest version of each component


tesk9 (commented):
This is looking good!

One quick note -- on the examples, the version # should correspond to the version # for the code. (The checkbox/premium checkbox code is funky -- I don't have a strong opinion on what should happen with that). What this video with sound to get what I mean better, hopefully:

https://user-images.githubusercontent.com/8811312/157305036-b2c303a3-f1ed-476e-9f5e-63a2220d765e.mov

It would be nice to have tests against this, so a person didn't have to know to do this. Someday!

celsobonutti:
Hey @tesk9, just updated the version number 😊 

On another side: what do you mean by funky? I think we can work on this in the future 👀

tesk9:
Thanks for updating! 🙏 

> what do you mean by funky?

Yeah!  So the "funky" thing about the Checkbox example is that there's 2 modules being demoed, so it's not clear that PremiumCheckbox and Checkbox are 2 separate modules.

Marked-up code screenshot showing that PremiumCheckbox is imported in the example, but it's not used in the example set up

We end up with Checkbox.V5 showing up on the examples page, and you kinda have to know that PremiumCheckbox exists separately. Or you have to look at the example source, which is not as convenient for a dev.

Screenshot of https://noredink-ui.netlify.app/#/doodad/Checkbox illustrating the misleading Checkbox.V5 header

---

This is looking good to me! I'm happy to have this merge as long as Brian's ready for it to merge!

src/Nri/Ui/RadioButton/V4.elm
```diff
@@ -0,0 +1,639 @@
...
+                        else
+                            config.labelCss
+                    ]
+                    [ Html.text label ]
+                , case ( config.premiumDisplay, config.premiumMsg ) of
+                    ( PremiumDisplay.Free, _ ) ->
+                        text ""
+
+                    ( _, Nothing ) ->
+                        text ""
```
bcardiff, line 412: This is fine and can be changed later, but in the spirit of showing that an option is PremiumUnlocked, shouldn't we show the pennant if the config.premiumDisplay is PremiumDisplay.PremiumUnlocked or PremiumDisplay.PremiumLocked ? This would be a non clickable pennant.

src/Nri/Ui/SideNav/V2.elm
```diff
@@ -0,0 +1,450 @@
...
+
+viewSidebarEntry : Config route msg -> List Css.Style -> Entry route msg -> Html msg
+viewSidebarEntry config extraStyles entry_ =
+    case entry_ of
+        Entry children entryConfig ->
+            if entryConfig.premiumDisplay == PremiumDisplay.PremiumLocked then
+                viewLockedEntry extraStyles entryConfig
+
+            else if anyLinkDescendants (isCurrentRoute config) children then
+                div [ Attributes.css extraStyles ]
```
bcardiff, line 133: Similar to the RadioButton changes, this is not introducing visual changes for PremiumUnlocked. If that is to land in another PR then we are good here.

bcardiff (approved):
LGTM as long as this PR should not cover visual changes for PremiumUnlocked

src/Nri/Ui/RadioButton/V4.elm
celsobonutti: I believe it'd be better to introduce the visual changes already, since that also implies on API changes:
`showPennant`, for example, doesn't make sense if the pennant is always shown for premium content, a name like `onClickPennant` would make more sense.

src/Nri/Ui/SideNav/V2.elm
celsobonutti: This visual change, on the other side, I think can be left for another PR, as it shouldn't change anything in the module's API

bcardiff:
We might rename `onLockedPennantClick` to `onLockedClick` when we make the entire control clickable on locked state.

Issue #843, opened 2022-02-24
Add Shadow.V1 


tesk9:
See https://github.com/NoRedInk/noredink-ui/pull/842

How many different types of box shadow should a project have? Different "intensities", maybe?

bendansby:
My answer to your question: https://paper.dropbox.com/doc/UI-Style-Guide-and-Caveats--Bchrvv5_p4rudpwMRawcKJTEAg-PvOLxeX3oyujYEzdJx5pu#:uid=984733103055528053538389&h2=Shadows

Cribbed from https://www.joshwcomeau.com/css/introducing-shadow-palette-generator/

PR #853, opened 2022-03-10
Adds Shadows

tesk9:
For https://github.com/NoRedInk/noredink-ui/issues/843

Note that this PR doesn't yet _use_ the shadows.

Shadow example
Preview example

@NoRedInk/design


bendansby:
So cool, thank you!

PR #856, opened 2022-03-11
Improve premium radio buttons locked clicks

bcardiff:
- Settle on API for onLockedClick to match #855
- Improves the clickable area of locked radio buttons to have similar UX as premium checkbox, and not just the pennant

gL6Kgbs1Cz

### DIFF
--- a/src/Nri/Ui/RadioButton/V4.elm
+++ b/src/Nri/Ui/RadioButton/V4.elm
@@ -1,6 +1,6 @@
 module Nri.Ui.RadioButton.V4 exposing
     ( view
-    , premium, onLockedPennantClick
+    , premium, onLockedClick
     , disclosure
     , onSelect
     , Attribute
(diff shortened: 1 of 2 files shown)


tesk9 (commented):
I think the locking/disabling behavior isn't quite right -- the underlying disabled input is still selectable.

I added `css styles =  Extra.none` in the RadioButton file in order to see the state of the underlying radio input, and I found I could still select the disabled input.

radio interactions

or maybe I'm misunderstanding something...?

bcardiff:
I wonder, would it be too bad to avoid having a radio and a checkbox at all when they are locked?

Disabled checkbox on master also have some similar issues, yet those are not new I think.

r83kuvYc8l


Another alternative could be to see if the container could receive the click instead of the control when they are locked, and keep the control disabled.

tesk9:
I'm not sure which alternative is the better move. 🤔 I know FAB friday folks talked through some of the disabled states a while back, but it's been a while... I'd definitely like to check at some point whether previous Checkbox versions worked better, or if this is a more recent problem.

I think this might be good to talk through synchronously, maybe with Kristine and Ben too. @celsobonutti will you be taking over this PR next week?

celsobonutti:
hey @tesk9, sorry, I was already AFK on Friday
Sure, I can take over this PR and will start working on it today. Feel free to invite me to any meetings ☺️

tesk9:
Hmm! Definitely not ideal, but it is what we had in previous versions of the components. 🤔 I'll schedule some synch time for us to discuss!

tesk9 (approved):
Ahh I'm happy with this approach! Thanks for implementing!

PR #857, opened 2022-03-11
Button.V10: adds ternary style

bcardiff:
A new button style 💅 !

Screen Shot 2022-03-11 at 16 43 00


bendansby (approved):
Nice!

bendansby:
To be pedantic, according to the dictionary, it should be "tertiary" 🙈

tesk9:
Code looks really good!

I am thinking we may want to rename the `secondary` and `primary` attributes on `SideNav`, seeing this PR.

`SideNav.secondary` looks very much like `Button.tertiary` 🤔  [See the deploy preview here.](https://deploy-preview-858--noredink-ui.netlify.app/#/doodad/SideNav)


Should I change SideNav.V2's `secondary` to `tertiary`, do you think? Or does `SideNav.secondary` still make sense since it's the non-default style in the SideNav?

tesk9 (approved):
Cool! This is good to merge then!

Issue #817, opened 2022-01-24
Sidebar items should wrap rather than expand the width of the sidebar


bendansby:
We don't want this:
<img width="1071" alt="image" src="https://user-images.githubusercontent.com/13528834/150837954-d313c234-eb6b-4d05-958e-f5cd16295ec1.png">

PR #859, opened 2022-03-11
Tessa/sidebar wrap

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/817 (At least partially. The sidebar nav is set up to expand to the full width of its parent container by default. If the /curriculum library is using flexbox and permitting flex grow on the sidebar, then the sidebar may keep growing!)

## Before
the height of each sidebar element was hardcoded to 45px, meaning tons of ugly overflowing content for long text.

## After
the height of each sidebar element is 45px if the content isn't multline, but padding is used to achieve this rather than a hardcoded height value.

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/SideNav/V2.elm
+++ b/src/Nri/Ui/SideNav/V2.elm
@@ -241,7 +241,8 @@ viewLockedEntry extraStyles entryConfig =
 sharedEntryStyles : List Style
 sharedEntryStyles =
     [ padding2 (px 13) (px 20)
-    , Css.property "word-break" "break-word"
+    , Css.property "word-break" "normal"
+    , Css.property "overflow-wrap" "anywhere"
     , displayFlex
     , borderRadius (px 8)
     , alignItems center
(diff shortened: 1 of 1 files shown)


bendansby:
Thank you!

src/Nri/Ui/SideNav/V2.elm
```diff
@@ -240,9 +240,8 @@ viewLockedEntry extraStyles entryConfig =
 
 sharedEntryStyles : List Style
 sharedEntryStyles =
-    [ paddingLeft (px 20)
-    , paddingRight (px 20)
-    , height (px 45)
+    [ padding2 (px 13) (px 20)
+    , Css.property "word-break" "break-word"
```
lindsaykwardell, line 244: Looks like MDN says `word-break: break-word` is deprecated, how do you feel about using the alternate suggestions?

image

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values

tesk9: Great catch, thank you!!! 🙏

lindsaykwardell (approved):

PR #864, opened 2022-03-16
Use 'Html.button' element for PremiumLocked checkboxes

celsobonutti:
This change fixes the problem introduced in #855, where locked/disabled checkboxes were acting as enabled underneath.

The solution, as discussed in a meeting, was changing these checkboxes into buttons.

I also have a question: do you think it make sense to actually have PremiumLocked Disabled checkboxes? I didn't take those in account, but they can be easily implemented.

### DIFF
--- a/src/Nri/Ui/Checkbox/V5.elm
+++ b/src/Nri/Ui/Checkbox/V5.elm
@@ -2,8 +2,9 @@ module Nri.Ui.Checkbox.V5 exposing
     ( Model, Theme(..), IsSelected(..)
     , view, viewWithLabel
     , selectedFromBool
+    , checkboxLockOnInside
+    , viewIcon
     )
-
 {-|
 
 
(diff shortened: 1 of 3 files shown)


tesk9 (approved):
> do you think it make sense to actually have PremiumLocked Disabled checkboxes

Hmm! Interesting. I'm trying to think when this would apply... I know that on the assignment form, a teacher with a non-site premium license who is attempting to edit who is assigned an assignment, we use the premium checkbox for the course. So a course checkbox might be disabled and displayed as premium, but not locked. Although if the premium level expired, perhaps then the locked upsell state might make sense?

I think don't block on it. We need to do more thinking on this front in general, but your team can keep taking incremental steps towards better premium modeling without solving everything at once. If we're wrong, and it's necessary, worst case is we need a new noredink-ui release.

Does that sound good to you?

celsobonutti:
Sure it does! Thanks, Tessa! 😊

PR #865, opened 2022-03-16
Show premium pennant on PremiumUnlocked sidenav entries

caseyWebb:
Adds premium pennant to sidebar entries with

```elm
SideNav.premiumDisplay PremiumDisplay.PremiumUnlocked
```

2022-03-16-172728_1854x1239_scrot
8

### DIFF
--- a/src/Nri/Ui/SideNav/V2.elm
+++ b/src/Nri/Ui/SideNav/V2.elm
@@ -48,7 +48,8 @@ import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Data.PremiumDisplay as PremiumDisplay exposing (PremiumDisplay)
 import Nri.Ui.Fonts.V1 as Fonts
 import Nri.Ui.Html.Attributes.V2 as ExtraAttributes
-import Nri.Ui.Html.V3 exposing (viewJust)
+import Nri.Ui.Html.V3 exposing (viewIf, viewJust)
+import Nri.Ui.Pennant.V2 as Pennant
 import Nri.Ui.Svg.V1 as Svg exposing (Svg)
 import Nri.Ui.UiIcon.V1 as UiIcon
 
@@ -142,7 +143,21 @@ viewSidebarEntry config extraStyles entry_ =
                                ]
                         )
                         []
-                        [ text entryConfig.title ]
+                        [ viewIf
+                            (\_ ->
+                                Pennant.premiumFlag
+                                    |> Svg.withWidth (px 25)
+                                    |> Svg.withHeight (px 19)
+                                    |> Svg.withCss
+                                        [ marginRight (px 10)
+                                        , position relative
+                                        , top (px 3)
+                                        ]
+                                    |> Svg.toHtml
+                            )
+                            (entryConfig.premiumDisplay == PremiumDisplay.PremiumUnlocked)
+                        , text entryConfig.title
+                        ]
                         :: List.map (viewSidebarEntry config [ marginLeft (px 20) ]) children
                     )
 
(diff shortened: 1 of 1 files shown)


caseyWebb:
There was a linear issue for this but no mockups. Looping @bendansby in to get 👀 on it.

bendansby:
Thanks for the ping @caseyWebb 

@steder @tilliemcinnis It is logical to have this, but seems a bit much. Thoughts?

tesk9:
Which users would this premium display be aimed at? Just the currently-free users and anonymous users? 

We offer premium access on a per-class basis for teachers with non-site licenses, which is why it is necessary to have the premium pennant on premium topics for these teachers early in the flow (so they don't get to the final step of the form and get frustrated that they can't assign the content that they wanted to assign to the given course).

However, for writing assignments, we do not limit access by the premium level of the course. If you're a teacher at a school or district with premium, you get premium premades. All premade planning diagnostics are also free, because all planning diagnostics are free.

<details><summary> Free teachers and anonymous users are already upsold on the AL from the sidenav. It's kind of annoying to get a sense of on prod or staging at the moment because of the free premades promotion... so unfold this for a demo gif</summary> 

premium locks on AL

</details>

Here is a screenshot of the settings for the sidenav example:
image

I feel like I'm rambling!

My core concern: **I'm worried about potentially using the "Premium" pennant to demo _unlocked_ premium content, because for many teachers, this icon already means "This content is premium and so cannot be assigned to your free classes"**

If you _do_ decide to go this direction, could you please add this to the styleguide example for the SideNav component before mergin? (Feel free to schedule some time to pair on it with me if you want! The Debug.Controls pattern is different than the rest of our Elm code -- it's aimed at dev convenience not at robustness)

caseyWebb:
> Which users would this premium display be aimed at? Just the currently-free users and anonymous users?

The current thinking is to highlight premium content for _everyone_, including premium users. See [this helper](https://github.com/NoRedInk/NoRedInk/blob/gro-208-migrate-premiumlevel-from-noredink-ui-to/monolith/ui/src/Data/PremiumLevel.elm#L79). However I see how that could cause confusion with regards to non-site license teachers being dissuaded from using content that is actually available to them.

We could choose to conditionally use `SideNav.premiumDisplay` based on the user's premium level in the writing section and not apply it, but that would also have to be done ([here](https://github.com/NoRedInk/NoRedInk/pull/37463).

But, if we're not going to show the premium pennant for premium users in some places, I don't think it makes sense to show it in others. Teachers and admins may see the premium flag on the grammar practice and expect it to be on all premium content, so hiding in in the writing may make it appear as if all the writing assignments are free, devaluing premium.

That said, seems like the best approach _is to_ only show for free and anonymous users, and that resolves this? Does something like the following make sense?

```elm
type PremiumDisplay
  = Promo      -- Flag Icon
  | Unlocked   -- Nothing
  | Locked     -- Lock Icon & Disabled
```

```elm
display : Authorization -> PremiumDisplay
display authorization =
    let
        unlockedIfAllowed =
            if allowedFor authorization.requirement authorization.actor then
                PremiumDisplay.Unlocked

            else
                PremiumDisplay.Locked
    in
    case authorization.requirement of
        Free ->
            PremiumDisplay.Unlocked

        PremiumPromo ->
            PremiumDisplay.Promo

        Premium ->
            unlockedIfAllowed

        PremiumWithWriting ->
            unlockedIfAllowed
```

tesk9:
That makes sense to me! @bendansby , thoughts?

caseyWebb:
I discussed with marketing (@LizVudhiwat) and consensus is to hold on the sidebar changes for now and re-evaluate after the free GD/QW promotion ends

Slack conversation [here](https://noredink.slack.com/archives/C021C0TCN0M/p1647540771574319)

PR #866, opened 2022-03-17
Bat/casts a shadow

tesk9:
Use the `Shadows` component where the same box-shadow is already used. This PR is basically a find-and-replace for previously hardcoded shadow values. There shouldn't be any visual changes.


micahhahn (approved):
That is.... so much nicer!

PR #867, opened 2022-03-18
Generate Tooltip example code

tesk9:
image

### DIFF
--- a/styleguide-app/CommonControls.elm
+++ b/styleguide-app/CommonControls.elm
@@ -81,7 +81,7 @@ httpError =
 content :
     { moduleName : String
     , plaintext : String -> attribute
-    , markdown : String -> attribute
+    , markdown : Maybe (String -> attribute)
     , html : List (Html msg) -> attribute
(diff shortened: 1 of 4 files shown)


asterite:
I'm reviewing this and I have an unrelated question... why isn't direnv the default choice? Reading the README I found I need to run `script/develop.sh` to try things out, but that didn't work. Then it says you can run `nix-shell`. And the last option is to use direnv if you don't want to go through that. So two things: why direnv isn't the default, and why the docs are backwards? Just asking to see if the readme could be improved.

asterite:
Wow, just now I realized there's a "Generate code" section. I always go to the docs to see how to use these, and in my mind I always complained that you could adjust things in the UI but the relevant code was nowhere to be found.

I wonder if it would be work asking others if they know about this. I missed it because it's initially collapsed, and flushed to the right, so it's pretty hidden (I'm used to reading things from left to right, and the left side is already pretty packed)

asterite (approved):
Looks great!

PR #869, opened 2022-03-21
Icon examples

bendansby:
Cleans up the display of icon examples.

image

### DIFF
--- a/script/percy-tests.js
+++ b/script/percy-tests.js
@@ -36,6 +36,19 @@ describe('Visual tests', function () {
     console.log(`Snapshot complete for ${name}`)
   }
 
+  const iconProcessing = async(name, location) => {
+    await page.goto(location)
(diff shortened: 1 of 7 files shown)


bendansby:
@tesk9 I hid the icon names because I find it easier to scan, though obviously the names have value for reference and searching. How hard would it be to add a show/hide name toggle?

tesk9:
I _gasped_ this is so pretty!

I don't think it would be hard to add a "Show names" checkbox! Shall I branch off this and add that?

tesk9:
I added a checkbox to every icon display page that controls whether to show the names or not, and I added some percy snapshots for when we're showing the icon names! 😄 

I didn't change any styles or anything like that

bendansby:
Wooo thanks so much! Love it

tesk9 (approved):

PR #872, opened 2022-03-21
Fix performance icon viewbox

bendansby:
So the icon isn't smol

https://user-images.githubusercontent.com/13528834/159297350-2b621eaa-4fe6-40fb-a664-863f636607a4.mov


tesk9 (approved):
Woo great fix thanks!

PR #874, opened 2022-03-21
Adds class name to premium checkbox flag

celsobonutti:
This change is part of a fix for [an alignment issue in curriculum](https://linear.app/noredink/issue/GRO-231/fix-alignment).

The Curriculum Pathway view was using a selector based on the `aria-label` for the Premium flag, which caused problems when it either didn't exist or had no `label` (used for `aria-hidden=true`). 

The solution was to add a hidden flag for the Free PremiumCheckboxes, and a class to be used on the page's selector.

### DIFF
--- a/src/Nri/Ui/PremiumCheckbox/V8.elm
+++ b/src/Nri/Ui/PremiumCheckbox/V8.elm
@@ -45,7 +45,6 @@ import Nri.Ui.Svg.V1 as Svg exposing (Svg)
 import Nri.Ui.Util exposing (removePunctuation)
 import String exposing (toLower)
 import String.Extra exposing (dasherize)
-import Svg.Attributes
 
 
 {-| Set a custom ID for this checkbox and label. If you don't set this,
--- a/src/Nri/Ui/PremiumCheckbox/V8.elm
+++ b/src/Nri/Ui/PremiumCheckbox/V8.elm
@@ -265,24 +265,25 @@ viewLockedButton { idValue, label, containerCss, onLockedMsg } =
 
 viewPremiumFlag : { hidden : Bool } -> Html msg
 viewPremiumFlag { hidden } =
+    let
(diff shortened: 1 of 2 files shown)


caseyWebb (approved):
🍐

tesk9 (commented):
The problem on the assignment library side is that we absolutely position the pennant, and in the new version of the checkbox, we are not successfully targeting the blank spot, right? I'm wondering if we can instead refactor to avoid absolutely positioning the flag. 

I played with this a bit in [this commit](https://github.com/NoRedInk/NoRedInk/commit/b1592e8c9568ab04c26df25ed59e347220abacc3). It seems like this approach would work, with a couple of caveats:
- it seems that the premium pennant/the empty space for the checkbox needs flexShrink zero applied to it (or it needs to set min-width explicitly), since it's in a flex container.
- the blue background color fo the row would apply to the premium pennant too I don't think this is necessarily a downside, but it would be a change. See screenshot:.

<details> <summary>Screenshot showing larger blue row</summary>

image


</details>

If you do want to stick with a selector-based approach, I wonder if it would make sense to expose the new class value? I'm thinking: if we're relying on external users of this library to know about the "premium-checkbox-flag-V8" string, then it should be part of the API. Also, if it's exposed, I think noredink-ui editors will be less likely to change it without checking for uses on the monolith side. Additionally, maybe a comment explaining more about why the class exists?

celsobonutti:
I 100% agree with avoiding absolute positioning tbh. I like your solution, and don't think using flex-shrink 0 is necessarily a problem.

Regarding the visual change: I don't think that's a problem, but I prefer to have @bendansby reviewing that 👀 

Aside from the selector approach, I was also wondering if it would be a good idea to expose a function to change the flag style (as we have for labels on Checkbox/RadioButton)

PR #878, opened 2022-03-22
Bat/disclosure indicator

tesk9:
Adds generated code example to DisclosureIndicator.

### DIFF
--- a/styleguide-app/Debug/Control/Extra.elm
+++ b/styleguide-app/Debug/Control/Extra.elm
@@ -1,12 +1,14 @@
 module Debug.Control.Extra exposing
     ( float, int
     , list, listItem, optionalListItem, optionalListItemDefaultChecked, optionalBoolListItem
+    , bool
     )
 
 {-|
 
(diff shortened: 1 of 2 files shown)


zwilias (approved):
The change in isolation looks good to me!

In broader terms, the `( String, a )` pattern representing "generated code and actual value" seems like it could be made somewhat nicer. To some extent Elm will helpfully guide people to using the right combination of `Tuple.first` and `Tuple.second`, yet it doesn't seem very ergonomic.

Perhaps we can do a nice refactoring at some point to either lift generated code into controls directly, or have a `WithCode a` type thing where accessing values doesn't need to go through `Tuple.{first,second}`.

It feels like it's worth continuing as is for a bit, tho, and get a better feel for the patterns that emerge so a nice, concise API making those patterns easier to grok can be designed!

Long story short: go for it 😄 

goforit

tesk9:
❤️ Yeah I totally agree! Not ergonomic yet, but there's potential!

PR #880, opened 2022-03-22
Adds flex-shrink: 0 to checkbox premium flags

celsobonutti:
Necessary to fix an alignment problem in Assignment Library (more context in #874)


tesk9 (approved):
Looks great, thanks!

PR #885, opened 2022-03-23
Ellie example links proof of concept

tesk9:
This is a (lightly brittle) proof of concept of direct links to Ellie with examples.

Now, if there's generated code to display, a link to Ellie with packages pre-installed and a minimal code example will be in place!

Why I say this is brittle:
- we must specify specific package versions when linking to Ellie. This means that information encoded in the link may get out of sync with the elm.json. In particular, the version of noredink-ui that's sent over to Ellie may not be the most up to date.
- we don't yet have testing set up for the generated code examples. This is something that I hope to do, but, you know, one step at a time and all that. So if the code example has a typo in it, then of course the Ellie won't work either
- Ellie doesn't have access to unpublished components. So while adding a new component or new part of an API, the Ellie example won't work

Why I think this is cool:
- interacting with code in a simple environment can be really useful for learning! It may be easier for devs to explore APIs by way of code, rather than through the Debug.Control interfaces. Having both avenues available will hopefully meet lots of people's needs!

https://user-images.githubusercontent.com/8811312/159812514-44293c28-de72-4561-8fc1-6d58b8455d18.mov


caiquefigueiredo:
@tesk9, was foxen properly tagged on this PR? I can review it, but I don't think I would be the best one to do it.

BrianHicks:
oh, sorry. It was that *I specifically* wanted to be on the review cycle 😆  I'll have a look at it later today unless this is a blocker to something else?

tesk9:
Oooooohhhhhh 😆 Okay my mistake! Thanks for clarifying! Later today would be lovely, thank you!

styleguide-app/Debug/Control/View.elm
```diff
@@ -62,6 +74,39 @@ viewExampleCode values =
...
+generateEllieLink example code =
+    Url.Builder.crossOrigin "https://ellie-app.com/a/example/v1"
+        []
+        [ Url.Builder.string "title" example.name
+        , Url.Builder.string "elmcode" (generateElmExampleModule example code)
+        , -- TODO: a system will be required to keep these values in line with the elm json values
+          Url.Builder.string "packages" "elm/core@1.0.5"
+        , Url.Builder.string "packages" "elm/html@1.0.0"
+        , Url.Builder.string "packages" "rtfeldman/elm-css@17.0.5"
+        , Url.Builder.string "packages" "NoRedInk/noredink-ui@15.8.1"
```
BrianHicks, line 91: what if we wrote something like `NoRedInk/noredink-ui@<NRI_UI_VERSION>` or something here, then in the Netlify build script we just did some `sed` call? That'd address some of the brittleness you were talking about, yeah?

BrianHicks: or, even better, we could pull it from page flags!

BrianHicks (approved):

styleguide-app/Debug/Control/View.elm
tesk9: I was thinking maybe we could have a build step that took the noredink-ui elm.json and processed it into JSON!

(This is partly why I tagged Foxen as reviewers on this particular PR -- I thought you all were likely to have neat ideas for improvement here!)

BrianHicks: oooh, that's a good idea. Lemme make a sketch in a new PR.

PR #886, opened 2022-03-24
Bat/menu example code

tesk9:
Adds basic partial code gen for Menu example. Doesn't support customizing entries, for instance.


styleguide-app/Debug/Control/View.elm
```diff
@@ -137,7 +137,20 @@ viewSection name children =
...
-        ++ "\n\t] "
+codeFromList =
+    codeFromListWithIndentLevel 1
+
+
+codeFromListWithIndentLevel : Int -> List ( String, a ) -> String
+codeFromListWithIndentLevel indent list =
+    let
+        indents =
+            String.repeat indent "\t"
```
ap-nri, line 148: This leaves us with invalid elm if we want to copy, but that's ok right?

ap-nri (approved):
@NoRedInkBot r+

ap-nri:
merged with a note about tabs being invalid Elm because that's just not a blocker

styleguide-app/Debug/Control/View.elm
tesk9: Thanks for this note! I have the tabs replaced in the Ellie version, but I think you're right that I should replace them in the rendered examples too.

PR #890, opened 2022-03-25
Bat/fix heading level

tesk9:
Example views should include an h1 and should be structured in a reasonable way.

Also, I kinda thought an underline might look nice???? @bendansby want to talk me out of it? 🙈


bendansby:
I'll see how I feel in a week. Easy enough to mysteriously disappear 😉

PR #891, opened 2022-03-28
sketch for loading JSONs

BrianHicks:
This is a little sketch for how we could maybe load all the dependencies we need for generating links. I think it would maybe make sense to just send Ellie all the package dependencies that we use in noredink-ui. Fairly straightforward, then: if the app does not compile due to a missing dependency, neither will the Ellies.

### DIFF
--- a/styleguide-app/EllieLink.elm
+++ b/styleguide-app/EllieLink.elm
@@ -1,9 +1,16 @@
-module EllieLink exposing (view)
+module EllieLink exposing (Config, view)
 
+import Dict exposing (Dict)
 import Example
 import Html.Styled exposing (..)
+import Http
 import Nri.Ui.ClickableText.V3 as ClickableText
 import Url.Builder
+
+
+type alias Config =
+    { packageDependencies : Result Http.Error (Dict String String)
+    }
 
 
 type alias SectionExample =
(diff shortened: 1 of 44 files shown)


tesk9:
@BrianHicks thanks for this awesome sketch! I carried it forward a bit, so that the Ellie links use the data you loaded. One thing I found -- using all the package dependencies made the Ellie too slow to get done with the initial compile, so I removed some of the package specifications.

styleguide-app/EllieLink.elm
```diff
@@ -14,36 +21,55 @@ type alias SectionExample =
...
+generateElmExampleModule config example =
+    let
+        maybeErrorMessages =
+            case config.packageDependencies of
+                Err httpError ->
+                    [ "{- "
+                    , "Something went wrong fetching the package dependencies!"
+                    , "You will need to install the packages by hand for this code to compile."
+                    , ""
+                    , "Error:"
```
BrianHicks, line 64: we might want a space after contextualizing that it's an error

```suggestion
                    , "Error: "
```

BrianHicks (commented):
looking great! One tiny suggestion, though.

I can't approve this since I authored it, but I think we're good to merge so here's a GIF!

tesk9 (approved):

PR #894, opened 2022-03-29
Bat/side nav label

tesk9:
Add `aria-label` support to the next SideNav version.

## API usage example:
```
SideNav.view
	{ isCurrentRoute = (==) "#some-route"
	, routeToString = identity
	, onSkipNav = SkipToContent
	}
	[ SideNav.navLabel "Limit search results"
	] 
	[ SideNav.entry "Quick Writes"
		[ SideNav.href "/search/quick-writes"
		]
	] 
```

### DIFF
--- a/src/Nri/Ui/SideNav/V3.elm
+++ b/src/Nri/Ui/SideNav/V3.elm
@@ -1,0 +1,449 @@
+module Nri.Ui.SideNav.V3 exposing
+    ( view, Config
+    , entry, entryWithChildren, html, Entry, Attribute
+    , icon, custom, css, nriDescription, testId, id
+    , onClick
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/SideNav/V3.elm
```diff
@@ -0,0 +1,530 @@
...
+    , onClick
+    , href, linkSpa, linkExternal, linkWithMethod, linkWithTracking, linkExternalWithTracking
+    , primary, secondary
+    , premiumDisplay
+    )
+
+{-|
+
+
+# Changes from V1
```
ap-nri, line 16: Do we keep these and add "Changes from V2?" or do we only do the most recent? Either way, we gotta add a thing here.

ap-nri (approved):
@NoRedInkBot r+

PR #895, opened 2022-03-29
Tike/confetti

steder:
Pairing with @tesk9 where we added a little bit of <3 to the confetti styleguide example.


tesk9:
(Rebased -- no merge queue/backmerge prior to test runs on this repo annoyance)

tesk9 (approved):

PR #898, opened 2022-03-30
Fix missing table header

tesk9:
Include the ths in the DOM even when they're not visually displayed so that screenreader users can navigate via the headings.


brian-carroll (approved):

brian-carroll:
@NoRedInkBot r+

PR #899, opened 2022-03-30
Specific page titles

tesk9:
Provide specific page titles for every route in the styleguide, implementing [Technique H25](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H25) for [Success Criterion 2.4.2](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-title.html).

### DIFF
--- a/styleguide-app/Main.elm
+++ b/styleguide-app/Main.elm
@@ -191,10 +191,7 @@ view_ model =
                         |> Html.map (UpdateModuleStates example.name)
 
                 Nothing ->
-                    Page.notFound
-                        { link = ChangeRoute Routes.All
-                        , recoveryText = Page.ReturnTo "Component Library"
-                        }
+                    notFound
 
         Routes.Category category ->
             withSideNav model.route
@@ -222,6 +219,14 @@ viewExample model example =
             { packageDependencies = model.elliePackageDependencies }
             example
         ]
+
+
+notFound : Html Msg
+notFound =
+    Page.notFound
+        { link = ChangeRoute Routes.All
+        , recoveryText = Page.ReturnTo "Component Library"
+        }
 
 
 withSideNav : Route -> List (Html Msg) -> Html Msg
--- a/styleguide-app/Main.elm
+++ b/styleguide-app/Main.elm
@@ -187,17 +187,8 @@ view_ model =
         Routes.Doodad doodad ->
             case List.head (examples (\m -> m.name == doodad)) of
                 Just example ->
-                    Html.div
-                        [ css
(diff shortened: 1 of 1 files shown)


BrianHicks (commented):
Code looks fine! "X in the NoRedInk Style Guide" seems really unusual though so I did some digging. https://accessibility.iu.edu/creating-content/web-content/titles.html has some good stuff, but the *most* interesting thing to me about that page is how they do their own titles. It looks like they're breadcrumbs, basically. I think it might be good to use that approach!

- `example.name ++ ": Examples: NoRedInk Style Guide"`
- `"Not Found: NoRedInk Style Guide"`
- `Category.forDisplay category ++ ": Category: NoRedInk Style Guide"`
- `"NoRedInk Style Guide"`

Overall that seems nice! Although I note the use of `:` as fairly unusual. I wonder if ` | ` as a separator would be acceptable to a screen reader. Looks like GitHub uses a &middot; (`\u00B7` I think?) and I assume they've done at least some research on what's best in a technical context.

BrianHicks:
wow, middot has a homepage https://middot.net/

tesk9:
For the Englishy page titles, I was following the suggestions in [the first stackoverflow answer here](https://stackoverflow.com/questions/24208808/accessibility-html-title-tag-separators-alternatives-to-the-vertical-line-p).

I think the `|` character will be understood by screenreader users, but for me anyway (as a very beginner user) it's kind of annoying/confusing, as VO reads it as "vertical line". 

Compare:

https://user-images.githubusercontent.com/8811312/160916145-b4cada64-e254-4236-ad10-c7da859f2886.mov

https://user-images.githubusercontent.com/8811312/160916131-c2cbacb5-de01-49b8-ada4-9e7793ca6ffa.mov

Does this change your opinion on using the character separator versus English connecting words? I will change it if you prefer the pipe or the colon

BrianHicks:
ah, bah, and it says "middle dot" for &middot;. I think probably colons or english make most sense, given that.

BrianHicks:
oh hey, some experimentation reveals hyphen as being maybe nice. It creates a clear visual hierarchy and voiceover pronounces it as a slightly longer gap between words.

tesk9:
> I think probably colons or english make most sense, given that.

Sounds good! I vagueeeelly prefer English, if that's alright?

BrianHicks (approved):

BrianHicks:
Yeah, totally great! Sorry if I’m coming off as pedantic—I’m just excited by the possibilities 😅

PR #900, opened 2022-03-30
Bump minor version

charlieshanley:
This bumps the minor version on account of the new icons added in #892 . After this is merged, I plan to tag and publish this version so I can refer to it in `NoRedInk`.


charlieshanley:
Wow, that was quick! Thanks!

PR #904, opened 2022-04-12
Bat/switch test

tesk9:
- adds elm program test support for testing against the styleguide app
- adds an example Switch test against the Switch component directly

Relevant to this conversation: https://noredink.slack.com/archives/C02NVG4M45U/p1649785774091919?thread_ts=1649720419.942899&cid=C02NVG4M45U


BrianHicks (changes requested):
it feels iffy that there's both a styleguide and library test that do the same thing. Was this what you intended? Can we just do the styleguide for future compatibility's sake?

tesk9:
Yeah totally! This is coming out of the [conversation with Derek about QA writing tests](https://noredink.slack.com/archives/C02NVG4M45U/p1649720419942899). So the test cases that I was writing were direct examples of the test he was asking about, written 2 ways as kind of an illustration.

I'll delete the non-styleguide `Switch` test. 👍

BrianHicks (approved):

PR #907, opened 2022-04-14
Fold axe and percy checks together

tesk9:
Merges the axe and percy tests. Some axe checks were lost when we changed the view to show cards rather than the full examples on the index page.

Following the ratcheting approach -- I'm skipping axe rules temporarily. I will fix them separately from this PR.


BrianHicks (approved):
looks fine to me!

PR #909, opened 2022-04-14
Bat/fix heading order

tesk9:
In many places on the styleguide, we were erroneously jumping from the h1 to h3s. This fixes, which has (EDIT: HAD) a styling implication (EDIT: before Ben & I discussed on this PR):

Navy bold h3s


bendansby:
I would consider the "small heading" style more the official h3, while tagline is more descriptive text for an h1.

tesk9:
Hmm okay! Maybe something to sort out in the next Heading version then! For now, do you think I should try to keep the preexisting navy styles/

bendansby:
Yeah that would look better 👍

tesk9:
Percy agrees that this looks the same except for the checkbox example headings which weren't previously styled at all. Merging! 🎉

Issue #916, opened 2022-04-20
Update a couple icons


bendansby:
- Lightbulb without sparklies (this should replace the existing lightbulb)
- Play arrow without circle (this should be a new icon)

There are icons below here. They're invisible. Good opportunity to use keyboard controls!
![play-triangle](https://user-images.githubusercontent.com/13528834/164306631-41c33872-730c-4659-ab60-bb154f3eaf04.svg)
![bulb](https://user-images.githubusercontent.com/13528834/164306637-ab838bbe-9676-4395-b8aa-7e8ed9ca93de.svg)

PR #917, opened 2022-04-20
Bat/new icons hurrah

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/916

Screen Shot 2022-04-20 at 1 58 32 PM
Screen Shot 2022-04-20 at 1 58 25 PM

@NoRedInk/design Look good to you? I'm so used to the glowing bulb. Looks like it got a real close buzzcut 😆


bendansby:
looks good!

staceyadams:
I'm so used to it in yellow that the gray is throwing me off. I'm seeing a hot air balloon now. (but also I'm running on very few hours of sleep)

Without the rays, it matches up better to the size/weight of our other icons.

tesk9:
Oh actually -- could you all double-check the Percy diff too? the default Message component looks pretty different without the rays

tesk9:
(not sure if you got notified again, sorry if this is a double-ping -- @NoRedInk/design )

staceyadams:
I don't think I've used Percy after they merged with browserstack. Who do I go to for access?

image

tesk9:
Following up in slack re Percy!

You can [preview the Message component here](https://deploy-preview-917--noredink-ui.netlify.app/#/doodad/Message)

The [published version is here ](https://noredink-ui.netlify.app/#/doodad/Message)

staceyadams:
I totally see buzzcut/ hot air balloon! I think it's fine though and will just take adjusting to.

As long as it's not this style of bulb, we're probably 👌 

image
image

tesk9:
Awesome, thanks!

PR #926, opened 2022-05-16
Support "disabled" links

tesk9:
Adds "disabled" link support by following the recommendations in [Scott O'Hara's "Disabling a link" post](https://www.scottohara.me/blog/2021/05/28/disabled-links.html).

Essentially, when the link is disabled:
- removes the href
- adds role link
- adds aria-disabled true

# Caveats

The styleguide example will NOT work correctly because of https://github.com/elm/browser/issues/34, which describes a problem where "a tags without href generate a navigation event".

However, I still expect this to work in the NoRedInk context in most cases, including the case that's the reason that we're adding "disabled" link support at all.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -98,6 +98,6 @@
         "avh4/elm-program-test": "3.3.0 <= v < 4.0.0",
         "elm/html": "1.0.0 <= v < 2.0.0",
         "elm-explorations/test": "1.2.2 <= v < 2.0.0",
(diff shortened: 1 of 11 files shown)


omnibs (approved):
The changes look good. The only request I'd have before a merge is, if you're game, adding the caveat and link to the elm/browser issue in the `disabled` function docs, so it's easy to find if anyone gets confused in the future.

Issue #921, opened 2022-05-09
Add Nri.Ui.Breadcrumbs example


bendansby:

tesk9:
Okay, I think it's actually done now!

PR #928, opened 2022-05-20
Add breadcrumbs

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/921 partway -- moves breadcrumbs into noredink-ui, and switches to an app design that uses the breadcrumbs, but does not implement a preview or an example.

Note that in this view, we are focused on a single component, but we're also still able to access the sidenav.

cc @NoRedInk/design trying out some puppychow -- what do you think?


bendansby:
Like it!

omnibs (approved):
Nice!

PR #931, opened 2022-05-23
Bat/improve page nav

tesk9:
When the route changes, focus the heading.

Fixes A11-789, and also improves the experience for screenreader users, since they'll be notified of the page heading.


ap-nri (approved):
@NoRedInkBot r+

PR #934, opened 2022-05-24
account for box-sizing: border-box in autoresize

BrianHicks:
Peer Review text inputs shrink because we are continually setting the minimum height 2px smaller because we do not currently account for the border. This fixes that by looking at the border if the computedStyles reveal that we're using `box-sizing: border-box` on the element.


tesk9 (approved):
Sweet, thanks!

BrianHicks:
seems to be a Percy flake. I'm going to merge this.

PR #936, opened 2022-05-24
enforce JS style with prettier

BrianHicks:
We do this in the monorepo but not in noredink-ui. This changes that!

@tesk9 requesting your review because you're in here more often than most (and IIRC doesn't a11ybats own nri-ui now?)


tesk9 (approved):
Love it, thank you!

PR #947, opened 2022-06-09
Add small content breakpoint

ArthurJordao:
This breakpoint is a recommendation from @bendansby to small contents. One use case for this is to remove stick position for some instructions on quiz engine, for example:
https://airtable.com/app1BuwWPTsZykPWh/pagO9Ogb7flpSdkCj?Inx0b=recPCDllBWeKBrvw2

### DIFF
--- a/src/Nri/Ui/MediaQuery/V1.elm
+++ b/src/Nri/Ui/MediaQuery/V1.elm
@@ -25,6 +25,9 @@ module Nri.Ui.MediaQuery.V1 exposing
 @docs quizEngineMobile
 @docs quizEngineBreakpoint
 
+@docs narrowMobile
+@docs narrowMobileBreakPoint
+
 -}
 
 import Css exposing (px)
--- a/src/Nri/Ui/MediaQuery/V1.elm
+++ b/src/Nri/Ui/MediaQuery/V1.elm
@@ -3,6 +3,7 @@ module Nri.Ui.MediaQuery.V1 exposing
     , mobileBreakpoint
     , quizEngineMobile
     , quizEngineBreakpoint
+    , narrowMobile, narrowMobileBreakPoint
     )
 
 {-| Standard media queries for responsive pages.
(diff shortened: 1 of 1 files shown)


tesk9:
Mind adding a little more context? Is 500px coming from @bendansby or someone else...?

ArthurJordao:
@tesk9 sorry, I tagged you but I was still adding description, but @bendansby asked me to add this breakpoint to noredink UI for fixing this issue:
https://airtable.com/app1BuwWPTsZykPWh/pagO9Ogb7flpSdkCj?Inx0b=recPCDllBWeKBrvw2

tesk9:
Got it, awesome! Great to see progress on the 320px viewport!

I wonder about the `smallContent` naming? It sounds like it's a description of, like... what the _content_ is, rather than the method-of-viewing the content. Maybe `narrowMobile`...?

ArthurJordao:
@tesk9 Agree, I like narrow mobile! One more thing, I just saw that there are some github actions that are falling, do you know what I did wrong?

tesk9:
Yup! So if you scroll down in the CI failure, you'll see that there's a problem with the doc comments. You'll need to add an `@docs` line for the new exposed helpers (like [here-ish](https://github.com/NoRedInk/noredink-ui/pull/947/files#diff-14aa9872bebc0e522ace04243864a393491cdace2a977fe15668079146b5c553L21))



One other thing you should know: I made a mistake on publishing timing about a week ago, which means the next release will be a major version. I'll try to get the major version published today after this merges, if that works for you?

tesk9 (approved):

ArthurJordao:
@tesk9 sounds go to me! Thank you :)

ArthurJordao:
@tesk9 I forgot to format it.

src/Nri/Ui/MediaQuery/V1.elm
```diff
@@ -73,3 +78,22 @@ quizEngineMobile =
...
 quizEngineBreakpoint =
     px 750
+
+
+{-| Styles using the `narrowMobileBreakPoint` value as the maxWidth
+-}
+narrowMobile : MediaQuery
+narrowMobile =
+    only screen
+        [ --`minWidth (px 1)` is for a bug in IE which causes the media query to initially trigger regardless of window size
```
bendansby, line 88: We might be able to get rid of IE fixes, since we no longer support it, but not sure if that violates "first, do no harm"

bendansby (approved):

src/Nri/Ui/MediaQuery/V1.elm
ArthurJordao: I don't know if violates. I am ok with it, the only thing is that it is a breaking change in the API behavior, and breaking changes are kind of evil 😈 . cc\ @tesk9 what are your thoughts on that?

tesk9: Eh I think either choice is fine. We'll almost certainly remove it in `MediaQuery.V2`. 🤷‍♀️ It's easiest not to change I guess, but we don't _need_ it

tesk9:
Not from your PR! I approved the diff -- Percy is messing up the Page/Modal screenshots sometimes, and I haven't tried to figure out what the issue is.

ArthurJordao:
@tesk9 ok! should we merge it?

tesk9:
YES! 🟢 🐎 🎉

PR #951, opened 2022-06-15
Adds a customizable Carousel

celsobonutti:
This PR adds a `Nri.Ui.Carousel.V1` module.
This is a simple Carousel that takes advantage of the `TabsInternal.V2` module and exposes it in a simple yet configurable way. 
Unlike other elements like `Tabs` and `SegmentedControl`, this module does not implement any styles, leaving it to the user.
A simple example of its functionality, copied from the our styleguide app example, can be seen below:

out

### DIFF
--- a/src/Nri/Ui/Carousel/V1.elm
+++ b/src/Nri/Ui/Carousel/V1.elm
@@ -1,94 +1,44 @@
-module Nri.Ui.Carousel.V1 exposing (..)
+module Nri.Ui.Carousel.V1 exposing (Item, buildItem, view)
 
 import Css exposing (..)
 import Html.Styled as Html exposing (Html)
-import Nri.Ui
-import Nri.Ui.Tooltip.V3 as Tooltip
 import TabsInternal.V2 as TabsInternal
 
 
-type Tab id msg
-    = Tab (TabsInternal.Tab id msg)
-
-
-type Attribute id msg
-    = Attribute (TabsInternal.Tab id msg -> TabsInternal.Tab id msg)
-
-
-type TabPosition
-    = Before
-    | After
-
-
-{-| Tooltip defaults: `[Tooltip.smallPadding, Tooltip.onBottom, Tooltip.fitToContent]`
--}
-withTooltip : List (Tooltip.Attribute msg) -> Attribute id msg
-withTooltip attributes =
-    Attribute (\tab -> { tab | tabTooltip = attributes })
-
-
-{-| Makes it so that the tab can't be clicked or focused via keyboard navigation
--}
-disabled : Bool -> Attribute id msg
-disabled isDisabled =
-    Attribute (\tab -> { tab | disabled = isDisabled })
+type Item id msg
+    = Item (TabsInternal.Tab id msg)
 
 
 {-| -}
-slideHtml : Html msg -> Attribute id msg
-slideHtml content =
-    Attribute (\tab -> { tab | panelView = content })
-
-
-labelledBy : String -> Attribute id msg
-labelledBy labelledById =
-    Attribute (\tab -> { tab | labelledBy = Just labelledById })
-
-
-tabHtml : Html Never -> Attribute id msg
-tabHtml content =
-    Attribute (\tab -> { tab | tabView = [ Html.map never content ] })
-
-
-{-| -}
-buildTab : { id : id, idString : String } -> List (Attribute id msg) -> Tab id msg
-buildTab config attributes =
-    Tab (TabsInternal.fromList config (List.map (\(Attribute f) -> f) attributes))
+buildItem : { id : id, idString : String, slideHtml : Html msg, controlHtml : Html Never } -> Item id msg
+buildItem config =
+    Item
+        (TabsInternal.fromList { id = config.id, idString = config.idString }
+            [ \tab -> { tab | panelView = config.slideHtml }
+            , \tab -> { tab | tabView = [ Html.map never config.controlHtml ] }
+            ]
+        )
 
 
 view :
     { focusAndSelect : { select : id, focus : Maybe String } -> msg
     , selected : id
-    , tabs : List (Tab id msg)
-    , tabStyles : Int -> Bool -> List Style
-    , tabListStyles : List Style
-    , containerStyles : List Style
-    , tabListPosition : TabPosition
+    , items : List (Item id msg)
+    , controlStyles : Bool -> List Style
+    , controlListStyles : List Style
     }
-    -> Html msg
+    -> { controls : Html msg, slides : Html msg }
 view config =
     let
         { tabList, tabPanels } =
             TabsInternal.views
                 { focusAndSelect = config.focusAndSelect
                 , selected = config.selected
-                , tabs = List.map (\(Tab t) -> t) config.tabs
-                , tabStyles = config.tabStyles
-                , tabListStyles = config.tabListStyles
+                , tabs = List.map (\(Item t) -> t) config.items
+                , tabStyles = always config.controlStyles
+                , tabListStyles = config.controlListStyles
                 }
     in
-    Nri.Ui.styled Html.div
-        "Nri-Ui-Carousel__container"
-        config.containerStyles
-        []
-        (case config.tabListPosition of
-            Before ->
-                [ tabList
-                , tabPanels
-                ]
-
-            After ->
-                [ tabPanels
-                , tabList
-                ]
-        )
+    { controls = tabList
+    , slides = tabPanels
+    }
--- a/src/Nri/Ui/Carousel/V1.elm
+++ b/src/Nri/Ui/Carousel/V1.elm
@@ -1,15 +1,34 @@
-module Nri.Ui.Carousel.V1 exposing (Item, buildItem, view)
+module Nri.Ui.Carousel.V1 exposing
+    ( view
+    , Item
+    , buildItem
+    )
+
+{-|
+
+@docs view
+@docs Item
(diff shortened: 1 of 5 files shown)


src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
...
+
+type Tab id msg
+    = Tab (TabsInternal.Tab id msg)
+
+
+type Attribute id msg
+    = Attribute (TabsInternal.Tab id msg -> TabsInternal.Tab id msg)
+
+
+type TabPosition
```
tesk9, line 18: Are you expecting to need both position styles? Or do you think the controls are likely always appear below the tabpanel?

Generally speaking, I'd rather not support use-cases we haven't encountered yet 🤔

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
...
+    = Before
+    | After
+
+
+{-| Tooltip defaults: `[Tooltip.smallPadding, Tooltip.onBottom, Tooltip.fitToContent]`
+-}
+withTooltip : List (Tooltip.Attribute msg) -> Attribute id msg
+withTooltip attributes =
```
tesk9, line 26: Are we expecting to display tooltips on the carousel controls? If not, I'd love to delete this 💀

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
...
+    Attribute (\tab -> { tab | tabTooltip = attributes })
+
+
+{-| Makes it so that the tab can't be clicked or focused via keyboard navigation
+-}
+disabled : Bool -> Attribute id msg
+disabled isDisabled =
```
tesk9, line 33: Are we expecting to disable carousel elements? (I'd assume we'd filter them out, rather than disable them...?) If not, I'd love to delete this 💀

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
...
+
+
+{-| -}
+slideHtml : Html msg -> Attribute id msg
+slideHtml content =
+    Attribute (\tab -> { tab | panelView = content })
+
+
+labelledBy : String -> Attribute id msg
+labelledBy labelledById =
```
tesk9, line 44: Do we definitely need this attribute?

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
...
+buildTab : { id : id, idString : String } -> List (Attribute id msg) -> Tab id msg
+buildTab config attributes =
+    Tab (TabsInternal.fromList config (List.map (\(Attribute f) -> f) attributes))
+
+
+view :
+    { focusAndSelect : { select : id, focus : Maybe String } -> msg
+    , selected : id
+    , tabs : List (Tab id msg)
+    , tabStyles : Int -> Bool -> List Style
```
tesk9, line 63: From an API perspective, I think we should consider adjusting `tabStyles` a bit.

1. It's a little bit surprising that the styles can change based on the _index_ of the carousel control (Although I'm not sure it's obvious that this is what is happening). I don't think it's typical for carousel controls to have differing styles based on index. 
2. Other details about the carousel control are managed by the Attributes pattern. I'd expect for the to follow either the config-heavy approach completely or to follow the Attributes-style approach completely.

This is the pattern I mean by `Attributes-style`:
```
css : List Style -> Attribute id msg
```

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
...
+    , tabListStyles : List Style
+    , containerStyles : List Style
+    , tabListPosition : TabPosition
+    }
+    -> Html msg
+view config =
+    let
+        { tabList, tabPanels } =
+            TabsInternal.views
```
tesk9, line 72: This is maybe a terrible suggestion? so definitely take it with a grain of salt.

But I'm wondering if it would make sense to return ` { tabList : Html msg, tabPanels : Html msg }` directly.

This way:
- we wouldn't need `containerStyles`
- we wouldn't need `tabListPosition`

and using the carousel would still be really flexible.

Just an idea!

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -0,0 +1,94 @@
+module Nri.Ui.Carousel.V1 exposing (..)
```
tesk9, line 1: I think we'll want to limit the exports before publishing

styleguide-app/Examples/Carousel.elm
```diff
@@ -0,0 +1,236 @@
...
+          , result = "Select the tab to the left of the currently-selected Tab"
+          }
+        , { keys = [ Arrow KeyboardSupport.Right ]
+          , result = "Select the tab to the right of the currently-selected Tab"
+          }
+        ]
+    , state = init
+    , update = update
+    , subscriptions = \_ -> Sub.none
+    , preview =
```
tesk9, line 165: If it's possible to add a little widget preview, this is where it would slot in

styleguide-app/Examples/Carousel.elm
```diff
@@ -0,0 +1,236 @@
+module Examples.Carousel exposing
```
tesk9, line 1: 🎉 😍

tesk9 (commented):
This is looking good so far!!!! Thank you for adding the new component and the new example! 😍 


I made some changes around the example, to extend what is configurable and to get a code sample wired up:


---

Do you think that `tab` will make sense to devs in the context of a Carousel? I wonder if the API should use different names... `control`? 🤔 I don't know what the bottom controls are usually called...

src/Nri/Ui/Carousel/V1.elm
celsobonutti: Deleted! 🎉

src/Nri/Ui/Carousel/V1.elm
celsobonutti: Deleted! 🎉

src/Nri/Ui/Carousel/V1.elm
celsobonutti: I don't think so, we can delete it~

src/Nri/Ui/Carousel/V1.elm
celsobonutti: 1. Agreed, I'll change it to be `Bool -> List Style`
2. I understand and agree, and would like to follow an Attributes-style approach, but, at the same time, I don't think there's a simple way to use in this case, as the `tabStyles` are passed to `TabsInternal.view`, which is responsible for rendering the actual controls. In any case, I would like to put some effort into this in the near future, but we are kind of in a hurry with the marketing website 😅

src/Nri/Ui/Carousel/V1.elm
celsobonutti: I... actually like it? I think it'd be good for this Carousel to be very customizable, since we're probably going to need it in different forms in very different places

styleguide-app/Examples/Carousel.elm
celsobonutti: Will do! 🎉

src/Nri/Ui/Carousel/V1.elm
celsobonutti: I think we can follow this: https://github.com/NoRedInk/noredink-ui/pull/951#discussion_r898583775
And remove this type 😊

src/Nri/Ui/Carousel/V1.elm
celsobonutti: +1

src/Nri/Ui/Carousel/V1.elm
tesk9: > What do you think about "control" for the tab and "slide" for the tab panel?

Sounds good!

> wondering if it makes sense to go config-heavy

Up to you! I like that the Attributes pattern makes extending components super simple and makes adding configuration options to the examples relatively simple. However, I'm not expecting Carousel to be used in so many places that optimizing for these nice-to-haves is necessary.

The config pattern definitely results in easier-to-follow internal code.

---

One thing I'm wondering: how configurable does the component actually need to be? Would it make sense to match the styles from the spec inside the Carousel component, rather than on the about page later?

tesk9 (commented):
It's looking good! The doc comments is the only merge-blocker, although I'm curious what you think about putting styles here rather than where the component will be used.

src/Nri/Ui/Carousel/V1.elm
celsobonutti: I'd prefer to implement the designs in the homepage instead, for two reasons:

* having it on the homepage allows us to iterate quicker on design feedback/changes
* the requirements for an accessible Carousel/Tab can be quite complicated to implement (specially in Elm). I believe having a configurable, yet simple, Carousel component can be of great help when we need one.

src/Nri/Ui/Carousel/V1.elm
```diff
@@ -1,94 +1,44 @@
-module Nri.Ui.Carousel.V1 exposing (..)
+module Nri.Ui.Carousel.V1 exposing (Item, buildItem, view)
```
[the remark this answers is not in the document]
tesk9: Guess I missed it! thank you!

tesk9 (approved):
Love it!

PR #956, opened 2022-06-21
Tooltip table take two

krinhorn:
- Add a hopefully usable heading to the table
- Make toggletip example more authentic so it makes more sense to folks looking at it
- Make clickable text small so it fits in the table better
- Remove redundant and possibly unclear reference to disclosure documentation
- Clarify conditions for using each component wrt interactive elements
    - And drop references to rich text
- Top align text in text-only cells
- Emphasize the primary column w bold styling

## For the reviewer
- Should be reviewable by commit
- [x] Can you please check that I did the vertical alignment for text-only cells properly? I had to add more padding to the leftmost column. I assume this is bc it's using `Table.string`, but the css in the console had me baffled bc I could not find the extra 7px of padding on the other columns. 🤷 
- [x] Make sure the interactive element conditions are correct.
- [ ] Can you check the unapproved percy snapshot? I don't think it has anything to do with what I did but it is weird.

## Outfoxes
Fixes A11-824

### DIFF
--- a/styleguide-app/Examples/Tooltip.elm
+++ b/styleguide-app/Examples/Tooltip.elm
@@ -184,7 +184,7 @@ of ***what*** the tooltip trigger does. The same text is provided to assitive te
           , usage = """
 Use when all of the following are true:
 - the tooltip trigger does more than just reveal the tooltip content
-- the content of the tooltip provides additional information about the functionality of the tooltip trigger itself.
+- the content of the tooltip provides additional information about the functionality of the tooltip trigger itself
 - the content of the tooltip doesn't contain interactive elements such as links
 
 Think of this as the "How."
--- a/styleguide-app/Examples/Tooltip.elm
+++ b/styleguide-app/Examples/Tooltip.elm
@@ -166,6 +166,7 @@ view ellieLinkConfig model =
 Use when all of the following are true:
 - the tooltip trigger does more than just reveal the tooltip content
 - the content of the tooltip is the same as the name of the tooltip trigger
+- the content of the tooltip doesn't contain interactive elements such as links
 
 Think of this as the "What."
 """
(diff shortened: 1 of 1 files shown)


styleguide-app/Examples/Tooltip.elm
```diff
@@ -203,6 +205,8 @@ Examples:
 Use when all of the following are true:
 - the tooltip trigger only opens the tooltip without doing anything else
 - the tooltip trigger ***isn't*** a "?" icon
+
+This type can (but is not required to) contain interactive elements such as links.
```
tesk9, line 209:
```suggestion
This type may contain interactive elements such as links.
```

styleguide-app/Examples/Tooltip.elm
```diff
@@ -220,6 +224,9 @@ This behavior is analogous to disclosure behavior, except that it's presented di
 Use when all of the following are true:
 - the tooltip trigger only opens the tooltip without doing anything else
 - the tooltip trigger ***is*** a "?" icon
+
+This type can (but is not required to) contain interactive elements such as links.
```
tesk9, line 226:
```suggestion
This type may contain interactive elements such as links.
```

tesk9 (approved):
There's something amiss with the Percy snapshots! Haven't had a chance to dig into it yet. But yeah not related to your change.

Looks like 7px of padding is added to the sides for the first 3 columns, to me? Vertical align top looks 👍

styleguide-app/Examples/Tooltip.elm
krinhorn: lol thanks. 🤦

tesk9:
Took a look! I think there's margin on the paragraph tag _inside_ the span:

https://user-images.githubusercontent.com/8811312/175094601-80e4037d-6c64-483b-b28d-d8c1b788ae9b.mov

krinhorn:
Ohhh interesting. So adding padding probably wasn't right in terms of consistency... but lmk if you think it seems like something I should try to fix or I'm just going to leave it.

tesk9:
Nah it's totally fine as-is 👍

PR #963, opened 2022-06-24
Add Menu opensOnHover

bcardiff:
When a `Menu` uses `opensOnHover True` the menu can be opened by mouse enter and closed by mouse leave.

This required some changes in the dom structure to include a menu-hover-bridge to avoid closing the menu when the user is transitioning from the menu trigger to it's content. I also tweaked the zIndex so the overlay is below the trigger button, this was not the case before.

Some tests were added, they introduce a bit of duplication with `tests/Spec/Nri/Ui/Tooltip.elm`.

I notice the Menu does not introduce aria-label components as in the case of the tooltip which would have lead to some slightly nicer test code.

I would have wanted to test that if `opensOnHover False` is passed, the menu will not open on mouse enter, but alas that test will fail since there is no event listener of the event wanted to be simulated.

### DIFF
--- a/tests/Spec/Nri/Ui/Menu.elm
+++ b/tests/Spec/Nri/Ui/Menu.elm
@@ -14,17 +14,22 @@ import Test.Html.Selector as Selector exposing (text)
 spec : Test
 spec =
     describe "Nri.Ui.Menu.V3"
-        [ test "Menu.view" <|
+        [ test "Opens when mouse enters" <|
             \() ->
-                let
-                    menuContent =
-                        "Hello"
-                in
                 program [ Menu.opensOnHover True ]
                     -- Menu opens on mouse enter
                     |> mouseEnter menuButtonSelector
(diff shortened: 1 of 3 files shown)


tests/Spec/Nri/Ui/Menu.elm
```diff
@@ -0,0 +1,124 @@
...
+                    -- Menu opens on mouse enter
+                    |> mouseEnter menuButtonSelector
+                    |> ensureViewHas (menuContentSelector menuContent)
+                    |> mouseLeave menuInteractiveAreaSelector
+                    |> ensureViewHasNot (menuContentSelector menuContent)
+                    |> ProgramTest.done
+        , test "Toggle when mouse clicks" <|
+            \() ->
+                program []
+                    -- Menu opens con mouse click
```
tesk9, line 29:
```suggestion
                    -- Menu opens on mouse click
```

tesk9 (approved):
Looks great!

PR #966, opened 2022-06-27
Bat/tests

tesk9:
Attempts to fix test flakes by updating puppeteer usage.


tesk9:
Didn't actually fix the problem, but at least the dependencies are more up to date 🤷‍♀️

Issue #965, opened 2022-06-27
Add azure border to checked state of checkboxes


bendansby:
This is currently implemented only on the Assignment Library; it should be global. Make sure the global update doesn't conflict with this local AL implementation.

<img width="662" alt="image" src="https://user-images.githubusercontent.com/13528834/175983982-a3bd17c7-b91c-4af0-9488-54d655a9ccc7.png">

tesk9:
I think we may want to change the AL implementation strategy, though? The border doesn't hug the contents the way I want it to:

<img width="116" alt="" src="https://user-images.githubusercontent.com/8811312/176062097-de89cd90-2a9d-4cbc-8ec6-fab3de7dd948.png">

bendansby:
Agreed, we want firm hugs!

PR #967, opened 2022-06-28
Bat/border checkbox

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/965

Adds an azure border for the fully-selected checkbox. 

@NoRedInk/design is this sufficient? I'm pretty sure there's an optical illusion making it look like there's a strip of the light blue right above the shadow?  (Please play with the deploy preview -- I want to be sure the toggling experience is what you'd like too, if you don't mind! 🙏  In particular, LMK if I need to hold this until the focus ring changes.)


bendansby:
Thank you! I added a border to the unchecked state as well to improve contrast (though not fully compliant; we might need an HC mode for that) and to make things match better between states.

tesk9:
Oooo I like it! It makes the difference between enabled and disabled states more clear too.

Should we add the border to the indeterminate and the premium-locked checkbox icons too?

bendansby:
yeeaah 👍

bendansby:
Yup yup!

PR #968, opened 2022-06-28
Darker radio button border

bendansby:
Makes the unchecked radio button border darker to match the updated checkbox and other inputs.

image


tesk9 (approved):
Looks good! A nice subtle improvement 👍

PR #969, opened 2022-06-28
Kraken/add modules icon

ap-nri:
Adds the modules icon!


lindsaykwardell (commented):
Overall seems good, just had one question.

elm.json
```diff
@@ -3,7 +3,7 @@
     "name": "NoRedInk/noredink-ui",
     "summary": "UI Widgets we use at NRI",
     "license": "BSD-3-Clause",
-    "version": "16.2.0",
+    "version": "16.3.0",
```
tesk9, line 6: (16.3.0 is already released now)

lindsaykwardell (approved):

elm.json
lindsaykwardell: I approved, but this should be updated too.

ap-nri: yep, rebased and rebumped

PR #970, opened 2022-06-28
Link icon svg

bendansby:
Updates the link icon to be slightly better looking in the eye of this beholder.

image


tesk9 (approved):
This beholder digs it!

PR #973, opened 2022-07-05
Add Menu.containsForm

bcardiff:
The Nri.Ui.Menu.V3 does not play nice when the content is a form. The user is expected to `<tab>` to change controls, yet menu by default is closed on `<tab>`.

This PR adds a `Menu.containsForm` defaults to `False` (so it's backward compatible) that will turn off the close-on-tab behavior.

https://user-images.githubusercontent.com/459923/177409704-699c8a7c-1656-429e-8bbe-f5f09b1da9de.mp4

### DIFF
--- a/src/Nri/Ui/Menu/V3.elm
+++ b/src/Nri/Ui/Menu/V3.elm
@@ -105,7 +105,7 @@ type alias MenuConfig msg =
     , menuId : String
     , zIndex : Int
     , opensOnHover : Bool
-    , disclosure : Bool
+    , purpose : Purpose
     }
 
 
@@ -115,6 +115,11 @@ type alias ButtonConfig =
     , hasBorder : Bool
     , buttonWidth : Maybe Int
     }
+
+
+type Purpose
+    = NavMenu
+    | Disclosure { lastId : Maybe String }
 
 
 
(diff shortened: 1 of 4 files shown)


src/Nri/Ui/Menu/V3.elm
```diff
@@ -202,11 +208,21 @@ opensOnHover value =
...
+{-| Makes the menu behave as a disclosure.
+
+For more information, please read [Disclosure (Show/Hide) pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/).
+
+You will need to pass in the last focusable element in the disclosed content in order for:
+
+  - any focusable elements in the disclosed content to be keyboard accessible
+  - the disclosure to close appropriately when the user tabs past all of the disclosed content
+
+You may pass a lastId of Nothing if there is NO focusable content within the disclosure.
```
tesk9, line 220:
```suggestion
```

tesk9 (commented):
I checked the deploy preview with VO, and I think we missed something while we were pairing, unfortunately. The pop-up has "role=menu", which shouldn't be there for the `disclosure` pattern:

VO thinks the popup is a menu
... because of the role=menu attribute

tesk9 (commented):
I think this is almost perfect! Just one final thing I noticed from the deploy preview -- it doesn't seem like the popout is closing on escape as I would have expected?

bcardiff:
Good catch, when we added the `WhenFocusLeaves.toAttribute` that causes the `Key.onKeyDown [Key.escape ...]` to no longer apply. The `WhenFocusLeaves.onTab` swallows any keycode that it's not a tab with a `Decode.fail` and it seems we can't emit two `on "keydown"` in the same node.

Should we add a fallbackDecoder to `WhenFocusLeaves.toAttribute` and `.onTab`? Should I create `Nri.Ui.WhenFocusLeaves.V2` for this?

tesk9:
Dang, I don't want to block you more! If you feel like you do have the time though, I think a version of WhenFocusLeaves that accepts a fallback value sounds great!

bcardiff:
No worries we are not in a rush. Would you prefer an overload in V1 or a new V2 with the required argument?

tesk9:
I think V2, if you don't mind, so that we can still do a minor bump 🙏 Thank you!

bcardiff:
@tesk9 I think adding a `toDecoder` to `Nri.Ui.WhenFocusLeaves.V1` offers some composability benefits to `toAttribute`. Let me know if this approach make sense to you, and some rewrites in that module.

tesk9 (approved):
Ahh I love it! Great composability! 👏  Awesome idea, and it makes the Menu code nice to follow too.

PR #976, opened 2022-07-07
Change transition time of menu caret

bendansby:
https://user-images.githubusercontent.com/13528834/177829422-af76292c-b11d-4683-9694-3b5cc367f416.mov


tesk9 (approved):
🐢

PR #983, opened 2022-07-14
Fix div inside an a alert from ARC Toolkit

omnibs:
ARC Toolkit's HTML checker let us know we had a div inside an `<a>`. We swapped it out without any side effects.


tesk9 (approved):
Thanks!!!

PR #986, opened 2022-07-15
Heading V3

bendansby:
Add a new version of Heading that uses Subhead as the default style for h2. To update this in the monolith we'll just need to find the instances of Heading.h2 that have no Heading.style and change them to Nri.Ui.Text, then do a find/replace.

Screen Shot 2022-07-15 at 11 05 38 AM


bendansby:
I presume there is more to do here since this is still showing up as V2 
image

tesk9 (commented):
It will be so nice to stop doing the `Heading.h2 [ Heading.style Heading.SomeThingIAlwaysForgetAndMessUpTheCasingFor ]` dance! Thanks for making this!

Possibly before we release, I'd like to talk through a couple of API/behavior changes for V3 (because of preexisting issues in V2...)

### Errors
So the first concern I have is potentially using just color to convey meaning. There are "error" styles that turn the Heading purple, but don't have another indicator for the error status. It's possible that in-context, it's more clear that there's an error??? But something I want to double-check, at least.

Screenshot of purple error headings

### API
And then the second thing is the API -- I'd like to make it more similar to the other components (like `Text`, for instance).

---

What do you think? Okay to hold off on merging this PR until we can sort out the two things above? (If you're wanting this to merge now, that's fine too -- there's always V4! )

bendansby:
Yeah, that’s fine! No rush.

tesk9:
@bendansby I made the changes! LMK if you have any concerns!

bendansby:
LGTM!

tesk9 (commented):
Okay!! This is done except that it turns out that `Page` is using an h2 (unnecessarily, in my opinion!), so there were some unintentional `Page` style changes.

I will aim to fix it in the coming days (my availability is going to be decreased for the rest of the week, though)

tesk9 (approved):

PR #996, opened 2022-07-22
A11-1117 Add support for more autocomplete values in TextInput

charbelrami:
Adding autocomplete value helps assistive technologies identify the purpose of the input as described in https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html


tesk9 (approved):
Looks great!

If you could just add a description, linking to the issue before merge, that would be great!

PR #998, opened 2022-07-23
Add count

tesk9:
Replaces:
- `Page.Teach.Facilitate.Benchmarks.New.Icon`'s `numeral`
- `Page.Curriculum.Views.LearningModuleIcon`'s `number`

I'm renaming to `count` unless anybody objects because I think it's clearer -- I thought the icon would be a one with a circle in it, when I first saw the names.

cc @NoRedInk/design


bendansby:
Wait, if it's already being used in Modules what's this? https://github.com/NoRedInk/NoRedInk/pull/39483 @erik-nri @ap-nri

tesk9:
It's used in [`viewRootFolderHeader`](https://github.com/NoRedInk/NoRedInk/blob/master/monolith/ui/src/Page/Curriculum/BrowseAndAssignLayout.elm#L1015)... 

(Not sure if this is already on your radar, but in looking at that code, I'm seeing casing on display copy, which is somewhat risky since the copy is highly likely to change over time, and folks who change the copy won't expect the icon and styles to change with it! I recommend a union type instead, or make these values configurable. LMK if you want to pair on it, and I can explain more what I mean)

asterite (approved):
Looks good!

lindsaykwardell:
> It's used in [`viewRootFolderHeader`](https://github.com/NoRedInk/NoRedInk/blob/master/monolith/ui/src/Page/Curriculum/BrowseAndAssignLayout.elm#L1015)...
> 
> (Not sure if this is already on your radar, but in looking at that code, I'm seeing casing on display copy, which is somewhat risky since the copy is highly likely to change over time, and folks who change the copy won't expect the icon and styles to change with it! I recommend a union type instead, or make these values configurable. LMK if you want to pair on it, and I can explain more what I mean)

Completely agree on the display copy, however this is what was specced and requested. Once modules gets some time in production we'll probably revisit to improve this code.

erik-nri:
@bendansby when I paired with alex on this we found the numeral use over in benchmarks and did not realize it was also implemented in LearningModules.

PR #1003, opened 2022-07-26
Tessa/fix open book

tesk9:
Adjust viewport for openBook icon.

# Before

Screen Shot of off-center icon

# After

Screen Shot of on-center icon

cc @NoRedInk/design


bendansby:
🥰

PR #1005, opened 2022-07-27
Fix Modal focus

tesk9:
When focus leaves the final focusable element in the modal, the first element in the modal should be re-focused. This behavior was broken due to browser defaults -- not sure when it broken exactly, but 🤷‍♀️ .

@juanedi and @mandla-noredink -- tagging you as reviewers since you reported the issue! 🙏 

Edit: and @adauguet !


mandla-noredink (approved):
image

PR #1007, opened 2022-07-28
New sort arrows

bendansby:
[Safari doesn't respect](https://stackoverflow.com/questions/48248512/svg-transform-rotate180-does-not-work-in-safari-11) `transform="rotate(180)` on SVGs, so this replaces the rotated `sortArrow` with a normalized downward pointing triangle. Of course another solution to this would just be to have a single `sortArrow` and rotate its instances with CSS. This PR also makes the triangles generally cuter.

## Before
image
image

## After
image
image


tesk9 (approved):
Oooo they ARE cuter!!! 😍 😍 😍 

I didn't realize that transforms on svgs wouldn't work for safari.. Hopefully I didn't do the same thing in other places 🤔

PR #1014, opened 2022-08-04
A11-1247 add `role="switch"` to Switch checkbox

charbelrami:
Makes the SR interpret the Switch checkbox input as a switch (on/off as opposed to checked/unchecked)

Screen Shot 2022-08-04 at 17 37 54

### DIFF
--- a/src/Nri/Ui/Switch/V2.elm
+++ b/src/Nri/Ui/Switch/V2.elm
@@ -19,6 +19,7 @@ module Nri.Ui.Switch.V2 exposing
     - REQUIRE label and id always
     - Move custom attributes to the container
     - change disabled to take a bool (which I think is the slighty more common pattern)
+    - Adds `role="switch"`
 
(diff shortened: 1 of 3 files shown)


src/Nri/Ui/Switch/V2.elm
```diff
@@ -214,6 +215,7 @@ viewCheckbox config =
     Html.checkbox config.id
         (Just config.selected)
         [ Attributes.id config.id
+        , Attributes.attribute "role" "switch"
```
tesk9, line 218: 🎨 Could we use [Accessibility.Styled.Role's `switch`](https://package.elm-lang.org/packages/tesk9/accessible-html/latest/Accessibility-Role#switch) helper instead?

tesk9 (approved):
Looks great! One small suggestion, but you can merge without applying it if you'd prefer

PR #1015, opened 2022-08-04
A11-1241 remove `aria-hidden` from Menu

charbelrami:
removes `aria-hidden` from the menu because we already have `display: none`


tesk9 (approved):
Fabulous!

@NoRedInkBot r+

PR #1016, opened 2022-08-04
Increase z-index of modals

bendansby:
I feel like there must be some backstory to why modals of all things have a z-index of 1, so LMK if this is the wrong move.

This is prompted by https://noredink.slack.com/archives/C02JPNAHEPN/p1659648695838389?thread_ts=1659622356.924599&cid=C02JPNAHEPN


tesk9 (approved):
This makes sense to me! I'm not sure there was a real _reason_ for the low z-index.

PR #1017, opened 2022-08-05
A11-1234 add `type="button"` to tabs

charbelrami:
adds `type="button"` to prevent form submissions

### DIFF
--- a/src/TabsInternal/V2.elm
+++ b/src/TabsInternal/V2.elm
@@ -119,7 +119,8 @@ viewTab_ config index tab =
                 Nothing ->
                     -- This is for a non-SPA view
                     ( Html.button
-                    , [ Events.onClick (config.focusAndSelect { select = tab.id, focus = Nothing })
+                    , [ Attributes.type_ "button"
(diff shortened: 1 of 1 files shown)


src/TabsInternal/V2.elm
```diff
@@ -136,6 +136,7 @@ viewTab_ config index tab =
                          Attributes.disabled (not isSelected && tab.disabled)
                        , Aria.selected isSelected
                        , Role.tab
+                       , Attributes.type_ "button"
```
tesk9, line 139: Looking at the name of the function `buttonOrLink`, I was reminded that sometimes the tab is an anchor element rather than a button.

[Elllie example showing that the tab can be an anchor](https://ellie-app.com/jftVLNmTL9ca1)
Screen Shot of example ellie


You can see this behavior in use in a couple of places in the app, but probably the easiest one to get to is Mock Data, in the assignment-type tabs.


Perhaps we should only attach the button type when the tag is a button?

tesk9 (approved):
Thank you! Looks great!

PR #1018, opened 2022-08-05
Add menu alignment offset

micahhahn:
We want our new kebab menu popout to look like below.  This PR adds a new function attribute generator called `alignmentWithOffset`.  This should not requiring bumping the menu version.

image

image


tesk9 (commented):
I think the reason the Menu doesn't look aligned is because the ClickableSvg button trigger is being shifted right, without also shifting the containing element that the Menu popout is positioned against.

How would you feel about shifting the container instead? I made a PR showing what I mean over here: https://github.com/NoRedInk/NoRedInk/pull/39786

PR #1021, opened 2022-08-08
Adds disabled and loading attributes/config

erik-nri:
This implements similar attributes to how it is being done over in
TextInput.

### Motivation
For hackday I was updating the sign up with email form and adding the disabled / loading attributes for the TextInputs. I noticed that the Select did not have similar functionality and thought it might be worth adding 😄.

One odd thing to note, when adding the fields to the examples app, I noticed that I needed to flip the order in the list of the select input and its label. If I did not the styles did not work out for some reason. I was pairing with @charbelrami at the time and we decided to hold off on spending a bunch of time trying to figure out why this was happening.

This update only incurs a minor change!

### DIFF
--- a/src/Nri/Ui/Select/V8.elm
+++ b/src/Nri/Ui/Select/V8.elm
@@ -5,7 +5,7 @@ module Nri.Ui.Select.V8 exposing
     , value
     , Attribute, defaultDisplayText
     , hiddenLabel, visibleLabel
-    , errorIf, errorMessage, guidance
+    , disabled, loading, errorIf, errorMessage, guidance
     , custom, nriDescription, id, testId
     , containerCss, noMargin
     )
@@ -37,7 +37,7 @@ module Nri.Ui.Select.V8 exposing
 
 @docs Attribute, defaultDisplayText
 @docs hiddenLabel, visibleLabel
-@docs errorIf, errorMessage, guidance
+@docs disabled, loading, errorIf, errorMessage, guidance
 @docs custom, nriDescription, id, testId
 @docs containerCss, noMargin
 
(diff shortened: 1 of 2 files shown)


src/Nri/Ui/Select/V8.elm
```diff
@@ -286,7 +310,14 @@ view label attributes =
             , valueToString = config.valueToString
             , defaultDisplayText = config.defaultDisplayText
             , isInError = isInError_
+            , disabled = disabled_
             }
+        , InputLabelInternal.view
```
erik-nri, line 315: This is the funky bit. The only time the styles look off is when you check disabled or loading in the example app. It makes is look like the label is behind the border for a non-obvious to me reason (also im really rusty at css so maybe it is actually something simple...).

Flipping the order had no effect on the "normal" state of things as far as I can tell.

tesk9 (commented):
Looks good! Thanks for adding to the example, especially! 🙏 

I have a couple of small requests before this merges:

1. Could you please run this by the design team, and make sure this is how they'd like to display loading and disabled Selects?

2. Could you please make sure the cursor doesn't make the input look interactive?

cursor

erik-nri:
> 1. Could you please run this by the design team, and make sure this is how they'd like to display loading and disabled Selects?
> 2. Could you please make sure the cursor doesn't make the input look interactive?

I will tackle number 2 and then follow up with Ben afterwards to make sure all is well, thanks @tesk9!

erik-nri:
https://user-images.githubusercontent.com/109546847/183560986-1fae7732-b6b3-40d1-8a02-8a796cff351b.mov

Updated! Will check in with design in the morning.

erik-nri:
Hello @bendansby! When able please take a look at the video above for the loading/disabled states for the Select input. It is currently implemented to be the same as TextInput. If it looks good to you I think we can merge it!

bendansby:
[I was thinking just the other day](https://noredink.slack.com/archives/C02NMHB1K55/p1659569662221389) that it'd be good for our input disabled state to be less hard to see. However, I think this translucent look for inputs and selects _does_ work for the "disabled because loading" state. So maybe ideally we would have 2 different kinds of disabled state, but that can be for another day.

bendansby (approved):

erik-nri:
Awesome thanks @bendansby! I think this is all set then @tesk9. Percy is piping up but its just for the new fields on the examples page 😄.

tesk9 (approved):

erik-nri:
@NoRedInkBot r+

PR #1022, opened 2022-08-09
A11-1227 add optional `role="region"` to accordion panels

charbelrami:
## Motivation

> Role region is especially helpful to the perception of structure by screen reader users when panels contain heading elements or a nested accordion.

It's optional because, on some pages, we have multiple panels from nested accordions expanded at the same time which may cause noise to screen readers. It can also be optional because panel is a `section`, a semantic element that doesn't need a `role` when it has `aria-labelledby`.

> Avoid using the region role in circumstances that create landmark region proliferation, e.g., in an accordion that contains more than approximately 6 panels that can be expanded at the same time.

## Reference
https://w3c.github.io/aria-practices/#wai-aria-roles-states-and-properties

### DIFF
--- a/deprecated-modules.csv
+++ b/deprecated-modules.csv
@@ -1,4 +1,5 @@
-Nri.Ui.Accordion.V1,upgrade to V3
+Nri.Ui.Accordion.V1,upgrade to V4
+Nri.Ui.Accordion.V3,upgrade to V4
 Nri.Ui.Checkbox.V5,upgrade to V6
 Nri.Ui.Heading.V2,upgrade to V3
 Nri.Ui.Menu.V1,upgrade to V3
--- a/elm.json
+++ b/elm.json
@@ -8,6 +8,7 @@
         "Nri.Ui",
         "Nri.Ui.Accordion.V1",
         "Nri.Ui.Accordion.V3",
+        "Nri.Ui.Accordion.V4",
         "Nri.Ui.AssetPath",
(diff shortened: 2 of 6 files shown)


tesk9 (commented):
To clarify my understanding, do you have an example in mind of when we would want the role to be region? I guess I want to be sure before we do an upgrade of all the accordion uses (I think there are 24 instances of Accordion.view across the NoRedInk codebase) that we will have a mix of `panelAsRegion = True` and `panelAsRegion = False`.

The other thing I'm wondering is whether the setting should be per-accordion-entry, rather than per-accordion-tree. 🤔 Would you ever want some accordions to be regions, and others not?

Finally -- if we're adding a new Accordion version, I'd love for it to address some of the other Accordion issues before we release, like the unnecessary aria-live attribute.

tesk9:
Thank you!

I don't think there's a particular reason that the Accordions component isn't used on the Assignments page -- probably the component didn't exist when the page was initially built.

> But if some specific cases are not worth the impact of updating all of the accordions, we can leave out this change.

I think the upgrade will be relatively simple. We should be able to match the existing behavior exactly with a global find and replace. The hardest bit will be going through and deciding which behavior we want for each Accordion use. Is that something that you are interested in doing?

tesk9 (approved):

tesk9:
Awesome, thank you for investigating! 🙏

PR #1024, opened 2022-08-09
Resolves A11-1254, A11-1235, and A11-1277: remove roving tabindex and `aria-hidden` from tab panels, add a fixed `tabindex="0"`

charbelrami:
Roving tabindex is not necessary on the panels. Panels with `display: none` are not focusable even with `tabindex="0"`. Added a fixed `tabindex="0"` to all panels to ensure they are focusable even when there are no focusable descendants.


tesk9 (approved):
Ahh awesome, thank you!

PR #1027, opened 2022-08-11
Tighten and standardize heading line heights

bendansby:
Want to get away from using custom pixel values for line height because it's easier to make consistent across time and space using proportional values. Also headings (1.2) should be a little tighter than body copy (1.5) I think. Will probably change my mind about those values at some point in the future and overrides are always a possibility.

## Before and After
image


tesk9 (approved):
I dig it! Thank you!

PR #1030, opened 2022-08-16
A11-1357 add data-selected to tabpanel for testing purposes

charbelrami:
This replaces `aria-hidden="false"`, which was used as a selector in test queries.


tesk9 (approved):
Looks good!

PR #1037, opened 2022-08-22
Bat/improve code infra

tesk9:
While experimenting with elm-codegen, I started reorganizing the code formatting helpers a bit. Hopefully, these changes will make the generated code tidier, at least until we can switch to elm-codegen.


omnibs (approved):
Changes look good! Cool stuff, I didn't know elm-codegen!

PR #1038, opened 2022-08-22
Adds TextInput example code

tesk9:
Adds example code for TextInput. All the Ellie links should be fully functional except for the newPassword one.

Screen Shot of the TextInput.number example code

Fixes A11-1475


tesk9:
@NoRedInkBot r+

PR #1039, opened 2022-08-23
Migrating the few icons leftover in modules

erik-nri:
I took a best guess stab at where to categorize these with the rest of
the examples.


erik-nri:
I also found that the AssignmentIcon.modules is almost exactly the same as the moduleCircle from LearningModuleIcons.
Screen Shot 2022-08-23 at 2 11 38 PM

In fact that screenshot makes them look exactly the same, however I swear there is actually a slight difference...

tesk9 (approved):
Thank you! Looks great to me!

erik-nri:
@NoRedInkBot r+

PR #1040, opened 2022-08-23
Add new design logos for Clever and Google

mfonism:
This PR adds the new design logos for Clever and Google to NRI UI.

Screenshot 2022-08-23 at 21 07 35

Two of them are currently named `cleverFull` and `googleClassroomFull` because the names I think they should actually take (`clever` and `googleClassroom`, respectively) are already used.

I made the decision to call add the `-Full` suffix for now so we don't have to track down where the older logos are used in `NoRedInk/monolith` to update them. The idea is that we can coordinate efforts to rename the logos to be consistent, and update their use in `NoRedInk/monolith` at a later time.

What do you think about this? Is there another way we could do this?

Resolves [ZEN-599](https://linear.app/noredink/issue/ZEN-599/[preferences]-add-new-design-icons-for-google-and-clever-to-noredink)


tesk9 (commented):
I re-organized the logos so that hopefully it's a little easier to find the one you want:
image

I'm getting a little bit confused about the purposes of all the logos.

Are we still trying to distinguish between Clever Library and Clever Sync in the app? (Is there a difference?) Why are some of the icons a navy-ish color? The [media kit](https://clever.com/about/media-kit) seems to suggest that the logo should be `#4274f6`, unless I'm misunderstanding?

> I made the decision to call add the -Full suffix for now so we don't have to track down where the older logos are used in NoRedInk/monolith to update them. The idea is that we can coordinate efforts to rename the logos to be consistent, and update their use in NoRedInk/monolith at a later time.

This is fine with me, although I think there are probably only ~2 uses of each written-out logo (teacher dashboard and manage classes).

tesk9 (approved):

mfonism:
Updates to the Logos:

* Clever's primary icons have been updated and renamed `cleverBlue` and `cleverWhite`
* Clever's secure sync icon (the full variant) has been renamed `cleverSecureSync`

mfonism:
Updates to the Logos:

* Clever's primary icons have been updated and renamed `cleverBlue` and `cleverWhite`
* Clever's secure sync icon (the full variant) has been renamed `cleverSecureSync`

Screenshot 2022-08-25 at 12 27 07

tesk9 (commented):
I pushed a commit that shares some code between the 2 clever text icons.

So one thing to be aware of is that if you remove icons, that will cause the next version bump to be a major bump. Generally speaking, we avoid major bumps when we can.

What we could do instead is more of a soft removal:
- remove the undesired icons from the styleguide
- add a deprecation notice to the docs for each icon
- add an issue in noredink-ui with a "hold-for-next-major-release" label attached so that somebody remembers to remove them when we do the next major release

tesk9:
I'll be OOO until Tuesday, but feel free to move this PR forward/merge without any additional input from me. 😄  
Mandla is also adding some icons -- you may want to coordinate the next release with him.

PR #1045, opened 2022-08-25
Added print icon to uiIcons

mandla-noredink:
Icon added for portfolio items printing view buttons


tesk9 (approved):
Wow I'm kind of astonished we didn't have this already!

It looks great!

PR #1055, opened 2022-08-30
A11-1340 add `"country-name"` autocomplete to TextInput

charbelrami:
Adds `"country-name"` autocomplete value to be used on `/teach/tour > outside the US`, part of [A11-1117](https://linear.app/noredink/issue/A11-1117).


tesk9 (approved):
Awesome, thank you!

PR #1056, opened 2022-08-30
Fix Tabs focus ring behavior

tesk9:
Fixes A11-1519

Only show the focus ring when focus rings should be visible:

https://user-images.githubusercontent.com/8811312/187511396-2ce35af5-941d-4513-b842-d443a1ec38d8.mov


BrianHicks (approved):

BrianHicks:
@NoRedInkBot r+

PR #1057, opened 2022-08-30
Adds reasonable defaults for accordion focus rings

tesk9:
Fixes A11-1521

Having the focus ring styles inset for the accordion will prevent fancy customized accordions from having content cut off. It should also look nicer for places like the mobile navbar.

### Before

image

Screen Shot 2022-08-30 at 4 53 08 PM

### After
image

Screen Shot 2022-08-30 at 4 53 19 PM

cc @NoRedInk/design


mfonism (approved):
Fantastic. I like!

accordion

PR #1059, opened 2022-08-31
Bat/segmented control focus ring

tesk9:
I am actually okay with the Before, and would be fine with leaving it as-is -- I like that the box shadows indicating selection status are nice and clear.

If we want to 100% not have overlap/hidden content, then we can use the inset styles instead. I think it's maybe harder to tell which part of the switch is selected, now -- focus is clear, but the non-color indicator of selection is less clear. Your call, @NoRedInk/design ! Or if you have other ideas, they're also welcome too (especially if they're easy 😛)

Fixes https://linear.app/noredink/issue/A11-1539/focus-ring-overlap-on-segmented-controls

## Before
Screen Shot 2022-08-31 at 3 58 58 PM

## After
image

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -3,7 +3,7 @@
     "name": "NoRedInk/noredink-ui",
     "summary": "UI Widgets we use at NRI",
     "license": "BSD-3-Clause",
-    "version": "18.2.1",
+    "version": "18.3.0",
     "exposed-modules": [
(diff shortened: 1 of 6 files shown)


bendansby:
Agreed that Before is preferable in some ways so we don't need to change.

PR #1061, opened 2022-09-01
Tessa/add animated icons

tesk9:
Extract hamburger-x-animated-icon to noredink-ui, re-done as an Svg so that we can use ClickableSvg (which in turn will resolve fixes A11-1495)

https://user-images.githubusercontent.com/8811312/188029681-62421f4a-e116-4e15-a0c1-dc622caf79c7.mov

cc @NoRedInk/design the animation is slightly different. I was trying to avoid the viewBox clipping the edges of the svg. Is this okay with you?

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -7,6 +7,7 @@
     "exposed-modules": [
         "Nri.Ui",
         "Nri.Ui.Accordion.V3",
+        "Nri.Ui.AnimatedIcon.V1",
         "Nri.Ui.AssetPath",
         "Nri.Ui.AssignmentIcon.V2",
         "Nri.Ui.Balloon.V1",
--- a/src/Nri/Ui/AnimatedIcon/V1.elm
+++ b/src/Nri/Ui/AnimatedIcon/V1.elm
@@ -2,7 +2,6 @@ module Nri.Ui.AnimatedIcon.V1 exposing (mobileOpenClose)
 
 import Css
 import Css.Animations
-import Nri.Ui.Colors.V1 as Colors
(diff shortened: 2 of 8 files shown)


bendansby:
Looks nice, but is it less wide than before?

tesk9:
I don't think it's narrower than it was before. I did zoom-out the viewbox a little bit to make room in the animation at the halfway point so the edges wouldn't get cut off, so the SVG is less trim-to-the-edges than they usually are. 🤔 I added 2px of whitespace to each side. 

The current version is 22x17, although the X does look like it has extra whitespace on the sides:

Screen Shot 2022-09-01 at 5 50 19 PM
Screen Shot 2022-09-01 at 5 50 03 PM

Accounting for 2px of whitespace on each side, the icon at 26px width should look pretty similar to the screenshots above.  Looks like the rects end up at 22.4x4.5.

image

Comparing that back again to the published version:
image
Looks like 22x3...

SO! I think the rects are fatter, more than anything else. Although they do _look_ narrower. I will narrow them up, and see if that resolves the discrepancies.

Is it cool if I remove UiIcon.hamburger from the styleguide and mark it as deprecated, so I don't need to update that one to match?

tesk9:
Better?

https://user-images.githubusercontent.com/8811312/188033087-4264f864-df7c-4341-808d-420628cb709b.mov

tesk9:
Dang -- just discovered that the transforms don't transition well in safari. I'll address it before merging this.

bendansby:
Looks great, thanks for adjusting that!

PR #1062, opened 2022-09-02
A11-1543 Extends the hover area of tooltips

charbelrami:
Since the tooltip appears when hovering over the trigger and disappears on mouse out, we want to extend the hover area to make it easier to get from the trigger to the tooltip.

[See more on motivation here.](https://noredink.slack.com/archives/C02NVG4M45U/p1662038535083739)

Visible extra hover area below the tooltip
Visible extra hover area to the left of the tooltip

The background color is only visible in the screenshots.


tesk9 (approved):
Awesome, looks great! This will definitely be better for users 🙏

PR #1063, opened 2022-09-02
Hack day project: adding Layout

tesk9:
`Nri.Ui.Layout.V1` will replace `Nri.Layout` from the monolith. It codifies the experimental spacings from the [UI Style Guide and Caveats doc](https://paper.dropbox.com/doc/UI-Style-Guide-and-Caveats--BobQllelpdS56NBITiRcrO6gAg-PvOLxeX3oyujYEzdJx5pu#:uid=905917270049954035442315&h2=:under-construction:-Spacing).

screenshot of example showing layout helpers in action

cc @NoRedInk/design


bendansby:
ahhh cool! love to be able to play around with it. what do the children represent?

tesk9:
Yay!
The children were meant to represent whatever content might be a "component" in:

> - Space between horizontally laid out components: 10px
> - Space between vertically laid out components: 20px

Very happy to change the example to be more specific or to make it more clear? If you have suggestions for me?

bendansby:
I think this layout component may not be the right place to specify spacing between certain kinds of elements. Perhaps that could be a separate helper, maybe first deployed on a repertoire page? Or in other words, maybe "layout" is too broad a name and this should be more of a "content container" component. Just my initial reaction.

tesk9:
What do you think about renaming to `Nri.Ui.Spacing.V1`?

bendansby:
That could be better, but I'm still thinking "consistent main container for content" is a different thing than "spacing between elements."

tesk9:
Hmm, I guess I'm thinking less "consistent main container for content" and more "spacing around centered elements, of which there is often just one"? If that makes sense?

bendansby:
Yeah, I definitely see where you're coming from! I'm happy to just view this as "v v helpful but somewhat not what I would personally do" if I am a blocker 😀

bendansby:
Also the grid seems ironically messed up in this branch? 
image

tesk9:
It is **TREMENDOUSLY** typical that this would be the PR that I would mess up the layout in 😆 thanks for catching!

bcardiff:
If we want to reserve the Layout namespace for something more I think Spacing or Formatting would make sense for the scope of the features introduced here. The direction looks good to me. As discussed with tessa I'm interested in some higher abstraction to define layout components as building blocks.

bcardiff (approved):

PR #1064, opened 2022-09-02
Extract RingGauge component

tesk9:
Screen Shot of the example

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/RingGauge/V1.elm
+++ b/src/Nri/Ui/RingGauge/V1.elm
@@ -1,0 +1,104 @@
+module Nri.Ui.RingGauge.V1 exposing (view)
+
+{-| A ring gauge made with svg that shows completion by stroking the circumference
+with a specific color.
+
+Uses the strategy discussed here:
+<https://medium.com/@heyoka/scratch-made-svg-donut-pie-charts-in-html5-2c587e935d72>
+
+@docs view
+
+-}
+
+import Css exposing (Color)
+import Nri.Ui.Colors.Extra exposing (toCssString)
+import Nri.Ui.Svg.V1 as Svg exposing (Svg)
+import Svg.Styled exposing (circle)
+import Svg.Styled.Attributes exposing (..)
+
+
+{-| Shows a configurable SVG ring guage
+-}
+view :
+    { backgroundColor : Color
+    , emptyColor : Color
+    , filledColor : Color
+    , percentage : Float
+    }
+    -> Svg
+view config =
+    let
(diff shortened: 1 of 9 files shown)


bendansby:
Quite atomic!

src/Nri/Ui/RingGauge/V1.elm
```diff
@@ -17,15 +17,25 @@ import Svg.Styled exposing (circle)
...
 
 
-{-| Shows a configurable SVG ring guage
+{-|
+
+  - background color : the center of the ring's color
+  - empty color: representation of the not-yet-done/unfilled part of the ring
+  - filled color: representation of the done part of the ring
+  - percentage: 0-100 value representing completion
+
```
omnibs, line 26: 💜

styleguide-app/Examples/RingGauge.elm
```diff
@@ -78,10 +81,49 @@ example =
                 |> Svg.withWidth (Css.px 200)
                 |> Svg.withHeight (Css.px 200)
                 |> Svg.toHtml
+            , Table.view
```
omnibs, line 104: When I saw this table I googled "how much contrast between colors accessibility" and found 4.5:1 as an answer.

🎨 Hint somewhere about this value, or mark the non-self-to-self comparisons below 4.5 some way to make it clear they don't have enough contrast

styleguide-app/Examples/RingGauge.elm
```diff
@@ -78,10 +81,49 @@ example =
...
+                [ Table.string
+                    { header = "Color contrast against"
+                    , value = .name
+                    , width = Css.px 50
+                    , cellStyles = always []
+                    , sort = Nothing
+                    }
+                , Table.string
+                    { header = "backgroundColor"
+                    , value = .value >> contrast settings.backgroundColor >> String.fromFloat
```
omnibs, line 114: 🎨 Use `myrho/elm-round` to shrink floats' string representation

omnibs (approved):
Looks good! I left a couple of 🎨 suggestions, feel free to take 'em or leave 'em!

styleguide-app/Examples/RingGauge.elm
tesk9: I double-checked [the success criterion guidance](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html), and I _think_ that these items only need 3:1 contrast, but I'm not 100% sure, which is why I didn't put the guidance. The rules around color contrast are complicated, since they depend on context!

PR #1065, opened 2022-09-06
Bat/add interactive highlighter

tesk9:
Fixes A11-1551

Copies in non-JS-dependent portions of Highlighter, and add a new example section.

https://user-images.githubusercontent.com/8811312/188735904-d9b16037-a2e6-4049-b709-c6c864c408f0.mov

cc @NoRedInk/design 

I'm not entirely sure why the styles look so bad at the moment, but I think it's something I want to hold off on trying to fix for noredink-ui for now. The goal is to work with the highlighter in isolation to get the keyboard and screenreader experiences nice -- we're not planning to make any style changes. So however the styles currently work (sass?? button style resets somewhere??) on the monolith, should continue to work despite this being extracted..


bendansby:
Understood, but curious which styles you're referring to? Also, which/what % of our highlighting experiences would this replace?

tesk9:
It's maybe hard to see in the video, but there's serious shenaniganry with the height of the elements.

### The not-independently-highlightable spaces are too tall:
screenshot of problem

---

### And there's unexpected overlaps:
screenshot of problem

bendansby:
Ah gotcha! Since this is explicitly for quiz engine highlighting, might it make sense to match that styling for the examples (big quizFont)? I know not the focus but just to reduce confusion/match the "feel"

krinhorn:
> Since this is explicitly for quiz engine highlighting

It's just for QE _for now_.

> might it make sense to match that styling for the examples (big quizFont)?

I think so, eventually, or make that a setting on the page if it's possible to do so. e.g., "Font style: Quiz Engine, Self Review, etc."

tesk9:
I think eventually, yes! all the styles you might want should be available as part of the Highlighter. However, I think the current pattern in noredink-ui is to not add styles just for the example if avoidable, since folks might expect styles to be added that aren't.
Does that make sense?

PR #1066, opened 2022-09-07
Bat/highlighter aria roles

tesk9:


tesk9:
I think that the CI failure is due to a bug in the axe-core ruleset. I made https://github.com/dequelabs/axe-core/issues/3649 over there, but for now I'll just skip the rule.

PR #1072, opened 2022-09-13
A11-1550 Highlighter keyboard nav and focus management

charbelrami:
Implements highlighter keyboard nav and focus management in accordance with the following spec.

Key | Function
-- | --
`ArrowRight` | Moves focus and selection one element to the right. If the last (LTR) element has focus, does nothing.
`ArrowLeft` | Moves focus and selection one element to the left. If the first (LTR) element has focus, does nothing.
`Space` | Applies highlight to the selected elements.
`Shift` + `ArrowRight` | Extends selection one element to the right. Moves focus one element to the right. If the last (LTR) element has focus, does nothing.
`Shift` + `ArrowLeft` | Extends selection one element to the left. Moves focus one element to the left. If the first (LTR) element has focus, does nothing.

An element can be word, space or sentence.

Manages focus using [Roving tabindex](https://w3c.github.io/aria-practices/#kbd_roving_tabindex)

### DIFF
--- a/src/Nri/Ui/Highlighter/V1.elm
+++ b/src/Nri/Ui/Highlighter/V1.elm
@@ -553,10 +553,6 @@ groupContainer viewSegment highlightables =
                         [ markedWith.name
                             |> Maybe.map (\name -> Aria.roleDescription (name ++ " highlight"))
                             |> Maybe.withDefault AttributesExtra.none
-
-                        -- Temporarily adding tabindex 0 so that the mark element can be focused,
-                        --so we will be able to tell how it will read
-                        --   tabindex 0
                         , css
                             [ Css.Global.children
                                 [ Css.Global.selector ":first-child"


src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -451,9 +553,10 @@ groupContainer viewSegment highlightables =
                         [ markedWith.name
                             |> Maybe.map (\name -> Aria.roleDescription (name ++ " highlight"))
                             |> Maybe.withDefault AttributesExtra.none
-                        , -- Temporarily adding tabindex 0 so that the mark element can be focused,
-                          --so we will be able to tell how it will read
-                          tabindex 0
+
+                        -- Temporarily adding tabindex 0 so that the mark element can be focused,
```
tesk9, line 557: Should we delete the comment and removed `tabindex 0` line?

tesk9 (approved):
This is great! SO excited about it!

PR #1073, opened 2022-09-14
Adjust SideNav styles

bendansby:
Add some styles that commonly have to be set per instance, like widths and margin, and adjust appearance of close/open button.

image

image


tesk9 (approved):
Love it, thank you!

PR #1074, opened 2022-09-14
Remove loading dots

bendansby:
V2 of Loading cuz I don't like the dots.

image

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -38,7 +38,7 @@
         "Nri.Ui.Html.Attributes.V2",
         "Nri.Ui.Html.V3",
         "Nri.Ui.InputStyles.V3",
-        "Nri.Ui.Loading.V1",
+        "Nri.Ui.Loading.V2",
         "Nri.Ui.Logo.V1",
         "Nri.Ui.MasteryIcon.V1",
         "Nri.Ui.MediaQuery.V1",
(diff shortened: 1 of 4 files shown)


bendansby:
Could we not just swap them easily with the pencil?

tesk9:
We can swap them, I suppose, but we will need to go through one by one to make sure they look okay.

For instance, a simple swap for writing portfolios goes from:

https://user-images.githubusercontent.com/8811312/190827615-f7733413-2cda-4950-8e73-6ef78c1d47b1.mov

to:

https://user-images.githubusercontent.com/8811312/190827617-85c7fcf8-1195-4009-ba7e-58fd743aa252.mov

Where the pencil gets cut off at the top...

Ideally, would the spinning pencil actually be what we use here? Rather than the skeleton-y approach that mimics the shape of what will appear?

bendansby:
Skeleton is a bit nicer I think, but only if it's customized to mimic the shape of the content; sometimes we use a boilerplate skeleton that looks awkward. Anyway, I will close this PR and maybe see about going the other way and replacing the dots with pencils first 👍

bendansby:
Oh I guess you're already doing that!

tesk9:
😆 Yeah, I'll replace 'em!

bendansby:
Better than room service!

PR #1077, opened 2022-09-15
Bat/breadcrumbs v2

tesk9:
Fixes A11-1592

Integrates secondary breadcrumbs into a new version of the breadcrumbs component.

Screen Shot 2022-09-16 at 4 07 54 PM

cc @NoRedInk/design

### DIFF
--- a/deprecated-modules.csv
+++ b/deprecated-modules.csv
@@ -1,2 +1,3 @@
+Nri.Ui.BreadCrumbs.V1,upgrade to V2
 Nri.Ui.Checkbox.V5,upgrade to V6
 Nri.Ui.Tabs.V6,upgrade to V7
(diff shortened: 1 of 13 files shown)


bendansby:
Nice! In your screenshot, the final crumb is Navy (as desired), but in the deploy preview, it is azure? 
image

tesk9:
Whether it's azure or navy depends on whether the final secondary crumb matches the current route or not (because we're asking, hey, should this be a link, or not?). So if you edit the currentRoute field to `/breadcrumb-subcategory-2`, and have 2 subcategories up there, it will render as a non-link heading.

image

tesk9:
I could make it always navy, regardless of whether it goes someplace, if that's your preference!

bendansby:
Nope, current behavior is fine. Thanks for clarifying!

zwilias (approved):
breadcrumbs

PR #1084, opened 2022-09-16
Additional tooltip directions & alignments

tesk9:
Adds additional breakpoint helpers for tooltips, so that we can customize positioning and styles specifically for the narrow mobile viewport (< 500px).

For the first item in https://github.com/NoRedInk/NoRedInk/pull/40370#issuecomment-1248761408


mfonism (approved):
Fantastic!

direction--compass-00

PR #1087, opened 2022-09-16
Respect prefers reduced motion

tesk9:
Adjusts spinningPencil and spinningDots and adds spinning in order to reduce motion when requested, while still providing a visual loading indicator.

<details><summary>Previous version of the description, before discussion on PR</summary>

The spinningPencil is a LOT when it just keeps rendered. This PR makes it respect prefers reduced motion settings. ... Which makes it much _less_ clear that it's meant to be a loading state. 

image

cc @NoRedInk/design 

</details>

### DIFF
--- a/src/Nri/Ui/Loading/V1.elm
+++ b/src/Nri/Ui/Loading/V1.elm
@@ -68,12 +68,22 @@ loading_ withCss =
 {-| -}
 spinningPencil : Nri.Ui.Svg.V1.Svg
 spinningPencil =
+    let
+        diagonalLength =
+            ceiling (sqrt (2 * 100 * 100))
+
+        paddingForAnimation =
+            (toFloat diagonalLength - 100) / 2
+    in
     UiIcon.edit
         |> Nri.Ui.Svg.V1.withLabel "Loading..."
         |> Nri.Ui.Svg.V1.withColor Colors.navy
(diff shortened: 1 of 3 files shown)


bendansby:
Would it be crazy to use dots instead of pencil when reduced motion is on?

tesk9:
Interesting idea!

I think that would work if the dots didn't translate, but instead had their color transform in-place, you know?


https://user-images.githubusercontent.com/8811312/190831348-24eeddfd-4285-48ef-94db-831226dd1a6d.mov


Personally, I find the pencil & the dots a bit dizzying (hence the button to show them). Not having actual motion but just _implied_ motion would probably help!

I'll push up the code to use color transitions instead of rotations, so you can see it on this branch in the preview!

tesk9:
https://user-images.githubusercontent.com/8811312/193638872-516fb396-3ded-471c-b4d3-00adb4334168.mov

@bendansby what do you think?

bendansby:
Very cool! Is it possible for the two to share the same bounding box?

tesk9:
Like so?

https://user-images.githubusercontent.com/8811312/193661040-fe3c8de1-5a70-48f9-859f-1b90f6e604e7.mov

bendansby:
Nailed it, thank you!

PR #1088, opened 2022-09-19
A11-1616 make arrow keys skip static elements

charbelrami:
This makes left/right and shift+left/shift+right skip static elements, like spaces, when navigating highlightable elements.

https://user-images.githubusercontent.com/5647344/191027017-edba66f5-4d09-4ee2-9b43-307c62811d2e.mov

Fixes A11-1639, A11-1653

### DIFF
--- a/src/Nri/Ui/Highlighter/V1.elm
+++ b/src/Nri/Ui/Highlighter/V1.elm
@@ -211,7 +211,7 @@ type KeyboardMsg
     | SelectionExpandLeft Int
     | SelectionExpandRight Int
     | SelectionApplyTool Int
-    | Space Int
+    | ToggleHighlight Int
 
 
 {-| Possible intents or "external effects" that the Highlighter can request (see `perform`).
@@ -404,16 +404,19 @@ keyboardEventToActions msg model =
                 Tool.Eraser _ ->
                     [ Remove, ResetSelection, Focus index ]
 
-        Space index ->
+        ToggleHighlight index ->
             case model.marker of
                 Tool.Marker marker ->
-                    [ Toggle index marker ]
+                    [ Toggle index marker
+                    , Focus index
+                    ]
 
                 Tool.Eraser _ ->
                     [ MouseOver index
                     , Hint index index
                     , MouseUp
                     , Remove
+                    , Focus index
                     ]
 
 
@@ -639,7 +642,7 @@ viewHighlightable highlighterId marker focusIndex highlightable =
                 , on "touchstart" (Pointer <| Down highlightable.groupIndex)
                 , attribute "data-interactive" ""
                 , Key.onKeyDownPreventDefault
-                    [ Key.space (Keyboard <| Space highlightable.groupIndex)
+                    [ Key.space (Keyboard <| ToggleHighlight highlightable.groupIndex)
                     , Key.right (Keyboard <| MoveRight highlightable.groupIndex)
                     , Key.left (Keyboard <| MoveLeft highlightable.groupIndex)
                     , Key.shiftRight (Keyboard <| SelectionExpandRight highlightable.groupIndex)


src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -134,7 +134,7 @@ init config =
     , hasChanged = NotChanged
     , selectionStartIndex = Nothing
     , selectionEndIndex = Nothing
-    , focusIndex = 0
+    , focusIndex = 2
```
tesk9, line 137: I think it's _incidentally_ true that the first focusable element in the example text is the second position, but I don't think it's guaranteed, unfortunately.

src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -302,50 +302,74 @@ update msg model =
                 ( model, Cmd.none )
 
 
+nextInteractive : Int -> List (Highlightable marker) -> Int
+nextInteractive index highlightables =
+    if Maybe.map .type_ (List.Extra.getAt index highlightables) == Just Highlightable.Interactive then
```
tesk9, line 307: I wonder if we could change `nextInteractive` and `previousInteractive` to take the _current_ index, instead of the index offset by 1 -- I'm thinking that we if did this (and changed the implementation to use a fold rather than recursion), then the places we use `nextInteractive` and `previousInteractive` could be simpler.

I think `foldl`/`foldr` would be more performant here -- at least, I think there would be fewer operations in most cases: `List.Extra.getAt` will need to do an operation at least `index` number of times  ([`getAt`](https://github.com/elm-community/list-extra/blob/8.6.0/src/List/Extra.elm#L125) uses [`drop`](https://github.com/elm/core/blob/1.0.5/src/List.elm#L622), which recurses through the undesired elements of the linked list to remove them.

tesk9 (commented):
Works great!!! Feels much nicer to use -- much less of a caterpillar-y feel 🦋 

Could you please associate the Linear story? And then I'm also starting to wonder about tests -- do you think those could be added as part of this PR, or are you thinking a separate PR or separate story..? (I'd be very happy to pair -- Elm test setup can be a bit involved 😅 )

src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -302,74 +302,94 @@ update msg model =
...
+        (\x ( nextIndex, hasIndexMatched ) ->
+            if hasIndexMatched then
+                ( x.groupIndex, False )
 
-    else
-        index
+            else
+                ( nextIndex, x.groupIndex == index )
+        )
+        ( index - 1, False )
```
tesk9, line 344: I know we discussed not having a Maybe here, but I think I thought of an edge case, unfortunately!

Suppose our highlightables are something like:

```
Static 0 | Static 1 | Static 2 | Interactive 3
```

If we're focused on Interactive 3, and we hit the left arrow, I think we'll end up on Static 2. And actually we should stay on Interactive 3.

I'm thinking that the out-of-bounds checks will need to move into `nextInteractive` and `previousInteractive`, and we'll need to use a Maybe. 🤔

tesk9:
> Linear story is associated in the title :)

Oops, I missed it! Thanks!

src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -618,27 +663,17 @@ viewHighlightable highlighterId marker focusIndex highlightable =
...
-                    ]
                 ]
                 (Just marker)
                 highlightable
 
 
 viewStaticHighlightable : String -> Highlightable marker -> Html msg
 viewStaticHighlightable highlighterId =
-    viewHighlightableSegment False -1 highlighterId [] Nothing
+    viewHighlightableSegment False (Just -1) highlighterId [] Nothing
```
tesk9, line 673: 💭 Does it make sense for `Just -1` to be `Nothing` instead?

tests/Spec/Nri/Ui/Highlighter.elm
```diff
@@ -0,0 +1,315 @@
...
+                |> done
+    , test "selects element on MouseDown, expands selection on MouseOver, and highlights selected elements on MouseUp" <|
+        \() ->
+            Highlightable.initFragments Nothing "Pothos indirect light"
+                |> program marker
+                |> ensureFocusOn "Pothos"
+                |> mouseDown "Pothos"
+                |> mouseOver "indirect"
+                |> mouseUp "Pothos"
+                |> ensureMarked [ "Pothos", "", "indirect" ]
```
tesk9, line 147: ❓ Why is the ensureMarked argument `[ "Pothos", "", "indirect" ]`, rather than `[ "Pothos", " ", "indirect" ]`? I think we want to highlight the space, right?

tests/Spec/Nri/Ui/Highlighter.elm
```diff
@@ -0,0 +1,315 @@
...
+        |> ensureView
+            (Query.findAll [ Selector.attribute (Key.tabbable True) ]
+                >> Query.count (Expect.equal 1)
+            )
+        -- We only manage focus for interactive elements.
+        -- Static elements should not have explicit tabindex -1 as it is redundant,
+        -- they are never included in the tab sequence
+        |> ensureView
+            (Query.findAll [ Selector.attribute (Key.tabbable False) ]
+                >> Query.count (Expect.equal 2)
```
tesk9, line 211: I think the assertion about the number of interactive elements belongs in a specific test, rather than a helper. I'd like for `ensureFocusOn` to potentially be usable when there are more than 3 interactive highlightables, you know?

tesk9 (commented):
Looking really good! Loving all the tests.

I'm wondering if we could add a failing test, and then merge https://github.com/NoRedInk/noredink-ui/pull/1095 in? It seems like the space tests should have a failure, since focus is lost on space on this branch currently. Or this PR could merge, and I could make sure the fixed focus behavior works on my branch. Up to you!

PR #1092, opened 2022-09-20
Bat/hover color

tesk9:
Fixes A11-1619

### Before
button hover color is azure

### After
button hover color is azureDark

cc @NoRedInk/design 

Adjust hover text color for secondary buttons.


bendansby:
We like the clean wins

tesk9:
It's nice to do an easy fix without anything complicated or any trade-offs at all!

BrianHicks (approved):

PR #1094, opened 2022-09-22
Permit row headers in `Nri.Ui.Table`

charlieshanley:
This adds`rowHeader` (alongside the existing `string` and `custom`) to the `Nri.Ui.Table` interface for creating columns whose cells are row headers -- IE `th`s with `scope = "row"`.

I encountered this need in the course of [ADM-1](https://linear.app/noredink/issue/ADM-1/implement-a11y-notes-on-admin-tools-page).


tesk9 (approved):
Awesome, thanks!

PR #1095, opened 2022-09-23
Fix space focus loss

tesk9:
Fixes A11-1636

1. Tab into the highlighter area.
2. Use arrow keys to focus on a single word.
3. Press space. The word should be highlighted **and your focus should remain on the word**
4. Use tab and the arrow keys again to focus on the word that's highlighted.
5. Press space. The word should be unhighlighted **and your focus should remain on the word**

https://user-images.githubusercontent.com/8811312/192010439-a55b0f6a-0078-4d12-85b3-5584d866cabd.mov


tesk9:
Would you like this to merge into your branch, or wait to merge into master after your branch merges..?

Issue #1098, opened 2022-09-26
Change all disabled states of components to use new styles


bendansby:
Visual specs: https://cloud.antetype.com/exports/link/156e2993-0281-44d4-a1d8-e9400cb51686?starting=1&toolbar=true

<img width="600" alt="Disabled" src="https://user-images.githubusercontent.com/13528834/192327265-12378db3-275d-4f06-82be-c3184677edef.png">

These components need their styles updated:
- [x] Button
- [x] Select
- [x] Text input
- [x] Switch
- [x] Checkbox
- [x] Radio button
- [ ] Text area
- [x] Menu

tesk9:
@bendansby which of the remaining would you like a dev to take on?

bendansby:
Sadly all the rest are beyond my abilities (at least without jumping into a giant sinkhole of time)!

tesk9:
The list in the issue is missing Menu, right?

bendansby:
Ah yes, added!

PR #1100, opened 2022-09-26
Tweak button state styles

bendansby:
Improve contrast of disabled state by adding a border:
image

Approximates fix for button item in https://github.com/NoRedInk/noredink-ui/issues/1098
Approximates because the bottom border is too thick

Change unfulfilled state to look enabled (because it is!)
image

Fixes https://github.com/NoRedInk/noredink-ui/issues/1011

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -906,8 +906,8 @@ getColorPalette (ButtonOrLink config) =
             , hoverBackground = Colors.gray92
             , text = Colors.gray45
             , hoverText = Colors.gray45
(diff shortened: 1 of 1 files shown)


tesk9 (approved):
Nice!

blunderdome:
I thought the header of a quick write with minimum word count not met looks not-as-good-as-before with the secondary action and unfulfilled buttons right next to each other:

image

bendansby:
Good call; we should move to "Disabled"/Read-only

bendansby:
Going with "secondary" for now https://github.com/NoRedInk/NoRedInk/pull/40915

Issue #1044, opened 2022-08-25
PremiumCheckbox customized styles fail to apply when locked for Premium


tesk9:
Custom styles apply correctly when the default premium display is used:
<img width="524" alt="Screen Shot 2022-08-24 at 6 02 30 PM" src="https://user-images.githubusercontent.com/8811312/186550634-244d2cf0-55ea-4a32-a3d4-8d67b017d426.png">

Custom styles disappear when the premium display is PremiumLocked:
<img width="544" alt="Screen Shot 2022-08-24 at 6 02 40 PM" src="https://user-images.githubusercontent.com/8811312/186550630-7bd7c6bf-5af8-4fe2-af5d-2159795d7233.png">

PR #1101, opened 2022-09-26
Disabled checkbox styles

bendansby:
Approximates fix for checkbox item of https://github.com/NoRedInk/noredink-ui/issues/1098

Approximates because using a grayscale filter means the colors in the checkbox are inexact.

image


tesk9 (approved):
Neat! I haven't used a grayscale filter before. Very cool

Heads up that the styles aren't applied for PremiumCheckboxes with an "OnLockdClick" action. I think this is somewhat part of https://github.com/NoRedInk/noredink-ui/issues/1044

image

PR #1102, opened 2022-09-26
Tessa/icon for narrow mobile

tesk9:
Support clickablesvgs having custom icons by breakpoint, with 2 new breakpoint customizations: narrow mobile and quiz engine mobile.


tesk9:
Ping, @zwilias 😄

zwilias (approved):
I spent quite a bit of time thinking about how that `case (x, y, z) of` expression could be made nicer, but I honestly can't come up with anything that doesn't introduce more indirection.

Anyway, if we even need to add even more icons and breakpoints, we'll have to revisit 😄 I do think we can come up with something equivalent to a fold, at that point!

hide

PR #1104, opened 2022-09-29
A11-1631 add highlighter screen-reader text

charbelrami:
adds sr-only text to indicate when entering and exiting a highlight

### DIFF
--- a/src/Nri/Ui/Highlighter/V1.elm
+++ b/src/Nri/Ui/Highlighter/V1.elm
@@ -615,7 +615,7 @@ groupContainer viewSegment highlightables =
                             , Css.Global.children
                                 [ Css.Global.selector ":first-child"
                                     (Css.before
-                                        [ Css.property "content" ("\" [start" ++ (Maybe.map (\name -> " " ++ name) markedWith.name |> Maybe.withDefault "") ++ "] \"")
+                                        [ Css.property "content" ("\" [start " ++ (Maybe.map (\name -> name) markedWith.name |> Maybe.withDefault "highlight") ++ "] \"")
                                         , invisibleStyle
                                         ]
                                         :: MediaQuery.highContrastMode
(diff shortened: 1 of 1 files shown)


src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -613,19 +613,55 @@ groupContainer viewSegment highlightables =
                             [ Css.backgroundColor Css.transparent
                             , Css.Global.children
                                 [ Css.Global.selector ":first-child"
-                                    (MediaQuery.highContrastMode
-                                        [ Maybe.map
-                                            (\name ->
-                                                Css.before
-                                                    [ Css.property "content" ("\"[" ++ name ++ "] \"")
-                                                    ]
+                                    ((Maybe.map
```
tesk9, line 616: I wonder if this and the end-highlight Maybe.map logic could be simplified? I think since we always want a before and after pseudo-element, we could move the maybe.mapping "in" further, to where it's used:

```
(Css.property "content"
    ("\" [start"
        ++ (Maybe.map (\name -> " " ++ name)
                markedWith.name
                |> Maybe.withDefault ""
           )
        ++ "] \""
    )
    :: srOnlyStyles
)
```

src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -613,19 +613,55 @@ groupContainer viewSegment highlightables =
                             [ Css.backgroundColor Css.transparent
                             , Css.Global.children
                                 [ Css.Global.selector ":first-child"
```
tesk9, line 615: Right now, the before and after elements are added to the first and last span children of the mark. 

Screen Shot of the DOM highlighting the structure

I'm wondering if instead, we should have the pseudo-elements belong to the mark itself? Conceptually, it seems like that's where it belongs to me 🤔 I think I added the before element to the first child for the high-contrast mode for visual reasons, rather than for semantic reasons

tesk9: Ahh okay! 👍

tesk9 (commented):
Definitely works to communicate highlight status! 🎉 and code is looking good!

One thing I'm not sure how to fix because I'm not sure why it's happening -- it seems like the "Start highlight" announcement is being read twice.


https://user-images.githubusercontent.com/8811312/193304253-8118f5be-15b4-45ab-b6d4-006f557f1b53.mov

An easier-to-resolve usability issue is that I realized that when there's not a "name" for the highlight, the :before and :after should probably have "highlight" in there so it's clear what they're for. e.g., Instead of "Sphinx of [start]black[end] quartz, judge my vow", "Sphinx of [start highlight]black[end highlight] quartz, judge my vow"


https://user-images.githubusercontent.com/8811312/193304993-b24bd979-f4d9-4a11-9dc6-d02deb48a5ac.mov

tesk9 (approved):
I'll slack you my VO settings! Thanks! (Feel free to merge whether you happen to figure out what's happening with the double-announcement or not 😄 )

PR #1107, opened 2022-10-04
Ensure labels aren't visible to users when hidden.

lindsaykwardell:
Fix KRA-392

The hidden labels for radio buttons are leaking into view due to the 1px width, causing them to appear as dots running down the screen. The goal of this PR is to resolve that, so the letters don't appear.


lindsaykwardell:
I started this PR with a straightforward change (reduce the width to 0px, from 1px), but I'm uncertain what a11y ramifications this could have, and am very open to suggestions.

lindsaykwardell:
Switched to using `invisibleStyle`, as it's a more standard option.

PR #1118, opened 2022-10-05
Tessa/header

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/922

Extracts Nri.Header from the monolith.
Sets up for use as a reusable component, including adding an example on the styleguide.

And finally, changes the styleguide to actually _use_ the new Header component.

Screen Shot 2022-10-05 at 3 04 06 PM

cc @NoRedInk/design


tesk9:
Ping @zwilias 😄

tesk9:
Ping, @zwilias

PR #1121, opened 2022-10-06
add application role

charbelrami:
Fixes A11-1718 and A11-1719

Adjusts NVDA keyboard behavior, making it match VoiceOver and ChromeVox. Prevents NVDA from navigating and reading one letter at a time. Prevents loss of focus when pressing the space key.

> "Some user agents and assistive technologies have a browse mode where standard input events, such as up and down arrow key events, are intercepted and used to control a reading cursor. This browse mode behavior prevents elements that do not have a widget role from receiving and using such keyboard and gesture events to provide interactive functionality."

https://w3c.github.io/aria/#application

### DIFF
--- a/src/Nri/Ui/Highlighter/V1.elm
+++ b/src/Nri/Ui/Highlighter/V1.elm
@@ -595,7 +595,6 @@ view_ viewSegment { id, highlightables } =
         |> p [ Html.Styled.Attributes.id id, class "highlighter-container", attribute "role" "application" ]
 
 
-
 {-| When elements are marked, wrap them in a single `mark` html node.
 -}
 groupContainer : (Highlightable marker -> Html msg) -> List (Highlightable marker) -> List (Html msg)


tesk9 (commented):
Won't [`role=application`](https://www.w3.org/TR/wai-aria-1.1/#application) break the static experience for some assistive technologies?

> Because only the focusable elements contained in an application element are accessible to users of some assistive technologies, authors MUST use one of the following techniques to ensure all non-decorative static text or image content inside an application is accessible:
> 
> 1. Associate the content with a focusable element using [aria-labelledby](https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby) or aria-describedby.
> 2. Place the content in a focusable element that has role [document](https://www.w3.org/TR/wai-aria-1.1/#document) or [article](https://www.w3.org/TR/wai-aria-1.1/#article).
> 3. Manage focus of descendants as described in [Managing Focus](https://www.w3.org/TR/wai-aria-1.1/#managingfocus), updating the value of [aria-activedescendant](https://www.w3.org/TR/wai-aria-1.1/#aria-activedescendant) to reference the [element](https://www.w3.org/TR/wai-aria-1.1/#dfn-element) containing the focused content.

tesk9:
Thanks for testing it so thoroughly!

PR #1122, opened 2022-10-06
Tessa/organize colors

tesk9:
Begin organizing color examples into smaller groups that are sorted by luminance. Also adds rgb version of the color, just in case someone is looking for a color that way instead of by hex.

### Before
Screen Shot 2022-10-06 at 12 16 09 PM

### After
Screen Shot 2022-10-06 at 12 15 57 PM

cc @NoRedInk/design

### DIFF
--- a/styleguide-app/Examples/Colors.elm
+++ b/styleguide-app/Examples/Colors.elm
@@ -301,4 +301,16 @@ viewColor ( name, color, description ) =
                 ]
             ]
             [ Html.text color.value ]
+        , Html.p
+            [ css
+                [ Css.fontSize (Css.px 14)
+                , Fonts.baseFont
+                , Css.margin Css.zero
+                ]
+            ]
+            [ color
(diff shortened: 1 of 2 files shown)


tesk9:
Sorry my brain is not parsing that -- could you say that another way? Maybe with an example color? 😅

bendansby:
These are kind of "legacy" highlighter colors that we had before we created a more systematic set of highlighter colors for self review and guided drafts (which are shown in the "highlighter colors" sections):
Napkin 15 10-06-22, 4 06 17 PM

I think we ultimately want to use the same set of highlight colors across assignments, so I'm suggesting we remove the legacy colors from the noredink-ui example.

tesk9:
How's this, with magenta and cyan in their own clear deprecated section?

localhost_8000_

bendansby:
That works 👍

ap-nri (approved):

PR #1123, opened 2022-10-06
Tessa/add nri panel

tesk9:
Moves Nri.Panel to noredink-ui.

Screen Shot 2022-10-06 at 4 09 20 PM

### Secondary styles
Screen Shot 2022-10-06 at 4 09 25 PM

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Panel/V1.elm
+++ b/src/Nri/Ui/Panel/V1.elm
@@ -149,7 +149,7 @@ view customizations =
                         Css.backgroundColor Colors.navy
 
                     Secondary ->
-                        Css.backgroundColor Colors.gray75
+                        Css.backgroundColor Colors.gray45
                 , Css.padding2 (Css.px 8) (Css.px 15)
                 , Fonts.baseFont
                 , Css.fontSize (Css.px 16)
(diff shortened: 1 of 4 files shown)


src/Nri/Ui/Panel/V1.elm
```diff
@@ -0,0 +1,181 @@
...
+    , theme = Primary
+    , css = []
+    , headerCss = []
+    , containerCss = []
+    }
+
+
+{-| -}
+type Attribute msg
+    = Attribute (Config msg -> Config msg)
```
BrianHicks, line 58: I've tried to explain this function-passing style to folks during onboarding and it's always confusing. Could this maybe just be a custom type with a reducer, for that reason?

BrianHicks (approved):
just a little comment which is fine if left un-addressed!

src/Nri/Ui/Panel/V1.elm
tesk9: I think theoretically folks shouldn't need to learn this pattern when they're onboarding, since it's mostly used in noredink-ui, which is owned by a11ybats?

Also I think almost all of the components use this style now -- I think Balloon.V1 might be the only list-of-attributes hold out at this point. Personally I like it better because all the updates are in one place, and it makes it straightforward to have shared patterns across components -- like with `InputErrorAndGuidanceInternal`

tesk9: I hear you on it being confusing though. I'll think about it, for noredink-ui on the whole

PR #1124, opened 2022-10-07
Add buffer for mobile modal height

bendansby:
The variable viewport height of Mobile Safari means that you might not be able to see the entire modal unless you have scrolled, so this adds a buffer to avoid that. Before and after shown below.

https://user-images.githubusercontent.com/13528834/194604626-a7d7dea4-0fcf-4087-9738-114e22317065.mp4


tesk9 (approved):
I dig it!

PR #1125, opened 2022-10-07
Remove flexwrap from breadcrumbs container

tesk9:
V3 of breadcrumbs should probably make the wrapping-or-not customizable. For now, prevent the wrapping.

Fixes https://noredink.slack.com/archives/C02NMHB1K55/p1665106112182239

### Before
Screen Shot 2022-10-07 at 2 43 57 PM

### After
Screen Shot 2022-10-07 at 2 43 15 PM

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/BreadCrumbs/V2.elm
+++ b/src/Nri/Ui/BreadCrumbs/V2.elm
@@ -255,7 +255,6 @@ navContainer label =
     styled nav
         [ alignItems center
         , displayFlex
-        , flexWrap wrap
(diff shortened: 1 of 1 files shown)


bendansby:
Thanks!

tesk9:
How would you feel about flex-wrap being enabled for mobile? the Percy snapshots have found some problems on the styleguide without it

bendansby:
That sounds good. I'll keep an eye on it once it propagates.

PR #1133, opened 2022-10-14
Highlighter supports MultiHighlighter answer views

tesk9:
Fixes A11-1747

Screen Shot 2022-10-14 at 4 09 19 PM

## For Reviewer
- Please check that you don't notice any changes to the Screen Reader experience. (Not expecting changes -- just it's possible, and worth double-checking)
- Please check the high contrast mode for windows looks nice. (I wasn't paying attention to whether I was logged into an assistive labs session and I _fully_ went through all 120 minutes 😭 no new testing for me until early Nov)

### DIFF
--- a/src/Nri/Ui/Highlighter/V1.elm
+++ b/src/Nri/Ui/Highlighter/V1.elm
@@ -737,11 +737,11 @@ viewInlineTag { showTagsInline, isInteractive, maybeTool } highlightable =
                         , Css.boxShadow5 Css.zero (Css.px 1) (Css.px 1) Css.zero Colors.gray75
                         , Css.display Css.none
                         , if showTagsInline then
-                            Css.batch [ Css.display Css.inline |> Css.important, MediaQuery.highContrastMode [ Css.property "forced-color-adjust" "none", Css.color Colors.black |> Css.important ] ]
+                            Css.batch [ Css.display Css.inline |> Css.important, MediaQuery.highContrastMode [ Css.property "forced-color-adjust" "none", Css.property "color" "initial" |> Css.important ] ]
 
                           else
                             Css.batch
-                                [ MediaQuery.highContrastMode [ Css.property "forced-color-adjust" "none", Css.display Css.inline |> Css.important, Css.color Colors.black |> Css.important ]
+                                [ MediaQuery.highContrastMode [ Css.property "forced-color-adjust" "none", Css.display Css.inline |> Css.important, Css.property "color" "initial" |> Css.important ]
                                 ]
                         ]
                     , -- we use the :before element to convey details about the start of the
--- a/src/Nri/Ui/Highlighter/V1.elm
+++ b/src/Nri/Ui/Highlighter/V1.elm
@@ -662,8 +662,8 @@ groupContainer showTagsInline viewSegment highlightables =
                     List.map viewSegment highlightables
 
 
-tagBeforeContent : Bool -> { mark | name : Maybe String } -> Css.Style
-tagBeforeContent showTagsInline markedWith =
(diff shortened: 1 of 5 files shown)


charbelrami:
The screen reader experience works properly 🙌 - tested with VO, ChromeVox (on a Chromebook machine), and NVDA (on a Windows machine).

High-contrast mode doesn't look nice in dark themes: 

high-contrast mode on windows - 1
high-contrast mode on windows - 2
high-contrast mode on windows - 3

tesk9:
Thank you!!

charbelrami:
@tesk9, does it look nice?

# Windows

## Not high-contrast mode
windows-not-high-contrast-mode

## High-contrast mode

### Themes
* Aquatic
windows-high-contrast-mode-aquatic

* Desert
windows-high-contrast-mode-desert

* Dusk
windows-high-contrast-mode-dusk

* Night sky
windows-high-contrast-mode-night-sky

# macOS

## Not increased contrast
macos-not-increased-contrast


## Increased contrast 
macos-increased-contrast

src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -737,10 +737,12 @@ viewInlineTag { showTagsInline, isInteractive, maybeTool } highlightable =
                         , Css.boxShadow5 Css.zero (Css.px 1) (Css.px 1) Css.zero Colors.gray75
                         , Css.display Css.none
                         , if showTagsInline then
-                            Css.display Css.inline |> Css.important
+                            Css.batch [ Css.display Css.inline |> Css.important, MediaQuery.highContrastMode [ Css.property "forced-color-adjust" "none", Css.color Colors.black |> Css.important ] ]
```
tesk9, line 740: @charbelrami is there a system color we could use instead of black?

charbelrami: replaced with `color: initial`

tesk9: I haven't used `Css.property "forced-color-adjust" "none"` before, do we still need it?

src/Nri/Ui/MediaQuery/V1.elm
```diff
@@ -92,13 +92,14 @@ prefersReducedMotion =
 {-| -}
 notHighContrastMode : List Style -> Style
 notHighContrastMode =
-    withMediaQuery [ "(forced-colors: none)" ]
+    withMediaQuery [ "(forced-colors: none) and (prefers-contrast: no-preference)" ]
```
tesk9, line 95: I'm not sure about the prefers-contrast: no-preference 🤔 

If the user has set that they `prefers-contrast: less`, I think we should consider them to _not_ be in high contrast mode. But I think this query would not match, even though it should. Right?

charbelrami: Right, I replaced it with `not(prefers-contrast: more)`. Thanks!

src/Nri/Ui/Highlighter/V1.elm
charbelrami: Yes, otherwise the user agent on Windows will force its own colors which results in inline tags with dark foreground on dark background and we would not be able to change them.

tesk9:
Thank you for fixing the contrast mode, @charbelrami ! I appreciate it!!!

src/Nri/Ui/MediaQuery/V1.elm
charbelrami: Accessibility -> Display -> Increase contrast
Screen Shot 2022-10-19 at 13 11 46

tesk9: Awesome, thank you!

PR #1135, opened 2022-10-21
Update disabled checkbox styles

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/1098

Screen Shot 2022-10-21 at 2 00 02 PM

cc @NoRedInk/design 

I also changed the checkmark and tilde color from #146AFF to azure. LMK if that's not cool and I'll undo it -- kinda assuming it was supposed to be azure?


bendansby:
Awesome, thanks so much!

charbelrami (approved):

PR #1138, opened 2022-10-24
Bat/fix highlighter start element

tesk9:
Fixes A11-1767

Ensures that the start element is rendered.

[Media queries don't support negation of just part of the query](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#inverting_a_querys_meaning), which means that the not high contrast media query was invalid, which meant the start element was never rendered.

I am removing the `prefers-contrast` portions of the media query entirely. So far, we've used the `highContrastMode` media query to fix places where colors have been overwritten automatically so that the content is still perceivable. We have not customized styles for cases when users have expressed a preference, but it would be up to us to change color palettes.

> Though prefers-contrast and forced-colors are both about forcing a certain contrast, the difference is pretty big: forced-colors changes all your sites colors, while with prefers-contrast it's your job to do that. With Forced colors your styling gets overwritten, and with prefers-contrast the visitor is probably okay with your styling, they just want more (or less!) contrast.
[Polypane post on contrast modes](https://polypane.app/blog/forced-colors-explained-a-practical-guide/)

I think these 2 use cases are distinct, and should be handled separately.

### DIFF
--- a/src/Nri/Ui/MediaQuery/V1.elm
+++ b/src/Nri/Ui/MediaQuery/V1.elm
@@ -92,7 +92,7 @@ prefersReducedMotion =
 {-| -}
 notHighContrastMode : List Style -> Style
 notHighContrastMode =
-    withMediaQuery [ "(forced-colors: none) and (not(prefers-contrast: more))" ]
+    withMediaQuery [ "(forced-colors: none) or (prefers-contrast: no-preference) or (prefers-contrast: less)" ]
 
 
 {-| -}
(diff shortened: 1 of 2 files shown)


charbelrami:
> Media queries don't support negation of just part of the query

😯

> I am removing the prefers-contrast portions of the media query entirely

macOS **increase contrast** will no longer work. But considering that **increase contrast** is not equivalent to high-contrast mode, I agree that we should handle these cases separately.

charbelrami (approved):

tesk9:
Thanks for reviewing! I didn't know about media query negation or the details about prefers-contrast, so this was a good learning experience for me 😅

PR #1141, opened 2022-10-26
Allow setting Message codeDetails [QUE-475]

omnibs:
Right now our choice is either a generic error message and helpful context for engineers, or a nicer error message and no helpful context for engineers.

With this PR we can do `[Message.markdown "Some helpful error message that reassures the user, tells them how to work around stuff and perhaps even where to look for help", Message.codeDetails "Fluctuations in the bi-neural electroplasma subinterlink."]`


tesk9:
I added the new helper to the example -- look okay to you, @omnibs ?

image

PR #1144, opened 2022-10-27
Adds Checkbox.V7

bcardiff:
A shinny new checkbox has appeared.

- Reworked api similar to other components based on Attributes
- Add support for guidance
- Dropped checkboxLockOnInside / Themes functionality. [*].
- Dropped disabledLabelCss functionality. Use labelCss instead in case when the checkbox is disabled. [*]
- (breaking-change) By default the label is visible (ie: V6.viewWithLabel), use hiddenLabel to migrate from V6.view.

[*] It seems to no longer be needed in monolith

- Maybe the examples should be migrated to use a similar structure as others.
- Maybe we could include `errorIf`, `errorMessage` for completeness, but there is no use case so far.

### DIFF
--- a/src/Nri/Ui/Checkbox/V7.elm
+++ b/src/Nri/Ui/Checkbox/V7.elm
@@ -6,6 +6,7 @@ module Nri.Ui.Checkbox.V7 exposing
     , selectedFromBool
     , containerCss, labelCss, custom, nriDescription, id, testId
     , disabled, enabled, guidance
+    , viewIcon
     )
 
 {-|
@@ -34,6 +35,11 @@ module Nri.Ui.Checkbox.V7 exposing
 @docs selectedFromBool
 @docs containerCss, labelCss, custom, nriDescription, id, testId
 @docs disabled, enabled, guidance
+
+
+### Internal
+
+@docs viewIcon
 
 -}
 
--- a/src/Nri/Ui/Checkbox/V7.elm
+++ b/src/Nri/Ui/Checkbox/V7.elm
@@ -383,7 +383,7 @@ viewEnabledLabel config icon =
         ]
         [ viewIcon [] icon
         , labelView config
-        , InputErrorAndGuidanceInternal.view config.identifier config
+        , InputErrorAndGuidanceInternal.view config.identifier (Css.marginTop Css.zero) config
         ]
 
 
@@ -414,7 +414,7 @@ viewDisabledLabel config icon =
         ]
         [ viewIcon [] icon
         , labelView config
-        , InputErrorAndGuidanceInternal.view config.identifier config
+        , InputErrorAndGuidanceInternal.view config.identifier (Css.marginTop Css.zero) config
         ]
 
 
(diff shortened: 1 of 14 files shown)


src/Nri/Ui/Checkbox/V7.elm
```diff
@@ -0,0 +1,480 @@
...
+    , hiddenLabel, visibleLabel
+    , selectedFromBool
+    , containerCss, labelCss, custom, nriDescription, id, testId
+    , disabled, enabled, guidance
+    )
+
+{-|
+
+
+# Changes from V6:
```
tesk9, line 14: Thanks for including these lovely upgrade notes!

tesk9 (commented):
Code looks fabulous! 😍 And the guidance slots in there so nicely!

The only concern I have is around the removal of the Theme (and maybe the removal of the disabled label styles?). Checkbox.V6 is used by PremiumCheckbox -- so the Locked theme is used in the monolith _by way of_ the PremiumCheckbox. It's possible the other styles are too? At any rate, I think it's worth upgrading the noredink-ui uses of Checkbox V6 to V8, to expose anything else that may still be necessary.

It also might be a good idea to ping Ben with a screenshot now (I slightly suspect that there may be too much whitespace between the label and the guidance).

bcardiff:
Internal dependencies updated. Although I preserve PremiumCheckbox.disabledLabelCss functionality, it can also be pruned directly since it's unused in monolith. But that would be a breaking change on PremiumCheckbox. So I preserve it for now.

bcardiff:
@bendansby these is the current look and feel of the guidance. It's the same as in RadioButton. When a guidance is used the label is pushed up and the whole height changes a little bit.

Is this 👍 or we need to tweak the layout?

Without guidance:

Screen Shot 2022-10-28 at 11 02 02

With guidance:

Screen Shot 2022-10-28 at 11 01 56

bendansby:
Tessa called it, looks great except the label and guidance should be a wee bit closer together.
image

src/Nri/Ui/Checkbox/V7.elm
```diff
@@ -0,0 +1,494 @@
...
+    List.foldl (\(Attribute update) config -> update config)
+        beginningConfig
+        attributes
+
+
+{-| View a checkbox (the label is only used for accessibility hints unless visibleLabel attribute is applied)
+-}
+view :
+    { label : String
+    , selected : IsSelected
```
stoeffel, line 216: Why don't we add onCheck to the config as well? It's never not set.
Additionally, I wonder if we should force adding an id by moving it to the config.

bcardiff: I would not mind moving them, but so far all components have the event handler and id setter in the attributes list. The mandatory arguments are the requirements for using them in regular forms + a11y if I understand correctly the design.

tesk9: The concept for having onCheck/other event handlers not in the main record is that as a dev you can make progress _towards_ using a component incrementally. Instead of needing to understand a big record full of options, I can start with something like: `Button.button "My button" []` and have _something_ show up, and I can work through everything else I want one piece at a time.

Also it makes the "preview" and example code features easier to support

PR #1146, opened 2022-10-28
Fix modal focus trap when the last id changes on the fly

tesk9:
See [Slack](https://noredink.slack.com/archives/C02NVG4M45U/p1666815109380059) for original context.

It seems that having the key event listener using Json.Decode.oneOf to decode events is causing the decoder to not get updated. Investigation steps:

- [using focus trap Ellie](https://ellie-app.com/jZJX3fkRnc3a1)
- [using when focus leaves decoder directly Ellie](https://ellie-app.com/jZK3Qzfj3D3a1)
- [moving the decoder code from when focus leaves into the view Ellie](https://ellie-app.com/jZK8xJbn66Va1)
- [removing the Decode.oneOf wrapping from the decoder Ellie](https://ellie-app.com/jZK9kMg7FCva1)

Underlying problem is already issued against elm/json here: https://github.com/elm/json/issues/15

### DIFF
--- a/src/Nri/Ui/FocusTrap/V1.elm
+++ b/src/Nri/Ui/FocusTrap/V1.elm
@@ -7,7 +7,8 @@ module Nri.Ui.FocusTrap.V1 exposing (FocusTrap, toAttribute)
 -}
 
 import Accessibility.Styled as Html
-import Accessibility.Styled.Key as Key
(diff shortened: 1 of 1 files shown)


micahhahn (approved):
Thank you so much for debugging this!

PR #1148, opened 2022-11-01
A11-1786 Add missing highlightable and highlighted classes

charbelrami:
Adds `"highlighter-inline-tag"` and `"highlighter-inline-tag-highlighted"` classes to inline tags, and `"highlighter-highlighted"` class to segments. Necessary to match the current monolith highlighter styles. In the monolith, some styles are applied when segments or inline tags are highlighted.


tesk9 (approved):
Looks reasonable! Although I don't have full context on these classes 😅

charbelrami:
In the monolith, some styles are applied only when segments are highlighted

```css
._8d6fb878 .highlighter-highlighted {
    line-height: 2.5em;
    padding: 6px 0;
}
```

Multi-highlighter segments with extra padding when highlighted

PR #1149, opened 2022-11-02
Adds another utility for creating safe Id strings

erik-nri:
While implementing something with Checkboxv7 we noticed an issue where when the id was being created from the label it would break the filter on the svg causing the fill to be black. In this case it was due to the label having something like "Angela's account" in it which broke the id on the comma.

### DIFF
--- a/src/Nri/Ui/Util.elm
+++ b/src/Nri/Ui/Util.elm
@@ -29,6 +29,7 @@ removePunctuation =
 
 
 {-| Creates a lowercased string that is safe to use for HTML IDs.
+Ensures that nonletter characters are cut from the front and replaces bad characters with a dash
 -}
 safeIdString : String -> String
 safeIdString =
--- a/src/Nri/Ui/Util.elm
+++ b/src/Nri/Ui/Util.elm
@@ -1,4 +1,4 @@
-module Nri.Ui.Util exposing (dashify, removePunctuation)
+module Nri.Ui.Util exposing (dashify, removePunctuation, safeIdString)
 
 import Regex
 
@@ -26,3 +26,22 @@ removePunctuation =
                 |> Maybe.withDefault Regex.never
     in
     Regex.replace regex (always "")
+
+
+{-| Creates a lowercased string that is safe to use for HTML IDs.
+-}
+safeIdString : String -> String
+safeIdString =
+    let
+        unsafeChar =
+            Regex.fromString "(^[^a-zA-Z]+|[^a-zA-Z0-9_-]+)" |> Maybe.withDefault Regex.never
+    in
+    Regex.replace unsafeChar
+        (\{ index } ->
+            if index == 0 then
+                ""
+
+            else
+                "-"
+        )
+        >> String.toLower
--- a/src/Nri/Ui/Util.elm
+++ b/src/Nri/Ui/Util.elm
@@ -16,13 +16,13 @@ dashify =
 
 
 {-| Convenience method for removing punctuation
-(removes everything that isn't whitespace or alphanumeric).
+(removes everything that isn't whitespace, alphanumeric, or an underscore).
 -}
 removePunctuation : String -> String
 removePunctuation =
     let
         regex =
-            Regex.fromString "[^A-z0-9\\w\\s]"
+            Regex.fromString "[^A-Za-z0-9\\w\\s]"
                 |> Maybe.withDefault Regex.never
     in
     Regex.replace regex (always "")
--- a/src/Nri/Ui/Util.elm
+++ b/src/Nri/Ui/Util.elm
@@ -34,15 +34,15 @@ Ensures that nonletter characters are cut from the front and replaces bad charac
 safeIdString : String -> String
 safeIdString =
     let
-        nonAlphaAtStartOfString =
+        nonAlphaAtStart =
             -- From the start of the string, match the contiguous block of
             -- not ASCII letters at the start of the String.
-            -- Note majiscule letters are before miniscule in the range,
-            -- so [A-z] is the same as [A-Za-z] as above in removePunctuation
-            "^[^A-z]+"
+            -- Why [a-zA-Z] and not [A-z]?
+            -- There are punctuation characters in that range: [\]^_` all come after Z and before a
+            "^[^a-zA-Z]+"
 
         nonAlphaNumUnderscoreHyphenAnywhere =
-            -- match any contiguous block of letters that aren't any of
+            -- any contiguous block of characters that aren't any of
             -- + ASCII letters
             -- + numbers
             -- + underscore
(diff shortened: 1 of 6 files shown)


src/Nri/Ui/Util.elm
```diff
@@ -26,3 +26,22 @@ removePunctuation =
...
     Regex.replace regex (always "")
+
+
+{-| Creates a lowercased string that is safe to use for HTML IDs.
+-}
+safeIdString : String -> String
+safeIdString =
+    let
+        unsafeChar =
+            Regex.fromString "(^[^a-zA-Z]+|[^a-zA-Z0-9_-]+)" |> Maybe.withDefault Regex.never
```
tesk9, line 37: I think we need to keep numbers, no? 

Could we use replacePunctuation here?

tesk9: Ohhhh I think I was reading the regex wrong

tesk9: I am curious though -- could we use removePunctuation?

tesk9 (commented):
Thanks for fixing this! it's definitely been a problem for a long time 🙈 I'm glad to see a helper like this! so useful!

Would you mind adding some unit tests before this merges? Because tbh I struggle to read regex but I'm good at reading test cases 😅

src/Nri/Ui/Util.elm
erik-nri: I've tagged @ap-nri as he is the regex mastermind behind this =D. We were messing around with this one day during elm bookclub and this was the implementation that we landed on to solve this problem somewhere in monolith. I must confess Regex might as well be hieroglyphics to me so I hope Alex can further elaborate/remember what we talked about in bookclub.

erik-nri:
Oh definitely! I'll get some tests in right away 😄

bcardiff:
cc: @celsobonutti , there are some `String.Extra.dasherize` occurrences in contento that might be worth updating with this behaviour, right?

erik-nri:
Tests added @tesk9! I spent a little time trying to figure out if I could fuzz test this but I struggled with figuring out how to do it for the case where leading non-alpha characters are removed. I think it would be cool to get working but I wanted to get some basic tests in first 😄

tests/Spec/Nri/Ui/Util.elm
```diff
@@ -0,0 +1,33 @@
...
+                    Util.safeIdString "#@!!232now we are in business"
+                        |> Expect.equal "now-we-are-in-business"
+            , test "hard mode" <|
+                \() ->
+                    Util.safeIdString "!@#21something else interesting321$ ... hi"
+                        |> Expect.equal "something-else-interesting321-hi"
+            , test "with capitol letters" <|
+                \() ->
+                    Util.safeIdString "1232!@#%#@JFEKLfds-----SFJK3@#@jj23FDS........''''\"\"***"
+                        |> Expect.equal "jfeklfds-sfjk3-jj23fds-"
```
tesk9, line 27: I think I would expect to keep the starting numbers here? Just in case someone did something like `safeIdString "1-premade"`

tesk9 (commented):
Wooo thanks for the tests! Much appreciated.

I mentioned on 1 test line, but I think about 3 of the cases are actually impacted: I would expect to keep leading numbers, not just mid and trailing numbers. Just in case!

tests/Spec/Nri/Ui/Util.elm
ap-nri: Yeah when I wrote the thing for Elm book club I was going with the rule most programming languages have for identifiers. In fact that follows the [HTML4 spec](https://www.w3.org/TR/html4/types.html#type-id) but not HTML 5 and up. MDN [recommends](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id#sect2) starting with a letter, but you're right that this has a pretty obvious path to creating duplicate IDs which are an actual problem, so I'll change that.

ap-nri: ehhhh never mind. It does not work. Javascript cannot handle it in `querySelector` or `querySelectorAll`, CSS cannot handle it. It only works with `getElementById`. https://codepen.io/ap-nri/pen/OJENQLY

src/Nri/Ui/Util.elm
```diff
@@ -26,3 +26,58 @@ removePunctuation =
...
+        unsafeChar =
+            [ nonAlphaAtStartOfString
+            , nonAlphaNumUnderscoreHyphenAnywhere
+            ]
+                |> anyOfThese
+                |> regexFromString
+
+        collapsePunctuationToOne =
+            -- match any contiguous group of either underscore or hyphen-minus characters
+            regexFromString "[_-]+"
```
ap-nri, line 67: note that I dropped the `\s`. That character class matches any whitespace character, but we already got rid of those with `Regex.replace unsafeChar` so we don't need it in here.

src/Nri/Ui/Util.elm
ap-nri: `removePunctuation` was written wrong. The char range `A-z` starts at `A` and goes to `z` including all of the punctuation marks between `Z` and `a` which are, in order with some context around them:
```
XYZ[\]^_`abc
```

ap-nri:
Added a grip of stuff. Cool if you're cool.

src/Nri/Ui/Util.elm
ap-nri: see my comments in the code for more of how this works.

src/Nri/Ui/Util.elm
```diff
@@ -16,13 +16,74 @@ dashify =
...
 {-| Convenience method for removing punctuation
-(removes everything that isn't whitespace or alphanumeric).
+(removes everything that isn't whitespace, alphanumeric, or an underscore).
 -}
 removePunctuation : String -> String
 removePunctuation =
     let
         regex =
-            Regex.fromString "[^A-z0-9\\w\\s]"
+            Regex.fromString "[^A-Za-z0-9\\w\\s]"
```
ap-nri, line 25: "But Alex, why is `[A-Za-z]` different from `[A-z]` when the majuscule letters precede the minuscules in ASCII (and thus UTF-8) character encoding?? Surely these are the same"

No, young learner. They are not. For in the before time, in the long long ago, we decided to be *clever*. And endowed as we are with a mere 26 letters in English and 32 being a nice round number in binary representation<sup>1</sup>, we decided that for any majuscule character stored in memory as a number, we could *idempotently* find its minuscule by performing a bitwise OR with the number 32. To do this we would represent "A" with 65 and "a" with 97 and pad out the middle with the square brackets, backslash, underscore, and the backpack.  And of course character ranges in regexes respect that and match only those punctuation marks in between Z and a.

But let's turn to some other troubling bits of this: `\w` means "any word character" which is a bit rich. It contains `A-Z` and `a-z` and `0-9` and `_` to boot. But it also matches just about any roughly Latin alphabet word character according to a number of regex testers (a fair bit short of *any* word character I'd say) but that's not even real life. Here's me trying to make a URL slug or an ID for a French seasonal fashion line:

```
> summer = "éte"+String.fromCharCode(0x0301)
'été'
> for(let i = 0, c = -1; !isNaN(c); ++i ) { c =  summer.charCodeAt(i); console.log(i, c)}
0 233
1 116
2 101
3 769
4 NaN
undefined
> summer.replace(/[^\w]/g, '-') // replace every non-word character with a hyphen-minus
'-te-'
> summer.replace(/\W/g, '-') // replace every non-word character with a hyphen-minus the other way
'-te-'
```

So really this regex could just be `Regex.fromString "[\\W\\S]"` but I'd like to know the author's intent first. 

---

1. Were that the soviets had won out with their nice 32 character alphabet in Russian<sup>2</sup>
2. yes I know the other Cyrillic languages use more letters<sup>3</sup>
3. yes I know that the ternary [Setun](https://en.wikipedia.org/wiki/Setun) may have made this a moot point)

tests/Spec/Nri/Ui/Util.elm
tesk9: Huh! I had no idea that numbers at the start were no good!

tesk9 (approved):
Dang thanks for all the explanation & the improvements & the tests! Great stuff!

erik-nri:
Yer a wizard Alex!

src/Nri/Ui/Util.elm
ap-nri: @tesk9 @erik-nri you're on a ride with me now. This isn't used in the monolith. Rather, in the monolith there's `ui/src/Util` which exports `removePunctuation` which is used only in `ui/src/Page/Learn/PeerReviews/Rate/Paragraph/BinaryJustificationStep.elm` in the function `generateId` which should PROBABLY just use our new safeIdString. Anyone wanna hackday with me on that tomorrow?

 Also remember when I said "I'd like to know the author's intent first" well the author is *in the room right now!*
https://github.com/NoRedInk/NoRedInk/commit/d712227a1a7c0d9e05d742b164fab556b07c655b

ap-nri: HAHAHAHAH we are laughing we are rejoicing we are having the very best time.

tesk9: 😆 the author isn't super good at regex! safeIdString for everybody everywhere!

PR #1150, opened 2022-11-02
Update accessible-html-with-css version

tesk9:
Starting goal is to fix peer review highlighter issues (Fixes A11-1769). However, because the underlying issue seems to have been https://github.com/elm/json/issues/15, the fix ended up being quite involved, and also ended up touching other components that also had keyboard event listeners.

In particular, we should carefully check the keyboard behavior of:
- Accordions
- Menus (both the disclosure style and the list of options style)
- Tooltips (just the disclosure style)
- Modals
- Tabs
- Carousels

Note that this will cause a major version bump, so I'll also be doing some clean up in another branch while we're shaking things up for this package.


charbelrami (approved):
Looks great! Keyboard works as expected 🙌

PR #1154, opened 2022-11-04
use one pattern for generating IDs

ap-nri:
We have a few different patterns to generate safe IDs. This consolidates on one of them and eliminates some unused code around that.

The `Utils` file only ever dealt with these functions, and since they are related to `Html.Attributes` I put the functions [in there](https://github.com/NoRedInk/noredink-ui/pull/1154/files#diff-8026630b1912d3b7566f094cd36ae5b065123f6ee3f3c6f9d9cc43e8b476fe84). Since that is an exported module we may use these functions in the monorepo to generate safe IDs one way in there too! I am amenable to more changes in that file such as
* Add string safety to `testId`, `nriDescription`, and `nriDescriptionSelector` but I would need to figure out what is problematic in `data-` attributes.
* Add a general helper for string attribute safety that we can use in the aforementioned places and export (again, more looking needed—there may not be anything that's generally unsafe!)
* only prepend `id-` when the start of the given ID string is unsafe. I didn't like the potential flakyness of that, but given how many tests use `cy.get('#` it may be worth doing!!

So I'm closing https://github.com/NoRedInk/NoRedInk/pull/41209 but I am leaving the link here. That's where I got the idea of prepending a known safe prefix `id-` rather than dropping leading unsafe characters (see [this codepen](https://codepen.io/ap-nri/pen/OJENQLY) to find out more of what's safe and what's unsafe when for IDs). The intermediate commits had a lot more rules and stuff, but I think they unnecessarily introduced the possibility of collision for the sake of making the ID strings more aesthetic or something. Silly.

The effort to update the monorepo can be a different chunk of work; it's impossible(?) to automatically test what may break when changing IDs (I'm thinking in particular of focus capture) so introducing this One True Function here, dog fooding it in NRI-UI, and allowing for tiny nibble changes in the monorepo is I think the best approach.

When this merges and is used in the monorepo a few Capybara and Cypress tests may break where we look for IDs. In particular Cypress. I am happy to help fix those!

Paired with @asterite on this on Hackday 2023-06-02


tesk9 (approved):
Ahh I am such a fan of this cleanup! Thank you for all of these fab improvements!

erik-nri (approved):

tesk9:
@ap-nri are you ready for this to merge?

LearningNerd:
Just found another place that would benefit from this:
https://github.com/NoRedInk/NoRedInk/pull/41654

tesk9 (approved):
Looks great, thanks!

PR #1160, opened 2022-11-10
Block labels

tesk9:
Fixes A11-1789

Screen Shot 2022-11-14 at 3 22 14 PM

https://user-images.githubusercontent.com/8811312/201782007-f6dfb004-066f-4385-9a66-6638c5a8137a.mov

cc @NoRedInk/design

### DIFF
--- a/deprecated-modules.csv
+++ b/deprecated-modules.csv
@@ -1,2 +1,3 @@
+Nri.Ui.Balloon.V1,upgrade to V2
 Nri.Ui.Checkbox.V6,upgrade to V7
 Nri.Ui.Tabs.V6,upgrade to V7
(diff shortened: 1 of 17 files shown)


tesk9:
Oops! the labels have box shadow, and shouldn't. I'll remove it as part of A11-1803

charbelrami:
In the following case the label is centered relative to the invisible highlight and not to the blank element. Is this expected?

example of a label centered relative to the invisible highlight

tesk9:
Yes, I think that's fine! One thing we could do is auto-emphasize when there's a label, but let's figure that out when we do the review with Phoenix. Thanks!

PR #1166, opened 2022-11-18
A11-1838 NVDA and JAWS: fixes focus loss when pressing space on a highlightable with marker or eraser

charbelrami:
This keeps the current highlightable focused when pressing space to toggle or erase a highlight.


src/Nri/Ui/Highlighter/V1.elm
```diff
@@ -485,13 +485,28 @@ pointerEventToActions msg model =
 
                         _ ->
                             []
+
+                focus =
```
tesk9, line 489: I thought we decided against changing the mouse behavior? or am I misunderstanding the purpose of `focus` here?

charbelrami: It doesn't change the behavior, it just keeps focus on the highlightable after pressing space.

PR #1167, opened 2022-11-18
Checkbox guidance error location

tesk9:
Move the guidance and error message out of the checkbox's label.

Also, create a hit area "bridge" between the icon and the label text for checkbox and radio button. This doesn't change the behavior for Checkbox, but it does mean an improvement for radio buttons:

https://user-images.githubusercontent.com/8811312/202780555-16d5c179-f2e1-4271-8fb1-222d42ec5ab1.mov

Paired with @bcardiff


bcardiff (commented):
🙈

PR #1169, opened 2022-11-22
"use strict" is fine

BrianHicks:
I'm changing some stuff up in the monolith builds, and the new setup adds "use strict" at the top of the exported file. Turns out we can't allow that! (we do `CustomElement = …` which is not allowed in strict mode, as it refers to the name without defining it.)

Anyway, this makes all our JS use "use strict" so this stops happening, and defines `CustomElement` instead of just assigning it. 👍

### DIFF
--- a/lib/CustomElement.js
+++ b/lib/CustomElement.js
@@ -1,3 +1,5 @@
+"use strict";
+
 /**
  * @module CustomElement
  */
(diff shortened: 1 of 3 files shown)


BrianHicks:
asking a11ybats for review because I *think* y'all own NRI-ui now? Please let me know if that's incorrect!

krinhorn:
@BrianHicks yup, that's us!

tesk9 (approved):
Looks reasonable to me!

PR #1174, opened 2022-11-28
Enable creation of Menus with html list tags

mandla-noredink:
Menu sub items are currently not contained within `<li>` tags or collected under a `<ul>` tag.

This PR builds this functionality into the Menu component using a `Menu.isList` attribute, to ensure backwards compatibility with the existing functionality that allows for the arbitrary grouping of elements inside a `fieldset` tag.


mandla-noredink:
Percy failure is from updates to Menu example page

tesk9 (commented):
This is looking good!

Small problem, though. I think right now, to end up with a `ul` containing the `li`s, a dev needs to know to use `isList` _with_ `group`. If you use `li` with plain `entry`, you end up with a `div` containing `li`s.

Screen Shot 2022-11-28 at 3 10 09 PM

mandla-noredink:
> This is looking good!
> 
> Small problem, though. I think right now, to end up with a `ul` containing the `li`s, a dev needs to know to use `isList` _with_ `group`. If you use `li` with plain `entry`, you end up with a `div` containing `li`s.
> 
> Screen Shot 2022-11-28 at 3 10 09 PM

Ah, great catch, thank you!

tesk9 (commented):
Structure is looking better!

There are still a couple of issues, though.

### Styling on ul breaks the arrow positioning
Screen Shot 2022-11-29 at 12 33 03 PM

### `ul` has role=dialog
`ul` has an implicit role=list. role=dialog doesn't make sense contextually. I believe we want role=menu.

Screen Shot 2022-11-29 at 12 35 11 PM

The [APG example](https://www.w3.org/WAI/ARIA/apg/example-index/menu-button/menu-button-links.html) is structured with a ul with role=menu, lis with role=none, and then the links with the menuitem role. I think this is roughly what we want.
image

tesk9 (approved):
Thank you!!!

PR #1175, opened 2022-11-29
Marker and eraser styles

tesk9:
Fixes A11-1993

1. Adds transitions to highlighter styles so that it's less flash-y to hover over multiple lines
2. Adds "bright" hover styles to default highligher
3. Adds default eraser tool styles
4. Adds marker with border, and fixes the high contrast styles for it

### "bright" hover styles
Screen Shot 2022-11-29 at 12 52 32 PM

### with-border styles
Screen Shot 2022-11-29 at 12 52 11 PM

### High-contrast with-border styles
Screen Shot 2022-11-29 at 12 49 43 PM

cc @NoRedInk/design nothing big to review here, I don't think. Just pinging to keep you in the loop!

### DIFF
--- a/src/Nri/Ui/HighlighterTool/V1.elm
+++ b/src/Nri/Ui/HighlighterTool/V1.elm
@@ -110,7 +110,8 @@ hoverStyles : Css.Color -> List Css.Style
 hoverStyles color =
     List.append
         sharedStyles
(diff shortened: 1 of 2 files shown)


charbelrami (approved):
Looks great!

PR #1177, opened 2022-11-30
Keep outline around transparently for high contrast os modes

tesk9:
One of the [technique examples for a two-toned focus ring](https://www.w3.org/WAI/WCAG21/Techniques/css/C40) shows the outline as transparent instead of `none`, which actually causes the outline to become visible in high contrast mode! (A transparent border is why our checkboxes look kinda funky in high contrast mode -- transparent borders and such get coerced to be visible in high contrast mode).

https://user-images.githubusercontent.com/8811312/204913293-d6eac024-0796-4bea-a370-66bde80eb49f.mov

cc @NoRedInk/design 

Fixes A11-1929


charbelrami (approved):
Looks good!

PR #1178, opened 2022-12-01
Menu list suggestions

tesk9:
@mandla-noredink I was still seeing some axe failures in https://github.com/NoRedInk/noredink-ui/pull/1174.

image

I think the failures were in part coming from the Menu trying to support dialog lists and disclosure lists. I tried a refactor here that only supports the "list" pattern for one particular case. I also adjusted some of the roles a bit more.

What do you think about this?


mandla-noredink:
This looks good, thank you!

mandla-noredink (approved):

PR #1181, opened 2022-12-01
Fix issues with z-index on switch

mfonism:
This PR removes the z-index from the switch component, and resolves issues where the the part of the switch on which that z-index is set is _painted_ on modals and other elements which should usually cover and hide the entire switch.

__Before:__

https://user-images.githubusercontent.com/50708034/205177743-5d941e64-e7ad-4e35-89e6-1a62119159eb.mov

__After:__

https://user-images.githubusercontent.com/50708034/205179021-43ad8f4f-c5eb-438c-98cd-825d66c26c2c.mov

Resolves [ZEN-713](https://linear.app/noredink/issue/ZEN-713/weird-button-overlap-on-integrations-page)


tesk9 (approved):
Great!

PR #1182, opened 2022-12-06
Question box omnibus

tesk9:
Fixes A11-2026

Screen Shot 2022-12-07 at 4 28 07 PM

Screen Shot 2022-12-07 at 4 27 36 PM

cc @NoRedInk/design


tesk9:
Forgot to include high-contrast mode screenshots!

It looks okay in high contrast mode. The panda icon doesn't render correctly, but once we have a real svg instead of an svg with an image tag inside it, I think it will be fine, since the UiIcons are fine.

Screen Shot 2022-12-08 at 12 49 29 PM

image

staceyadams:
Sorry I missed the ping 3 days ago! These may be covered by a different issue but noting here in case they're not.

1. Can the button widths match the left/right edges of the speech bubble (excluding the point) without super complex logic? The pink lines below indicate roughly where. And with more padding on the left so everything between the pink lines is still centered.

Pasted_Image_12_8_22__3_13_PM

2. Is there a max-width in place? (I'm not sure we talked through the [UI requirements](https://paper.dropbox.com/doc/Scaffolding-Block-Requirements-Source-of-Truth-Doc-cHk8S6y1Ab2JIRnCB5Ldr#:h2=UI-requirements) of the requirements doc)

3. Can we try small buttons instead of medium buttons? That feels more proportionate to the size of the text in the speech bubble. Full disclosure: after testing with 3rd graders, we may decide to bump the button size up. Maybe also the text size too. My big struggle right now is to make it 3-5 friendly but also 6-12 friendly and consistent with the rest of the site.

4. Is Gumby the AIM icon? Double nostalgia!

tesk9:
Thank you for the feedback, @staceyadams!!

I scootched the buttons to be lined up with the text bubble (when there's a character, and the text bubble renders). I also added the max-width (I found the 440 value in the doc -- sorry I missed it initially!) and I switched to small buttons.

Screen Shot 2022-12-12 at 11 24 59 AM
Screen Shot 2022-12-12 at 11 25 18 AM
Screen Shot 2022-12-12 at 11 25 06 AM

> Is Gumby the AIM icon? 

😆 He's `UiIcon.stretch` ! He's from learning modules

staceyadams:
> > Is Gumby the AIM icon?

> 
> 😆 He's UiIcon.stretch ! He's from learning modules

omggg 😆😆😆 

Thanks for the changes!

Is the no-character version a fallback in case we don't have a character to display? For scaffolding, we plan to always have the character/speech bubble. If this version is needed, can we keep the white container but remove the pointy bit that turns it into a speech bubble? I think the text is easier to read on a white background.

tesk9:
> Is the no-character version a fallback in case we don't have a character to display? 

Yup! it's also what I started with, since the code I got from Phoenix didn't include it. I kept it optional because it seemed like there were character experiments to run. We can make the character required and/or make the panda the default, if you want! 

Some options:
1. character is optional. Devs must explicitly include the panda icon if they want it (or any character) to show up.
2. character is optional. If not overriden, the panda will show up.
3. character is always the panda. If a different icon is desired, devs will need to change it in noredink-ui.

staceyadams:
👍 Let's keep current behavior for now until we figure out more requirements for characters. I could see a world where we don't show characters to high schoolers or in the student character selection version of the interests page, there's an option to select "no character"

staceyadams:
Or #2 since we want to do classroom testing ASAP and the panda fits those needs. When we're working on the character project, we can add in the override for no character at that point if needed?

charbelrami (approved):
Looks great!

tesk9:
> When we're working on the character project, we can add in the override for no character at that point if needed?
 
 Definitely!
 
 --
 
 Do you have a preference between the current behavior and #2 (possibly with the no-character setting?)? I'm very happy to do either!

staceyadams:
If current behavior is no character, I have a preference for #2. As far as priority of what we need in the short term, the no-character setting isn't necessary but is something we definitely want to keep in mind when we work more on characters next quarter. If it's easier to add in now because that state already exists, go for it!

tesk9:
Okay! Panda is the default, but devs can override to _not_ show a character or to use a different character.

If overridden to not show a character, it looks like this:

image

PR #1184, opened 2022-12-07
A11-2039 remove brackets from pseudo-elements

charbelrami:
This removes the square brackets from the highlighter pseudo-elements because NVDA reads them.


tesk9 (approved):
Makes sense, thank you!

Issue #1185, opened 2022-12-08
ClickableSvg: Rename tertiary style and add new tertiary style matching the tertiary style of Button


bendansby:

tesk9:
How does `quaternary` sound? 😅

bendansby:
Plus one!

tesk9:
Cool! I think probably a two-phase change is good idea here:

- dupe the current tertiary styles and call them quaternary
- release, upgrade, find-and-replace all tertiary uses to quarternary
- and then edit the tertiary styles to match the existing tertiary styles

I can get started on it, if you want...? LMK

PR #1186, opened 2022-12-08
Dupe tertiary styles to quaternary

tesk9:
First step of fixing https://github.com/NoRedInk/noredink-ui/issues/1185

Dupes the current ClickableSvg tertiary styles as `quaternary`. (Have I spelled that right everywhere? What a weird word)

image


bendansby (approved):
Checks out!

PR #1187, opened 2022-12-08
Adjust tooltip bridge

tesk9:
Fixes A11-1904

Relevant to KRA-610

## Before this PR

It turns out that the tooltip component actually had 2 hover bridges!

The rectangle highlighted in mustard had an nri description claiming it to be the hover bridge.
Screen Shot 2022-12-08 at 11 06 20 AM

But it was actually _underneath_ another element that was functioning as the hover bridge, highlighted in green:
Screen Shot 2022-12-08 at 11 06 51 AM

Note that both hover bridges are actually rendering on top of the tooltip trigger, which was causing the covered portion of the element to not be clickable.

You might be thinking "I know! Pointer events none will allow the clicks to travel through to the trigger!" Totally! But this will also make the hover bridge not actually keep the tooltip open when the user hovers it.

hover bridge

---

## After this PR

I've removed the extraneous hover bridge.

I've removed the overlap from the remaining hover bridge. I think that when there's a triggering element with rounded edges, we may lose some of the bridge behavior where the trigger has a cut-out. However, I think this is a reasonable tradeoff (and if the user mouses to the tooltip area quickly, it won't matter).

Screen Shot 2022-12-08 at 11 03 41 AM

Screen Shot 2022-12-08 at 11 03 53 AM

Screen Shot 2022-12-08 at 11 04 00 AM

Screen Shot 2022-12-08 at 11 03 48 AM

Another option is to mess with a negative z-index to get the bridge behind the trigger. However, I think this would be a mistake to do because it will require adding non-static positioning to all tooltip triggering elements -- a risky move! Also, this non-static positioning would be challenging to override (we wouldn't want to use `css`, because of potential class name re-ordering surprises, and inline styles are high precedence).

Note that the green background is for demo purposes. It is not present in the actual PR.

### DIFF
--- a/src/Nri/Ui/Tooltip/V3.elm
+++ b/src/Nri/Ui/Tooltip/V3.elm
@@ -1010,6 +1010,7 @@ hoverBridge { isOpen, direction } =
             [ Css.boxSizing Css.borderBox
             , Css.padding (Css.px tailSize)
             , Css.position Css.absolute
+            , Css.backgroundColor Colors.red
             , Css.batch <|
                 case direction of
                     OnTop ->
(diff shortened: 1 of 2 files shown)


bcardiff (approved):
Awesome!

PR #1191, opened 2022-12-09
Lighten greenDark

bendansby:
Keeps it at the 3:1 threshold but a bit more vibrant since we're using it more now.

image


tesk9 (approved):
Looks good! I'll merge when tests pass

PR #1194, opened 2022-12-12
Bat/stretched balloons

tesk9:
Blocks need balloons that can have extra-tall arrows in order to position them against each other nicely (See https://linear.app/noredink/issue/A11-1804/overlapping-tagged-labels)

This PR only adds support for arbitrarily-tall balloon arrows.

Screen Shot 2022-12-12 at 3 55 28 PM

## High Contrast Note:
This PR also changes the appearance of balloon arrows in high-contrast mode (because I had to change approach from a rotated square to the border-trick in order to get the arrow to extend).

## Before

Screen Shot 2022-12-12 at 4 54 09 PM

## After

Screen Shot 2022-12-12 at 4 51 05 PM

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Balloon/V2.elm
+++ b/src/Nri/Ui/Balloon/V2.elm
@@ -555,6 +555,10 @@ arrowPosition ({ position, arrowAlignment, theme } as config) =
         color =
             theme.backgroundColor
 
+        highContrastColor =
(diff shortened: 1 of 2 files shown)


staceyadams:
Cool! The extra tall arrow was just one option for handling labels running into each other. Definitely open to other ideas too!

PR #1199, opened 2022-12-15
Fix overlapping block labels

tesk9:
Fixes A11-1804

To prevent overlapping labels on Blocks, we need to measure the content in the balloon and then adjust the arrow height accordingly. This requires some tasks to run to measure the page content. In the example page, I've wired the tasks to run on-click, rather than on load or window resize events.

Note that the current solution is not maximally vertically "compact." If it needs to be, I can keep working on this. I don't want to assume the MVP target, so pausing here for feedback. (See video if this is confusing)

Screen Shot 2022-12-15 at 10 55 23 AM

Screen Shot 2022-12-15 at 10 55 33 AM

https://user-images.githubusercontent.com/8811312/207933198-cacac5d8-7a57-4aaf-ad85-57b0490825b0.mov

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Balloon/V2.elm
+++ b/src/Nri/Ui/Balloon/V2.elm
@@ -6,7 +6,7 @@ module Nri.Ui.Balloon.V2 exposing
     , onBottom, onLeft, onRight, onTop
     , alignArrowStart, alignArrowMiddle, alignArrowEnd
     , arrowHeight
-    , custom, id, nriDescription, testId
+    , custom, id, contentId, nriDescription, testId
     , containerCss
     , css, notMobileCss, mobileCss, quizEngineMobileCss
     )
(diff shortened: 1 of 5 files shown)


staceyadams:
> Note that the current solution is not maximally vertically "compact." If it needs to be, I can keep working on this. I don't want to assume the MVP target, so pausing here for feedback. (See video if this is confusing)

This works for MVP! Agree that the example with color and age on the same level and the middle one in the second position is nicer but I don't think we need to solve for it yet. We'll revisit if it turns out that more than 2 labels next to each other are a frequent occurrence.

It might be helpful to explain why there's the measure & render button. Something like "Measure & render to see how this example will really look" but in way fewer words 😅 (or as helper text outside the button)

tesk9:
image

micahhahn (approved):
This looks great!!! I'm reviewing this in Brian's stead as he's out sick today and I'd really like to see this merged for ongoing work in displaying blocks!

tesk9:
Fabulous, thank you!

Just so you're aware, Stacey & I discussed the labels today, and we think there's actually some more work to be done positioning-wise, since there's a chance that labels get cut-off at the edge of the viewport.

image

So the API may change a little bit today/next week to account for horizontally offsetting labels if necessary at the edge of the screen.

PR #1201, opened 2022-12-16
Block example tweaks

tesk9:
Fixes A11-2070
Fixes A11-2105

Screenshot of the Block examples

cc @NoRedInk/design


staceyadams:
Woo!

One slight change: can the example text be left-aligned instead of centered?

tesk9:
Definitely!

image

What do you think about the label placement on the final element here? Okay, or...?

tesk9:
Oh also! @staceyadams do you want the examples in the table left-aligned too?

staceyadams:
Can we scooch the label over to the right a bit more? Is this only in the case where the beginning point is to the right of the end point? If so, maybe the width of about ~4-5 letters?

Yes to left-aligning the examples in the table too.

Thanks!

tesk9:
> Can we scooch the label over to the right a bit more so it's not right at the edge?

Yes, that sounds good. I will do this separately.

--

I'll left-align the table examples in this PR.

tesk9:
image

charbelrami (approved):
great!

PR #1203, opened 2022-12-19
Label width adjustments

tesk9:
Fixes A11-2071

Adds an explicit width to the absolutely positioned balloon content.

Screen Shot 2022-12-19 at 12 48 10 PM
Screen Shot 2022-12-19 at 12 48 22 PM

cc @NoRedInk/design (screenshots are _not_ using a nbsp. Max-width of the balloon is set to 150px).


staceyadams:
🙌

PR #1204, opened 2022-12-20
Bat/compact labels

tesk9:
Fixes A11-2112

- ensures that the vertically-lower labels are always positioned on top of the higher labels
- ensures that labels do not get cut-off by the viewport on the left or the right

Screen Shot 2022-12-20 at 12 50 46 PM
Screen Shot 2022-12-20 at 12 51 47 PM
Screen Shot 2022-12-20 at 12 51 25 PM

cc @NoRedInk/design


micahhahn (approved):
Amazing work as always Tessa!

PR #1207, opened 2022-12-21
Question box V2 -- positioning within the sentence, within viewport boundary, and inside of emphasis with blank segments

tesk9:
Fixes A11-2118 

localhost_8000_ (3)

cc @NoRedInk/design


tesk9:
Okey doke! Updated the title and description. I think this is ready to go once tests pass! 🎉

PR #1208, opened 2022-12-21
Question box  -- positioning wrt the viewport edges

tesk9:
Fixes A11-2118

https://user-images.githubusercontent.com/8811312/208976916-29959b4b-f432-4c08-a045-2837c263a431.mov

## Left edge
Screen Shot 2022-12-21 at 11 25 14 AM

## Right edge

I am not certain I can reasonably account for the character width when repositioning -- I think panda may just have to get cut off on the edge. Or he could sit all the way inside the question box? It didn't seem worth solving for right now, given that the character design may not be final.

Screen Shot 2022-12-21 at 11 25 19 AM

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Block/V2.elm
+++ b/src/Nri/Ui/Block/V2.elm
@@ -60,6 +60,7 @@ import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Html.Attributes.V2 as AttributesExtra exposing (nriDescription)
 import Nri.Ui.Mark.V2 as Mark exposing (Mark)
 import Nri.Ui.MediaQuery.V1 as MediaQuery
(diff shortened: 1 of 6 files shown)


micahhahn (changes requested):
This looks great! 

Something appears to be wrong with the style guide demo though... I know that percy isn't pressing "Measure & render", but I still am not imagining that the stacked question boxes are right: 

image

tesk9:
I'll try to fix percy separately if I have time. I am not a big fan of puppeteer. It's flakey (or I'm doing it wrong 🤷‍♀️ ).  I wouldn't worry about the percy screenshots though -- that's not reflective of how it looks before or after you click the measure & render button.

micahhahn (approved):

PR #1209, opened 2022-12-21
Allow more flexible QuestionBox anchors

tesk9:
Fixes A11-2104

Support anchoring the question box to either the entire emphasis-with-block or to a particular element within it.

image

### DIFF
--- a/script/puppeteer-tests.js
+++ b/script/puppeteer-tests.js
@@ -85,6 +85,20 @@ describe("UI tests", function () {
     await callback(option);
   };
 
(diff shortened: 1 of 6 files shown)


micahhahn:
@tesk9 Looks like there are axe failures from CI?

tesk9:
Fixed!

micahhahn:
Can we chat about this at our pairing tomorrow?  I'm trying to wrap my head around the new API, but I'm not entirely sure I've got a handle on it.

PR #1210, opened 2022-12-22
Question box -- clarify the API

tesk9:
- restructures the examples. The table now only shows the `pointingTo` views, since that's where the complexity comes in
- adds more code examples, to hopefully illustrate the API a bit better
- removes padding bottom helper from Block. instead, `Block.withQuestionBox`, `Block.wordWithQuestionBox`, and `blankWithQuestionBox` now require a `Maybe Dom.Element` measurement, which is used by Block to add the appropriate spacing and to mark the QuestionBox as of the `pointingTo` variety


micahhahn (approved):
Looks good! I think you have an elm type error in there that CI is failing on, but this is good to merge once its green!

tesk9:
Thanks! Missed the test compilation (again 🙈)

PR #1211, opened 2022-12-22
Allow focus on guidance

tesk9:
Fixes A11-2030

Adds `QuestionBox.guidanceId`, which will enable app devs to manage focus when transitioning between question boxes.


micahhahn (approved):
👍 

Same CI error as the other PR, maybe something is broken on master 😳

tesk9:
I think what's broken on master is, uh, the shake rules 😅 . I ran `shake` before I requested review from you, and it worked fine, but it didn't run the elm tests. Maybe `shake ci` has different rules -- I don't remember 🤔

PR #1217, opened 2022-12-29
Tessa/pagination

tesk9:
Fixes A11-1793


juanedi:
@tesk9 I can't seem to find a way to log into Percy and review the changes there. Mind if I leave that part to you? Thanks 🙏

PR #1219, opened 2023-01-04
narrower modal titles when X button present

ap-nri:
inspired by GRO-72

without button.

image

with button

image


tesk9 (approved):
Looks good to me!

bendansby (approved):
Our long translucent nightmare is over

PR #1220, opened 2023-01-04
Question box adjustments

tesk9:
Fixes A11-2159

### Before

https://user-images.githubusercontent.com/8811312/210859525-d40741da-cdb9-45ec-8874-b9c62a6490ae.mov

### After

https://user-images.githubusercontent.com/8811312/210859637-3f89837f-7133-45eb-800a-043539d77278.mov

cc @krinhorn 

### Before

Before

### After

After

cc @NoRedInk/design


src/Nri/Ui/QuestionBox/V3.elm
```diff
@@ -0,0 +1,424 @@
...
+standalone =
+    setType Standalone
+
+
+{-| This type of `QuestionBox` has an arrow pointing to the block whose id you pass in, if that block is on the final line of its paragraph.
+The QuestionBox should be vertically aligned beneath the block, viewport permitting.
+
+Pass in the id for the block the QuestionBox should point to.
+
+You will need to pass 2 measurements, taken using Dom.Browser, in order for the question box to be positioned correctly.
```
micahhahn, line 122: Should this be "3 measurements"?

micahhahn (commented):
This looks great Tessa! This API should be easier for us to use.

I do have a question: Are we concerned about not knowing what the box points to when we drop the arrow for non-last-line content? It seems like screen readers might have more information here than visual learners. 

Also, just a random thought, did we consider using a 50% transparent arrow if it's going to overlap content?

tesk9:
> Are we concerned about not knowing what the box points to when we drop the arrow for non-last-line content? It seems like screen readers might have more information here than visual learners.

I had some concern here too, but Stacey and co have done user testing that suggests arrow-less will be fine, and it's more important that the sentence be readable. Curriculum will be aware of the arrow/no-arrow situation as they author questions.

micahhahn (approved):

tesk9:
> Also, just a random thought, did we consider using a 50% transparent arrow if it's going to overlap content?

I didn't, personally. @staceyadams's figma shows that she has considered pretty much everything! Although maybe not partial transparency?

I would recommend against transparency, personally. It probably wouldn't work for high contrast mode, and I think it would really complicate the contrast situation/make the sentence harder to read (esp. for our beginning readers!)

staceyadams:
> I had some concern here too, but Stacey and co have done user testing that suggests arrow-less will be fine, and it's more important that the sentence be readable. Curriculum will be aware of the arrow/no-arrow situation as they author questions.

Fine, but not ideal. There's so much more I want to explore here like matching the QuestionBox color to the highlight color since that's customizable both for labels and emphasis now, but also there are probably cases where that doesn't make sense so not a quick decision. Realistically, due to bandwidth/timeline, I think we'll be launching with what we have now unless we find in classroom testing that students are having a difficult time making the connection.

> Also, just a random thought, did we consider using a 50% transparent arrow if it's going to overlap content?

I hadn't considered transparency but have the same concerns Tessa mentioned. I also don't think the arrow is that noticeable or necessary and isn't the make-it-or-break-it part of making the connection to the display element.

staceyadams:
@tesk9 Can we talk through this use case either here or elsewhere? (Not a blocker for this PR!)

image

Some thoughts off the top of my head:
* Can we detect which section is longer and attach it to the center of that point?
* Probably easier to always say attach to the start or end though, like in the case of labels? Does it have to be a location based on pixels (or other unit) or can we say first/last word?
* This could also be solved by going back to the previous idea of the QuestionBox only being positioned by a specific word instead of a set of words but this likely has a lot of ramifications in other places. One major con of this would be that for every tagged position an author sets that's more than one word, they would have to create a second one instead of using the same for the display element and scaffolding container.
* Dropping the arrow from QuestionBox entirely (see comment above for other arrow thoughts) is maybe the easiest solution?

staceyadams:
⬆️ I got that from the link you shared yesterday: https://deploy-preview-1220--noredink-ui.netlify.app/#/doodad/QuestionBox

I just realized that may no longer be the latest so feel free to ignore if already solved for!

krinhorn:
Thanks, Tessa! This is a good solution - adding some notes for folks interested in learning:

In terms of the accessibility features, I think the new implementation is less about providing a programatic association and more about providing convenience for screen reader users who may not be as easily able to scroll back or move their eyes back to the highlighted text. As @micahhahn  noted, there is no direct visual association for sighted students between the box and the highlight; or rather, we are making an assumption (or have tested?) that sighted students would understand that the question is about the highlighted text. My understanding is that Curriculum will write scaffolding tips such that there is no ambiguity about which words we're asking about, which effectively makes the highlights or arrows a non-essential usability feature for sighted folks.

If there were any _essential_ visual associations, we'd need to be sure there was an equitable alternative, whether programmatic, in text, or via the DOM order (like Tessa's previous implementation.)

(I'm a fan of dropping the arrows entirely; it could cause confusion if it seems to be pointing to something the question isn't specifically about, and it isn't particularly easy to see anyway. It's also easier for QA or anyone else to support fewer cases.)

The new `aria-details`/`aria-describedby` implementation would be a good candidate for doing user testing with screen reader users sometime. We might find that folks don't need the extra convenience, or that they don't want it, or that they prefer a hidden link to jump back to the highlight, or something else entirely.

tesk9:
We discussed synchronously! Going to remove the point. I'll merge this PR as-is, and make a separate PR removing the point for all conditions.

PR #1221, opened 2023-01-05
Ensure rightIcon attributes don't change pre-existing rendering

tesk9:
Fixes https://github.com/NoRedInk/NoRedInk/pull/41952#issuecomment-1372393691

Note that this removes rightIcon support from ClickableSvg. It is still available from ClickableText and Button.

See https://github.com/NoRedInk/noredink-ui/pull/1188 for original PR adding rightIcon


charbelrami (approved):

charbelrami:
> Note that this removes rightIcon support from ClickableSvg. It is still available from ClickableText and ClickableSvg.
I think you intended to say "It is still available from ClickableText and Button"

tesk9:
Thanks!

PR #1225, opened 2023-01-09
Allow caller of Nri.Ui.Message to specify they don't want an icon

charlieshanley:
I've started seeing message elements with the icons omitted in designs (see the validation error message [here](https://cloud.antetype.com/exports/link/6e567e77-d32e-410d-a982-2fbfebd57da4/?toolbar=true&starting=14) for an example).

This allows consumers of `Nri.Ui.Messge` to specify `noIcon` to omit the icon that comes by default with the theme, and still get the `Message` goodness.

Relates to [ADM-562](https://linear.app/noredink/issue/ADM-562/update-noredink-ui-to-have-messages-with-icon-omitted).


charlieshanley:
After consulting with A11ybats, Design had second thoughts about Messages without icons. Closing.

PR #1226, opened 2023-01-09
Remove QuestionBox point

tesk9:
localhost_8000_ (5)

cc @NoRedInk/design


micahhahn (approved):
Ah well it was good while it lasted

PR #1227, opened 2023-01-09
Highlighter supports auto-expanding adjacent highlights

tesk9:
Adds a new version of highlighter that automatically optionally supports going from this state:

Screen Shot 2023-01-09 at 2 39 53 PM

to this state:

Screen Shot 2023-01-09 at 2 39 59 PM

When the highlights are the same type.

When the highlights are _not_ the same type, e.g.:

Screen Shot 2023-01-09 at 2 41 23 PM

they will not be joined.

---

Why do this?

Formatting questions expect highlights to auto-join. In order to do A11-2089, this work is necessary.

Fixes A11-2089


charbelrami (approved):
great!

PR #1230, opened 2023-01-10
Re-implement ClickableSvg rightIcon

tesk9:
This PR avoids changing the styles for ClickableSvg _except_ for when the rightIcon is present, in order to avoid regressions.

Screen Shot of table showing gear clickable svgs with right icon down arrows

cc @NoRedInk/design


bendansby:
Do you think you could add a bit more spacing between the two icons, matching what we have on Manage Classes?

tesk9:
Screen Shot 2023-01-10 at 10 12 50 AM

Better?

charbelrami (approved):
looks good!

PR #1235, opened 2023-01-11
Bat/block adjustments

tesk9:
Fixes A11-2177

# Label text alignment
## Before
Screen Shot 2023-01-11 at 1 26 40 PM

## After

Screen Shot 2023-01-11 at 1 26 12 PM

# Label min width
## Before

Screen Shot 2023-01-11 at 1 27 07 PM

## After

Screen Shot 2023-01-11 at 1 27 18 PM

# About section

I think this looks kinda janky. At some point, I'd like to do a project to allow any example page to add arbitrary links to get added to the links section of the header in the upper right. But I don't think I have time to do so this week. I can make copy adjustments or move the link down in the page?

## After

image

---

cc @NoRedInk/design

### DIFF
--- a/src/Nri/Ui/Mark/V2.elm
+++ b/src/Nri/Ui/Mark/V2.elm
@@ -340,6 +340,7 @@ viewBalloon config label =
               -- if the border radius of the balloon changes, this value might need to change as well
               Css.minWidth (Css.px 30)
             , Css.maxWidth (Css.px 150)
+            , Css.textAlign Css.center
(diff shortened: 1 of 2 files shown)


staceyadams:
re: About section
What's there now meets my needs of just having a link on the page at all 👍 

Also open to changing ":additional things to know" to something else as the title doc. I couldn't come up with anything better at the time or now

micahhahn (approved):
Looks great!

Issue #1214, opened 2022-12-27
Add `openInNewTab` affordance automatically when `linkExternal` is used


tesk9:
This [slack conversation](https://noredink.slack.com/archives/C02NVG4M45U/p1669918324279939) has some discussion on what we'd like. I think ultimately, we should be adding the icon in the same spot/using the same pattern every time.

My inclination is use `rightIcon`, rather than `icon`, so that the accessible name includes the imp. extra info ("opens in a new tab") at the end rather than at the beginning.

I don't think we'll be able to add tooltips really conveniently -- those we'll still need to add by hand if we want them.

cc @NoRedInk/design thoughts?

bendansby:
Makes sense to me 👍

nyghtly-derek:
I encountered [an example](https://github.com/NoRedInk/NoRedInk/pull/42044) of this today where the new tab link is _within_ a tooltip. I'm not sure if that matters, but wanted to flag just in case there could a problem with tooltip inception.

tesk9:
This is un-done in https://github.com/NoRedInk/noredink-ui/pull/1246 for [reasons explained in slack](https://noredink.slack.com/archives/C9MEFL3KL/p1674515210335959) for ClickableSvgs. We'll need to figure something out for ClickableSvgs still!

PR #1237, opened 2023-01-13
Automatically add opens in new tab affordance

tesk9:
When a Button, ClickableText, ClickableSvg, or SideNav entry link to an external page, automatically add `UiIcon.openInNewTab` icon with "Opens in a new tab" alt text to the right side of the link.

Fixes https://github.com/NoRedInk/noredink-ui/issues/1214

Note that this change also required adding rightIcon support to SideNav:
Screen Shot 2023-01-13 at 2 15 17 PM

When upgrading to the version of noredink-ui that includes this change, we should be careful to remove existing left-icon affordances so that the icon doesn't show twice.

cc @NoRedInk/design

### DIFF
--- a/src/ClickableAttributes.elm
+++ b/src/ClickableAttributes.elm
@@ -1,5 +1,5 @@
 module ClickableAttributes exposing
-    ( ClickableAttributes, init
+    ( Config, ClickableAttributes, init
     , onClick, submit, opensModal
     , toButtonAttributes
     , href, linkWithMethod, linkWithTracking
@@ -10,7 +10,7 @@ module ClickableAttributes exposing
 
 {-|
 
-@docs ClickableAttributes, init
+@docs Config, ClickableAttributes, init
 
 
 # For buttons
--- a/src/ClickableAttributes.elm
+++ b/src/ClickableAttributes.elm
@@ -71,64 +71,77 @@ init =
 
 
 {-| -}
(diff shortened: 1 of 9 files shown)


bendansby:
How is the size of rightIcon determined? For example, it looks a bit too big in the sideNav example here.

tesk9:
The size of the rightIcon depends on the component -- there aren't shared sizing rules across Button, ClickableText, etc.

For SideNav specifically, I copy-pasted the existing icon rendering and changed the margin to be on the left instead of the right. I can easily make it smaller, if you don't want it to be the same size as the left icon.  Currently, it's 20px by 20px with 5px of left margin.

tesk9:
I wonder if partly it's a weight problem in the icon?

Putting it right next to the copy icon, the openInNewTab icon looks bigger and I think it might be because the lines are thicker?
image

staceyadams:
I think it's definitely the weight and maybe the arrow drawing you to look in a specific direction.

I was so sure the openInNewTab icon arrow would extend higher than the copy icon. I can't trust my eyes!

Pasted_Image_1_13_23__2_56_PM

tesk9:
Wait so should I make a change here?

bendansby:
I personally would make rightIcons a bit smaller in the SideNav example, but we don't need to decide on that here and now.

tesk9:
I'll hold off for now, then! thanks

charbelrami (approved):
looks good!

PR #1239, opened 2023-01-18
Bat/highlighter markdown

tesk9:
Fixes A11-2006

Adds a helper, `Highlightable.fromMarkdown : String.String -> List.List (Nri.Ui.Highlightable.V1.Highlightable ())`, that interprets empty links as needing to be tagged with highlights.


charbelrami (approved):
looks good!

PR #1240, opened 2023-01-20
A11-2191 Standardize ClickableText styles

charbelrami:
This standardize ClickableText styles in Balloon, Container, Heading, Message, Panel, and Text. - [A11-2191 : Standardize ClickableText/link behavior inside of Texts, Messages, etc.](https://linear.app/noredink/issue/A11-2191/standardize-clickabletextlink-behavior-inside-of-texts-messages-etc)


tesk9:
@charbelrami I'm not sure this matches the desired behavior 🤔

I'm pretty sure we want:
- disabled styles to show for links and for buttons
- clickabletext to _usually_ not have the border-bottom underlines
- all links (_including_ clickabletext links) within `Text`, `Message`, and any other component that supports rendering markdown to show up with styles identical to clickabletext styles, but _with_ border-bottom underlines

Right, @NoRedInk/design ?

bendansby:
Right, the underlines are for when ClickableTexts are within a paragraph to help differentiate them from the normal text. When CTs are standalone they don't need that differentiation.

charbelrami:
Thanks! Asked Ben about disabled links [here](https://noredink.slack.com/archives/C02NMHB1K55/p1674234423328789?thread_ts=1674139035.616359&cid=C02NMHB1K55). @tesk9 are you aware of any specific instances where this may be necessary?

tesk9 (commented):
I think that this PR may have inadvertently changed some spacing? I'm seeing a percy diff 👀 Ideally, I think, this PR wouldn't cause a big diff on the styleguide. If it does, I think it will be pretty challenging to integrate on the monolith.

I also wonder if we could share more of the custom styles? So that they're not repeated across the different components?

charbelrami:
thanks!

src/Nri/Ui/ClickableText/V3.elm
```diff
@@ -235,7 +235,8 @@ hideTextFor : MediaQuery -> Attribute msg
 hideTextFor mediaQuery =
     css
         [ Css.Media.withMedia [ mediaQuery ]
-            [ Css.Global.descendants
+            [ Css.borderStyle Css.none |> Css.important
```
tesk9, line 238: ❓ what is this border style removal for?

tesk9 (commented):
I think we _do_ want underlines in this example, where ClickableTexts appear inline:

Screen Shot 2023-01-31 at 4 13 08 PM

src/Nri/Ui/ClickableText/V3.elm
charbelrami: for removing border when text is hidden https://noredink.slack.com/archives/C02NMHB1K55/p1674230782897309?thread_ts=1674139035.616359&cid=C02NMHB1K55

charbelrami:
> I think we _do_ want underlines in this example, where ClickableTexts appear inline:
> 
> Screen Shot 2023-01-31 at 4 13 08 PM

That's because it's not markdown, it's html. I'm only applying underline when it's markdown. Is that what is expected? (https://noredink.slack.com/archives/C02NMHB1K55/p1674508156568749?thread_ts=1674490873.796319&cid=C02NMHB1K55)

tesk9:
Ahh right! Okay I definitely misled you before, sorry.

I think the key rule is: anytime anytime a link or button ClickableText appears _inline_ with other content, we want a bottom border. So we wouldn't want a bottom border to be auto-applied to Container's HTML children, but we _would_ want the bottom border for a Text's HTML children, if that makes sense.

Maybe the best approach is to add an attribute to ClickableText for "withUnderline" or "appearsInline" or something like that, and ask devs to use it when links or buttons appear inline?

charbelrami:
> Ahh right! Okay I definitely misled you before, sorry.
> 
> I think the key rule is: anytime anytime a link or button ClickableText appears _inline_ with other content, we want a bottom border. So we wouldn't want a bottom border to be auto-applied to Container's HTML children, but we _would_ want the bottom border for a Text's HTML children, if that makes sense.
> 
> Maybe the best approach is to add an attribute to ClickableText for "withUnderline" or "appearsInline" or something like that, and ask devs to use it when links or buttons appear inline?

that makes sense! Thank you!

src/Nri/Ui/Container/V2.elm
```diff
@@ -356,5 +357,10 @@ to `p` tags by user agents.
...
 -}
 markdown : String -> Attribute msg
-markdown =
-    Attribute << Content.markdown
+markdown content =
+    Attribute <|
+        \config ->
+            { config
+                | content = Content.markdownContent content
+                , css = config.css ++ MarkdownStyles.anchorAndButton
```
tesk9, line 365: Could we flip the order of `config.css ++ MarkdownStyles.anchorAndButton` to `MarkdownStyles.anchorAndButton ++ config.css` ? So that the custom styles that the user passes in are always after the default markdown styles, and they can be overridden easily if necessary?

tesk9: (Here and in the other component uses?)

tesk9 (commented):
This is looking good!

I really like the appearsInline attribute in particular -- nice & clear! Could you please add it to the settings section on ClickableText so folks can experiment with it/won't assume that the underline will appear automatically? Thanks!

tesk9 (approved):
Woo thank you!

PR #1243, opened 2023-01-20
Allow for any example to support any # of external links, and scootch block guidance up

tesk9:
## Before

Screen Shot 2023-01-20 at 4 26 27 PM

## After

Screen Shot 2023-01-20 at 4 26 13 PM

Note that I also tried to make the destination of "Docs" and "Source" more clear. I think it's less clean, but hopefully more usable?

cc @NoRedInk/design

### DIFF
--- a/styleguide-app/Example.elm
+++ b/styleguide-app/Example.elm
@@ -25,6 +25,7 @@ type alias Example state msg =
     , view : EllieLink.Config -> state -> List (Html msg)
     , categories : List Category
     , keyboardSupport : List KeyboardSupport
+    , extraResources : List ( String, String )
     }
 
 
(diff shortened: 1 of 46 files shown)


bendansby:
Hmm, those long labels are pretty hard to read. I feel like have a standardized set of succinct labels across components would be better for users to know what to click to get more info of a certain kind.

staceyadams:
Fully agree but am having a hard time coming up with something succinct.

Intended audience: LX project team, including non-eng like product and curriculum folks
Content: it's all this additional stuff about how the block works and states you may end up in and would probably have questions about. It also serves as a place to store decisions for why something is the way it is because that question will inevitably come up of "why can't we do x instead"?

"Paper doc"? It's what it is literally but not very descriptive for usage

tesk9:
Maybe "About"?

PR #1249, opened 2023-01-27
Add tts capabilities to question box - phx-465

ArthurJordao:
This adds a new option to the question box to allow the user to add a text-to-speech button inside the component. This is an arbitrary HTML that will be presented to the user on the left side.
This will be shown as the following:
image

### DIFF
--- a/src/Nri/Ui/QuestionBox/V3.elm
+++ b/src/Nri/Ui/QuestionBox/V3.elm
@@ -321,10 +321,19 @@ viewGuidance config referencingId markdown_ =
                 ]
 
         Nothing ->
-            viewSpeechBubble config
-                referencingId
-                [ Balloon.markdown markdown_
-                , Balloon.css [ Css.margin2 (Css.px 10) (Css.px 20) ]
+            div
+                [ css
+                    [ Css.displayFlex
+                    , Css.justifyContent Css.flexEnd
+                    , Css.position Css.relative
+                    ]
+                ]
+                [ config.leftButton
+                , viewSpeechBubble config
+                    referencingId
+                    [ Balloon.markdown markdown_
+                    , Balloon.css [ Css.margin2 (Css.px 10) (Css.px 20) ]
+                    ]
                 ]
 
 
--- a/src/Nri/Ui/QuestionBox/V3.elm
+++ b/src/Nri/Ui/QuestionBox/V3.elm
@@ -281,7 +281,13 @@ viewBalloon config referencingId attributes =
     Balloon.view
         ([ Balloon.html
             (List.filterMap identity
-                [ Maybe.map (viewGuidance config referencingId) config.markdown
+                [ Just <|
+                    div [ css [ Css.displayFlex ] ]
+                        (List.filterMap identity
+                            [ Just config.leftButton
+                            , Maybe.map (viewGuidance config referencingId) config.markdown
+                            ]
+                        )
                 , viewActions config.character config.actions
                 ]
             )
@@ -293,7 +299,7 @@ viewBalloon config referencingId attributes =
 
 
 viewGuidance :
-    { config | id : Maybe String, character : Maybe { name : String, icon : Svg }, leftButton : Html msg }
+    { config | id : Maybe String, character : Maybe { name : String, icon : Svg } }
     -> Maybe String
     -> String
     -> Html msg
@@ -309,8 +315,7 @@ viewGuidance config referencingId markdown_ =
                     , Css.position Css.relative
                     ]
                 ]
-                [ config.leftButton
-                , viewCharacter character_
+                [ viewCharacter character_
                 , viewSpeechBubble config
                     referencingId
                     [ Balloon.markdown markdown_
@@ -321,19 +326,10 @@ viewGuidance config referencingId markdown_ =
                 ]
 
         Nothing ->
-            div
-                [ css
-                    [ Css.displayFlex
-                    , Css.justifyContent Css.flexEnd
-                    , Css.position Css.relative
-                    ]
-                ]
-                [ config.leftButton
-                , viewSpeechBubble config
-                    referencingId
-                    [ Balloon.markdown markdown_
-                    , Balloon.css [ Css.margin2 (Css.px 10) (Css.px 20) ]
-                    ]
+            viewSpeechBubble config
+                referencingId
+                [ Balloon.markdown markdown_
+                , Balloon.css [ Css.margin2 (Css.px 10) (Css.px 20) ]
                 ]
 
 
--- a/src/Nri/Ui/QuestionBox/V3.elm
+++ b/src/Nri/Ui/QuestionBox/V3.elm
@@ -3,8 +3,8 @@ module Nri.Ui.QuestionBox.V3 exposing
     , id, markdown, actions, character
     , standalone, pointingTo
     , containerCss
-    , guidanceId
     , setTextToSpeechView
+    , guidanceId
     )
 
 {-|
@@ -14,6 +14,7 @@ module Nri.Ui.QuestionBox.V3 exposing
 @docs id, markdown, actions, character
 @docs standalone, pointingTo
 @docs containerCss
+@docs setTextToSpeechView
 
 @docs guidanceId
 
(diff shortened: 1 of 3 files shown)


src/Nri/Ui/QuestionBox/V3.elm
```diff
@@ -298,15 +311,37 @@ viewGuidance config referencingId markdown_ =
...
+                                , withMedia [ quizEngineMobile ]
+                                    [ Css.left Css.auto
+                                    , Css.top Css.auto
+                                    , Css.float Css.left
+                                    , Css.padding4 Css.zero (Css.px 5) Css.zero Css.zero
+                                    , Css.position Css.static
+                                    ]
+                                ]
+                            ]
+                            << List.singleton
```
tesk9, line 331: I don't think that the positioning of the tts view should happen from within questionbox. The ReadAloud code is set up to handle positioning (it already has position relative and zindex -- I think in this case, you would override it with position absolute and use top: 8px and left: -24px). Having positioning logic in noredink-ui seems more brittle to me, since if we, for instance, decide we want the readaloud icons to be much bigger, we would then have to do an entire noredink-ui release cycle before we could make the change.

(Why position absolute? I'm noticing that the position: relative is pushing the guidance too far right. Compare the blue space on the left side of the guidance:

Screen Shot 2023-01-27 at 1 39 25 PM
Screen Shot 2023-01-27 at 1 39 29 PM

It's totally possible that the blue space should increase a bit, based on the figma files I've last seen, but if so, I would guess it should be balanced with the character on the right.

But the character on the right and its positioning is subject to/likely to change! So again, I think having the position styles elsewhere is a good call.
)

src/Nri/Ui/QuestionBox/V3.elm
```diff
@@ -298,15 +311,37 @@ viewGuidance config referencingId markdown_ =
...
+                 , Just <|
+                    viewSpeechBubble config
+                        referencingId
+                        [ Balloon.markdown markdown_
+                        , Balloon.onLeft
+                        , Balloon.alignArrowEnd
+                        , Balloon.css [ Css.minHeight (Css.px 46) ]
+                        ]
+                 ]
+                    |> List.filterMap identity
```
tesk9, line 343: I think changing the modeling from `textToSpeechView : Maybe (Html msg)` to `textToSpeechView: Html msg`, initialized as `textToSpeechView = text ""` might simplify the view code a bit.

Currently, there's actually a bug in that code that I think becomes a bit more obvious if there's no need for the List.filterMap logic:

https://user-images.githubusercontent.com/8811312/215192259-fd3bc83e-a331-4abf-ad53-b7931b9af477.mov

src/Nri/Ui/QuestionBox/V3.elm
```diff
@@ -3,6 +3,7 @@ module Nri.Ui.QuestionBox.V3 exposing
     , id, markdown, actions, character
     , standalone, pointingTo
     , containerCss
+    , setTextToSpeechView
```
tesk9, line 6: I wonder if this name should be more generic? It's definitely true that the plan is to use this for TTS, but if it accepts arbitrary HTML, then we're not requiring that "the plan" be followed 😅

tesk9 (commented):
Thank you for extending the question box example! Super helpful. I do have a couple of recommendations, left in individual comments.

src/Nri/Ui/QuestionBox/V3.elm
ArthurJordao: Nice, I was wondering if this code should be at NRI UI or not, but I think it makes more sense to be outside NRI UI.

src/Nri/Ui/QuestionBox/V3.elm
ArthurJordao: yeah, I think so!

ArthurJordao:
Hey @tesk9 I did some changes to the API and moved the CSS on the example only.

tesk9:
Thanks! I think that the 🐛 is still there, though, so I pushed a failing test for it.

ArthurJordao:
@tesk9 wow I forgot to check the bug, thank you!

src/Nri/Ui/QuestionBox/V3.elm
```diff
@@ -309,10 +321,19 @@ viewGuidance config referencingId markdown_ =
...
-                referencingId
-                [ Balloon.markdown markdown_
-                , Balloon.css [ Css.margin2 (Css.px 10) (Css.px 20) ]
+            div
+                [ css
+                    [ Css.displayFlex
+                    , Css.justifyContent Css.flexEnd
+                    , Css.position Css.relative
+                    ]
+                ]
```
tesk9, line 330: How would you feel about adding the new content in higher in the DOM instead of down in viewGuidance? I'm thinking maybe [here](https://github.com/NoRedInk/noredink-ui/blob/master/src/Nri/Ui/QuestionBox/V3.elm#L272), in `viewBalloon`? I think that way, no new wrapping position-relative divs need to be added.

tesk9 (commented):
I'm still not 100% sure about the API (TTS isn't always rendered as 1 button -- sometimes it is two buttons stacked vertically with a slide animation, for instance) but I'll leave that up to you & Phoenix. I had 1 small further suggestion which you're welcome to ignore. Feel free to merge once the other folks have reviewed!

src/Nri/Ui/QuestionBox/V3.elm
ArthurJordao: Makes sense @tesk9, just did it.

ArthurJordao:
> I'm still not 100% sure about the API (TTS isn't always rendered as 1 button -- sometimes it is two buttons stacked vertically with a slide animation, for instance) but I'll leave that up to you & Phoenix. I had 1 small further suggestion which you're welcome to ignore. Feel free to merge once the other folks have reviewed!

If the two buttons appear together couldn't it be inside the same div and we pass in the same property? I never saw this 2 buttons variations, so I am not aware of this case 😬

tesk9:
Yeah totally! The types totally work as-is, I'm just wondering about the name "left button".


https://user-images.githubusercontent.com/8811312/215819724-c8535c36-b99c-4d15-9a73-a5acac105263.mov

See how there's a play button, a pause button, and a stop button plus their respective tooltips? It's more than just a singular button.

This is not a big deal though. I don't feel strongly about it at all

src/Nri/Ui/QuestionBox/V3.elm
```diff
@@ -270,7 +281,13 @@ viewBalloon config referencingId attributes =
     Balloon.view
         ([ Balloon.html
             (List.filterMap identity
-                [ Maybe.map (viewGuidance config referencingId) config.markdown
+                [ Just <|
+                    div [ css [ Css.displayFlex ] ]
```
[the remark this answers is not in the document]
tesk9: Hmm, but with the div, and with display flex, extra blank space gets added to the left side of the view. And I kinda think it's more than there should be? It's more than is on the right side of the question box for the character.

Screen Shot 2023-01-31 at 9 25 40 AM
Screen Shot 2023-01-31 at 9 25 45 AM

--

I think if the TTS stuff is positioned absolutely, it will be relative to the whole question box, so shouldn't matter whether there is a div here or not.

If I change the UiIcon.play icon positioning to absolute, and the left-button rendering back to:

```
(List.filterMap identity
                [ Just config.leftButton
                , Maybe.map (viewGuidance config referencingId) config.markdown
                , viewActions config.character config.actions
                ]
            )
```

it looks like this:

Screen Shot 2023-01-31 at 9 30 18 AM

Which is probably not enough whitespace on the left side! Maybe worth talking to Stacey about what the desired spacing is -- I could see wanting more left spacing iff TTS is present and I could see wanting more left spacing whether TTS is present or not.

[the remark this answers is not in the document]
tesk9: Again, though, I don't want to be a blocker on this. If Stacey says the current behavior is fine, that's fine with me.

ArthurJordao:
> Yeah totally! The types totally work as-is, I'm just wondering about the name "left button".

ooh, that makes sense, maybe we can rename it to `leftActions`? Wdyt?

ArthurJordao:
Yeah totally! The types totally work as-is, I'm just wondering about the name "left button".

ooh, that makes sense @tesk9 , maybe we can rename it to leftActions? Wdyt?

tesk9:
"leftActions" sounds good to me! thanks!

ArthurJordao:
nice @tesk9 ! Can you help me later with approving/creating a release for the nri-ui?

tesk9:
Sure! Do you want to wait to merge on getting approval from @micahhahn and/or @nunocf since they're tagged, and from @NoRedInk/design ?

micahhahn (approved):
Looks good to me! Simple and straightforward.  Nice work Arthur!

PR #1254, opened 2023-01-31
[PHX-485] Support markdown in the Block API

micahhahn:
This PR adds support for markdown to the Block API in two places: block labels and block contents. 

For block labels we simply start interpreting the label string as markdown using existing the existing `markdown` feature of the Ballon API. 

For block contents I've added two new functions `bold` and `italic` that work only on the `Content` type.  This should give content-creation enough flexibility to handle all cases of markdown in question segments. 

## Context

Fixes [PHX-485](https://linear.app/noredink/issue/PHX-485/support-markdown-in-block-api)

image

        - an accurate preview
        - valid sample code
        - correct keyboard behavior


src/Nri/Ui/Block/V4.elm
```diff
@@ -304,6 +329,20 @@ blankWithId =
...
+    Markdown Bold
+
+
+{-| Wraps a group of content in a `em` tag
+-}
+italic : List (Content msg) -> Content msg
+italic =
+    Markdown Italic
+
+
```
micahhahn, line 345: I thought perhaps that `strong` and `emphasis` might be better names semantically, but was worried about `emphasis` causing confusion with how we also us emphasis to describe display elements.

src/Content.elm
```diff
@@ -51,6 +53,22 @@ markdownContent content =
...
+            -- The library always parses into a paragraph and never into `PlainInline`
+            List.map (Markdown.Inline.toHtml >> fromUnstyled) inlines
+                |> p [ css [ Css.margin Css.zero ] ]
+                |> List.singleton
+
+        _ ->
+            Markdown.toHtml Nothing content
+                |> List.map fromUnstyled
+
+
```
micahhahn, line 71: I considered simply changing `markdownContent`, but there are enough callers I was scared of breaking other places.

src/Nri/Ui/Mark/V2.elm
micahhahn, line 1: I believe I updated all places where we should call `stripMarkdownSyntax`, but please give it a close look to make sure I didn't miss something!

micahhahn:
@tesk9 I noticed that `msg` is essentially a phantom type of `Content` now that we have pulled the QB stuff out of `Content`.  I'm guessing we were holding off on removing `msg` until we do a major version bump to `V5`?

tesk9 (commented):
This is looking great! Thanks for updating the examples and for structuring your PR for a super-straightforward review. 🙏 

---

The only thing I'd like to request is some tests around the "stripMarkdownSyntax" behavior (not the function itself -- just that Block and Highlighter's `::before` and `::after` pseudoelements end up with the right content, and that the sometimes-visually-displayed start tag renders correctly _as markdown_ for highlighter as well.)

Charbel added super useful helpers for testing ::before and ::after content [here](https://github.com/NoRedInk/noredink-ui/blob/master/tests/Spec/Nri/Ui/Highlighter.elm#L352) through line 430. I think pulling these helpers out somewhere for use in the Block specs would be great!

Could you also please add a patch note about the change to Mark for markdown support? Thanks!

---

>  I noticed that msg is essentially a phantom type of Content now that we have pulled the QB stuff out of Content. I'm guessing we were holding off on removing msg until we do a major version bump to V5?

I think this is a mistake that nobody had noticed yet! Or at least, this isn't something that I was keeping track of. I think it's just an oversight. But yeah I think waiting for V5 probably makes sense.

micahhahn:
@tesk9 I've added a patch changelog message to Mark and added a couple unit tests - let me know if those aren't comprehensive enough.  (I'm not quite sure I know what you mean by "sometimes-visually-displayed" start tag)

micahhahn:
@tesk9 Thank you for the unit test examples! I've merged them in and fixed them along with some other changes we talked about.

src/Content.elm
```diff
@@ -51,6 +53,20 @@ markdownContent content =
         |> List.map fromUnstyled
 
 
+{-| Provide a string that will be rendered as markdown and placed in a paragraph tag, with the default margin removed.
+-}
+markdownInline : String -> List (Html msg)
```
tesk9, line 58: Love the rename!

src/Content.elm
```diff
@@ -51,6 +53,20 @@ markdownContent content =
         |> List.map fromUnstyled
 
 
+{-| Provide a string that will be rendered as markdown and placed in a paragraph tag, with the default margin removed.
```
tesk9, line 56:
```suggestion
{-| Provide a string that will be rendered as inline markdown, with the default wrapping paragraph removed.
```

tesk9 (approved):
Thanks! Looks great!

Issue #1253, opened 2023-01-31
Nri.Ui.Button markdown rendering for labels is broken.


micahhahn:
If I pass a label of "Markdown **here**" to the Button API, the words come out squashed together: 

I'm 99% confident this is due to us setting `display: inline-flex` on the text container.  When I take that style away the text looks as expected 🙃.  Of course I'm not sure why that style is there in the first place - Chesterton's fence and all that. 

https://user-images.githubusercontent.com/12899207/215862822-a8755987-cc6e-4835-8572-d23b1a73905e.mov

micahhahn:
@tesk9 Perhaps we can talk about this in our pairing today as well - I think it's likely curriculum will run into this bug.

PR #1255, opened 2023-01-31
Fix markdown parsing on button

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/1253

Markdown string used for both screenshots: "Label **bold**   *emphasis*"

### Before

Screen Shot 2023-01-31 at 3 56 27 PM

### After
Screen Shot 2023-01-31 at 3 55 31 PM

--

cc @NoRedInk/design I've changed the default label text for buttons so it's clear there's a markdown renderer involved. TBH, I was totally surprised by the issue here, because I thought the string was plaintext! LMK if you want the default to go back to plaintext. Happy to change it.

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -826,10 +826,10 @@ viewLabel config =
         []
         (case config.icon of
             Nothing ->
-                renderMarkdown config.label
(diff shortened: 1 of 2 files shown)


bendansby:
Today We All Learned

micahhahn (approved):
Thanks for fixing this Tessa!

PR #1258, opened 2023-02-03
Add datetime input options

lindsaykwardell:
Adds options for `date` and `datetime` to TextInput. Both of these expect `Maybe Time.Posix` as their values. Examples are available in the styleguide.


tesk9 (commented):
Looks great! If you have time, I'd love to see some tests (largely to explain/enforce the necessity of dropRight and the String.slice) against the inputs. But no worries if not -- I am very excited about this improvement!!!! 🙏

lindsaykwardell:
> Looks great! If you have time, I'd love to see some tests (largely to explain/enforce the necessity of dropRight and the String.slice) against the inputs. But no worries if not -- I am very excited about this improvement!!!! 🙏

Tests and better description for the two new functions have been added. I just did a couple basic tests for each one to ensure the date formats in the value will be checked. Let me know if you think there should be more.

tesk9:
Looks good to me! thanks 🙏

erik-nri (approved):
Awesome!!!

PR #1268, opened 2023-02-13
More involved review suggestions for QuestionBox V4

tesk9:
Before, it was difficult to tell visually what the question box was being positioned against. My hope is adding the Container wrapper and an example with the question box typically positioned at the right of the content will make the behavior more clear.

## Before

localhost_8000_ (7)

## After

localhost_8000_ (6)


ArthurJordao (approved):

ArthurJordao:
Hey thank you @tesk9 for the suggestion I just merged it to my PR

PR #1271, opened 2023-02-13
Adds a little highlight toolbar preview

tesk9:
## Context

Fixes A11-2300

Screen Shot 2023-02-13 at 3 14 15 PM
cc @NoRedInk/design a lil highlighter toolbar preview! Much simplified -- not sure if you have ideas about how to include the label text in the preview, or if you think this is sufficient for easy browsing through the catalog?


bendansby:
Could make it vertical with text labels, which would also fill out the space a bit more?

tesk9:
Oh yeah I think that's better!

image

Thanks for the suggestion!

bendansby:
Nice 👍

charbelrami (approved):
Looks good!

PR #1277, opened 2023-02-14
:wrench: Menu tooltips

tesk9:
## Context

It turns out that our tooltip and our menu component don't play nicely together: either the Menu keyboard events work or the Tooltip keyboard events work.

The API previous to this restricted the behavior such that only the Tooltip keyboard events worked. This is not a state we can work from, since it fully blocks users (keyboard-only users just _won't_ be able to get to the menu contents).

This PR adds first-class tooltip support to Menu, which allows us to choose the Menu keyboard behavior over the Tooltip keyboard behavior. This also significantly improves the user experience of adding a tooltip to a Menu.

Note that the tooltip experience won't be perfect: for example, while focused on the menu's triggering button, hitting escape will _not_ close the tooltip (which is a WCAG failure).

Screen Shot 2023-02-14 at 4 39 19 PM

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog-app/Examples/Menu.elm
+++ b/component-catalog-app/Examples/Menu.elm
@@ -139,11 +139,12 @@ view ellieLinkConfig state =
         , version = version
         , update = UpdateControls
         , settings = state.settings
-        , mainType = Just "RootHtml.Html { focus : Maybe String, isOpen : Bool }"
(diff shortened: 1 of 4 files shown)


charbelrami (approved):
thank you!

PR #1278, opened 2023-02-15
A11-2301 add tests for HighlighterToolbar

charbelrami:
This adds tests for HighlighterToolbar, which ensure the correct implementation of the aria-pressed attribute and verify that the visual indicator only appears when necessary.


tesk9 (approved):
Lovely, thank you!

PR #1279, opened 2023-02-15
Don't prefix `select option[value]` with "nri-select-"

caseyWebb:

## Context

`Nri.Ui.Select` currently prefixes option `value` attributes with `nri-select-`; this is undesirable for the use-case of creating a form that posts to an endpoint we don't control. For example, the premium application on https://noredink.com/premium posts directly to Pardot (the action is set to a form handler created in Pardot and `POST`'d on submit). This causes the values to be sent to Pardot as "nri-select-my-cool-option" instead of "My Cool Option", which in turn makes Salesforce data unusable.

This change includes a version bump to Select even though the public API does not change because it changes the testing behavior (many automated tests looking for options with "nri-select-").

This change was discussed with @tesk9 on Slack

N/A

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/forbidden-imports.toml
+++ b/forbidden-imports.toml
@@ -153,6 +153,9 @@ hint = 'upgrade to V8'
 [forbidden."Nri.Ui.Select.V7"]
 hint = 'upgrade to V8'
 
+[forbidden."Nri.Ui.Select.V8"]
+hint = 'upgrade to V9'
+
 [forbidden."Nri.Ui.SideNav.V1"]
 hint = 'upgrade to V3'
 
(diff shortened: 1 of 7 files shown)


forbidden-imports.toml
```diff
@@ -153,6 +153,9 @@ hint = 'upgrade to V8'
 [forbidden."Nri.Ui.Select.V7"]
 hint = 'upgrade to V8'
 
+[forbidden."Nri.Ui.Select.V8"]
+hint = 'upgrade to V9'
```
ap-nri, line 157: I can't find anything in the readme about how this file works, but should all the forbidden versions of `Nri.Ui.Select` now hint 'upgrade to V9' ?

PR #1280, opened 2023-02-15
Fix missing whitespace for Highlighter.staticMarkdown with Highlightable.formMarkdown

tesk9:
## Context

While working on A11-1825, I noticed that when we switched to using the highlighter view, the whitespace around the mark element was lost. This is because Highlighter.staticMarkdown uses a markdown implementation that (reasonably) trims whitespace. However, we definitely want to keep all our whitespace within the context of a paragraph!

So this PR changes the highlighter implementation such that when we're rendering as markdown, we trim whitespace and then re-add it as to the rendered markdown as separate text nodes.

### Before

Screen Shot 2023-02-15 at 2 58 08 PM

### After

Screen Shot 2023-02-15 at 2 58 23 PM

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog-app/Examples/Highlighter.elm
+++ b/component-catalog-app/Examples/Highlighter.elm
@@ -123,6 +123,13 @@ example =
                     , sort = Nothing
                     }
                 , Table.string
+                    { header = "Highlightable."
+                    , value = .highlightable
+                    , width = Css.pct 10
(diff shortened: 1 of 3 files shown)


charbelrami (approved):
Looks good!

PR #1281, opened 2023-02-15
Select test setup

tesk9:
Hey @caseyWebb ! I saw your PR (https://github.com/NoRedInk/noredink-ui/pull/1279) cause you tagged me, and I think there's an issue with the decoder lookup.

```
valueLookup =
            optionStringChoices
                ++ groupStringChoices
                |> List.map (\x -> ( x.id, x.value ))
                |> Dict.fromList

        decodeValue string =
            Dict.get string valueLookup
                |> Maybe.map Json.Decode.succeed
                |> Maybe.withDefault
                    -- At present, elm/virtual-dom throws this failure away.
                    (Json.Decode.fail
                        ("Nri.Select: could not decode the value: "
                            ++ string
                            ++ "\nexpected one of: "
                            ++ String.join ", " (Dict.keys valueLookup)
                        )
                    )

        onSelectHandler =
            Events.on "change" (Events.targetValue |> Json.Decode.andThen decodeValue)
```

I think we're using the look-up table as `Dict Id Value`, but the value that we decode is actually the event's target value, which I think is the `value`.

I wrote a test that hopefully explains what I'm seeing.

### DIFF
--- a/component-catalog-app/Examples/Select.elm
+++ b/component-catalog-app/Examples/Select.elm
@@ -18,7 +18,7 @@ import Example exposing (Example)
 import Html.Styled
 import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Heading.V3 as Heading
-import Nri.Ui.Select.V9 as Select exposing (Choice)
+import Nri.Ui.Select.V9 as Select
 import Nri.Ui.Text.V6 as Text
 
 
@@ -363,16 +363,9 @@ weightLifterLabel =
     "81 kg 2020 Olympic Weightlifters"
 
 
-toOption : Choosable -> { label : String, value : Choosable }
+toOption : Choosable -> Select.Choice Choosable
 toOption c =
     { label = choosableToLabel c, value = c }
-
-
-grouped : List (Select.ChoicesGroup Choosable)
-grouped =
-    [ { label = texMexLabel, choices = List.map toOption allTexMex }
-    , { label = weightLifterLabel, choices = List.map toOption all81kg2020OlympicWeightlifters }
-    ]
 
 
 initChoices : Control ( String, Select.Attribute Choosable )
@@ -382,7 +375,7 @@ initChoices =
         toOptionString c =
             "{ value = " ++ choosableToCodeString c ++ ", label = \"" ++ choosableToLabel c ++ "\" } "
 
-        toChoice : List Choosable -> ( String, List { label : String, value : Choosable } )
+        toChoice : List Choosable -> ( String, List (Select.Choice Choosable) )
         toChoice choosables =
             ( """Select.choices
         choosableToLabel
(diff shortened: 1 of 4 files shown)


ap-nri:
@tesk9 fixed decoder, made the value of the Select in the example controlled and added a note to the example about being careful with the values. Let me know if it's doing too much/the most.

component-catalog-app/Examples/Select.elm
```diff
@@ -82,38 +83,58 @@ example =
...
+                    Note that if the value is bound (and why would you ever make a Select where it isn't?)
+                    then changing the list of options will not change its value.
+                    Furthermore, Select will only fire an event when a new, non-default value is selected.
+                    The current value is
+                """
+                        ++ getABetterNameToString state.selectedValue
+                        ++ "."
+                        ++ """
+                   if you change the "choices" above to "Overflowing text choices" there is no way to change the current value
+                   to Zaphod (the value of the single option with a very long name).
```
tesk9, line 99: Maybe it's worth changing the value/overflowing-text pair to be more obviously related to each other? Maybe change the label to be the full verse of the sonnet, and then have the value be something corresponding? Like `Shakespeare` or something? or something else entirely that is related?

component-catalog-app/Examples/Select.elm
```diff
@@ -82,38 +83,58 @@ example =
...
+                |> Html.Styled.map ChangedTheSelectorValue
+            , Text.mediumBody
+                [ Text.plaintext <|
+                    """
+                    Note that if the value is bound (and why would you ever make a Select where it isn't?)
+                    then changing the list of options will not change its value.
+                    Furthermore, Select will only fire an event when a new, non-default value is selected.
+                    The current value is
+                """
+                        ++ getABetterNameToString state.selectedValue
```
tesk9, line 95: Super useful to have this listed out explicitly! thanks

component-catalog-app/Examples/Select.elm
tesk9: Or maybe... I think the 🐛 here that prevents the overflowing text from being selectable is in the modeling: having 2 discrete lists of options that aren't displayed at the same time, but are actually all part of the same union type.

Perhaps instead it would make sense to remove the option to use "short choices" or not and instead always include a long long long piece of text in the singular option list. Does that make sense?

src/Nri/Ui/Select/V9.elm
```diff
@@ -342,7 +342,7 @@ viewSelect config_ config =
         valueLookup =
             optionStringChoices
                 ++ groupStringChoices
-                |> List.map (\x -> ( x.id, x.value ))
+                |> List.map (\x -> ( x.strValue, x.value ))
```
tesk9, line 345: I'm noticing `groupStringChoices` looking at this code again.. Maybe we should add another test for when we group the options?

tesk9 (commented):
Looks great! Thanks for fixing!

I can't approve since this is originally my PR, but maybe @caseyWebb should review as well?

tesk9:
Thank you for making all the great Select example improvements! It's much clearer what's happening! 😍

ap-nri:
> Thank you for making all the great Select example improvements! It's much clearer what's happening! 😍

ty. hackday: https://github.com/ghcjs/ghcjs

Issue #1265, opened 2023-02-10
Allow width per breakpoint on Nri.Ui.Menu


bendansby:
And allow px or %

tesk9:
😱 that first screenshot ahhhhhhhhhhhhhhhhhhhhhhhhhhh 🙈

PR #1287, opened 2023-02-21
:wrench: supporting different width behaviors for breakpoints

tesk9:
Fixes A11-2309

## Context

The goal is to be able to specify how the Menu trigger width should behave at different breakpoints, so that we don't have nice-looking desktop views and janky-looking mobile views. See https://github.com/NoRedInk/noredink-ui/issues/1265 for some example screenshots.

Since Menu now integrates directly with Button and the other clickables, I'm addressing this need by adding per-viewport width support to Button. 

I don't think we'll want clickable texts to expand to take up tons of horizontal space, so I'm not making changes the ClickableText component.

I also think the width behavior of clickable svgs is a _bit_ less complicated? And it makes sense to let the user use mobileCss directly. Example of doing this is here: https://github.com/NoRedInk/NoRedInk/pull/42777.

Additionally, I wanted the width behavior to be easy to test and observe, so I split out an interactive example from the Button table, and made the Button table entirely non-interactive.

### Width behavior by breakpoint demo

https://user-images.githubusercontent.com/8811312/220441598-81b384b2-7534-412e-bc84-3f9512d822fe.mov

### Examples Table Before

image

### Examples Table After

image

cc @NoRedInk/design 

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog-app/Examples/Button.elm
+++ b/component-catalog-app/Examples/Button.elm
@@ -170,6 +170,15 @@ initDebugControls =
                         , ( "fillContainerWidth", Button.fillContainerWidth )
                         ]
                     )
+                |> ControlExtra.optionalListItem "mobile width"
+                    (CommonControls.choice moduleName
+                        [ ( "exactWidthForMobile 120", Button.exactWidthForMobile 120 )
+                        , ( "exactWidthForMobile 70", Button.exactWidthForMobile 70 )
+                        , ( "boundedWidthForMobile 100 180", Button.boundedWidthForMobile { min = 100, max = 180 } )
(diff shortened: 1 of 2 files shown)


bendansby:
What if we want to use defaulTrigger for Menus, which we generally will?

tesk9:
`defaultTrigger` uses `Button` under the hood, so you can pass a list of `Button` attributes when you make it!

bendansby:
Awesome, thanks for doing this!

charbelrami (approved):
Looks great!

Issue #1264, opened 2023-02-10
Add optional icon for Nri.Ui.Select


bendansby:
[Linear issue
](https://linear.app/noredink/issue/A11-2310/support-icon-for-nriuiselect)
Just like Nri.Ui.Menu

tesk9:
Example use: course switcher.

PR #1289, opened 2023-02-22
Support icon in Select

tesk9:
## Context

Fixes A11-2310
Fixes https://github.com/NoRedInk/noredink-ui/issues/1264

Screen Shot 2023-02-22 at 2 55 31 PM

Screen Shot 2023-02-22 at 2 55 02 PM

Screen Shot 2023-02-22 at 2 55 13 PM

cc @NoRedInk/design 

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog-app/Examples/Select.elm
+++ b/component-catalog-app/Examples/Select.elm
@@ -164,6 +164,7 @@ initControls =
             (Control.value ( "Select.disabled", Select.disabled ))
         |> ControlExtra.optionalListItem "loading"
             (Control.value ( "Select.loading", Select.loading ))
+        |> CommonControls.icon moduleName Select.icon
 
 
 initChoices : Control ( String, List (Choice String) )
--- a/component-catalog-app/Examples/Select.elm
+++ b/component-catalog-app/Examples/Select.elm
@@ -28,7 +28,7 @@ moduleName =
 
 version : Int
 version =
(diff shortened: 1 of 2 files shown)


bendansby:
Ahh I love it, thank you! I was thinking about asking to make the icon azure, but first I double checked what we have for Menu, which in noredink-ui is the same as here (current color/gray20), but in the monolith is azure.

image

image
image

Do you know why the discrepancy?

tesk9:
It looks like in the monolith, we're explicitly setting the icon color to azure for each menu trigger.

I don't remember why, in Menu.V4, I didn't make the icon color automatically azure -- maybe I was worried about it being hard to override? or something? I don't know. Or maybe I thought there was some case where it wasn't supposed to be azure? Or maybe it's just the text color is supposed to be black, and so it's tricky API-wise to get the svg to be azure? That's probably it -- tricky, from the API standpoint, to get the text color to be black and the icon color to be azure, since that's not a unique button theme.

Clearly, I should address that as a problem though, since I confused myself on this PR and thought the icon was _supposed_ to be black!

tesk9:
image

Azure looks nicer!

bendansby:
Great, let's go with that!

charbelrami (approved):
Looks good!

PR #1294, opened 2023-02-27
A bit of reorganization

bendansby:
- Changes "Interactions" category to "Assignments" and restrict those components to that category
- Moves a few components to new category "Navigation" out of "Layout"


tesk9 (approved):
Thanks!

PR #1299, opened 2023-02-27
Adjust question box style to conform the design

ArthurJordao:
# :wrench: Modifying a component

## Context

Requested changes from design, that I forgot to implement on v4. [More context here](https://linear.app/noredink/issue/PHX-481#comment-82b336e6)
## :framed_picture: What does this change look like?
image

- Follow the the [Guidelines for an optimal design ping](https://paper.dropbox.com/doc/Guidelines-for-Sharing-User-Facing-Changes-with-Design--BpL8hpJLMugy6033aT5m0JdaAg-bdKGQtYH9qO9I00hUkA6k) to add helpful screenshots/videos for design.
- Ping design on this PR to get approval on your changes.

## Component completion checklist

- [x] I've gone through the relevant sections of the [Development Accessibility guide](https://paper.dropbox.com/doc/Accessibility-guide-4-Development--BiIVdijSaoijjOuhz3iTCJJ1Ag-rGoHpC91pFg3zTrYpvOCQ) with the changes I made to this component in mind
- [x] Changes are clearly documented
    - [x] Component docs include a changelog
    - [x] Any new exposed functions or properties have docs
- [x] Changes extend to the Component Catalog
    - [x] The Component Catalog is updated to use the newest version, if appropriate
    - [x] The Component Catalog example version number is updated, if appropriate
    - [x] Any new customizations are available from the Component Catalog
    - [x] The component example still has:
        - an accurate preview
        - valid sample code
        - correct keyboard behavior
        - correct and comprehensive guidance around how to use the component
- [x] Changes to the component are tested/the new version of the component is tested
- [x] Component API follows standard patterns in noredink-ui
    - e.g., as a dev, I can conveniently add an `nriDescription`
    - and adding a new feature to the component will _not_ require major API changes to the comopnent

PR #1300, opened 2023-02-28
Update pull_request_template.md

krinhorn:
## What is this?
Trying to avoid [this problem](https://github.com/NoRedInk/noredink-ui/pull/1299#pullrequestreview-1316541699) in the future or at least try to make folks feel accountable by ensuring actions are checkboxes instead of bullets.

Also clarified the process for "pinging" design and added expectations about what to expect _from_ design.

## What should the reviewer check?
Don't bother looking at the commits - just look at the file changes.

Does it matter if design is mentioned in the PR body vs a comment? For instance, if a dev creates a draft PR and then later updates the PR body to @ mention design, does design get a notification?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -36,8 +36,8 @@ Please link to any relevant context and stories.
 
 ## :framed_picture: What does this change look like?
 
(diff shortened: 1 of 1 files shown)


tesk9 (approved):
Thanks for this change! I think this will help 🙏 

> Does it matter if design is mentioned in the PR body vs a comment? For instance, if a dev creates a draft PR and then later updates the PR body to @ mention design, does design get a notification?

I often create a draft PR and then @ mention design later, and they appear on the PR, so I think it works just fine! Comments work the same, afaik.

PR #1307, opened 2023-03-01
More reorg

bendansby:
- Rename Assignments to Instructional (it's not parallel, I know 😭)
- Alphabetize categories in sidebar
- Change "Assorted Components" link label to "Additional Components"

image


tesk9 (approved):
Works for me! Thank you

PR #1315, opened 2023-03-14
Update fieldset styling

bendansby:
- Gets rid of ugly black borders around fieldset groups (which are defined in Menu for some reason?)
- Updates styling of fieldset titles

Fix for https://github.com/NoRedInk/NoRedInk/pull/43111#issuecomment-1468476515

Screenshot 2023-03-14 at 11 03 26 AM


tesk9 (approved):
Looks good, although I think that since it's not super clear that the solo entry _after_ the group isn't in the group, we'll want to avoid isolated entries after groups.

PR #1320, opened 2023-03-15
Updating publish docs

steder:
Including additional documentation on publishing flow around git tag err…ors introduced by branch protection breaking the default flow proposed by elm publish.


steder:
Let me know if this language makes sense.  A better fix might be to try to upstream a change to elm publish that tweaks the commands it presents, but I think that might not get picked up for a while.

tesk9 (approved):
Looks good! Thank you!

Issue #1321, opened 2023-03-17
Updated X icon to be consistent.


ravi-morbia:
X icon now has 5px stroke instead of 7px, to be consistent with other icons in the set. Attached is the svg.

![image](https://user-images.githubusercontent.com/128178614/225997656-132f120a-8d74-436c-92b1-62b08039df92.png)

![x](https://user-images.githubusercontent.com/128178614/226000137-b734e1b1-9450-4156-ba47-3563d4032ac4.svg)

PR #1322, opened 2023-03-17
Change out x icon

tesk9:
Changes out the X icon

Fixes https://github.com/NoRedInk/noredink-ui/issues/1321

### Before

Screen Shot 2023-03-17 at 1 25 57 PM

### After

Screen Shot 2023-03-17 at 1 26 18 PM

## Viewport change

I'm a little worried about this change, to be honest. I think there are places where this might make the x icon too small, e.g., when closing a modal... I'd prefer to get the icon weight consistent without making the icon size smaller wrt the viewport, if possible.

But if you're confident, I'm happy to go with it!

Screen Shot 2023-03-17 at 1 28 03 PM

cc @NoRedInk/design


bendansby:
Yeah, I think the bounding box should be flush with the vector. Ravi, was it not being flush intentional? (going to figure out why I can't @ mention you)

tesk9:
Oh yeah! And the other thing we should consider: do we want any changes to [AnimatedIcon](https://noredink-ui.netlify.app/#/doodad/AnimatedIcon)'s close icon as well? (the weight is very different over there)

ravi-morbia:
Thanks for flagging this concern @tesk9! I think when we redo the icon set as a whole, I would like to maintain equal visual weight (e.g. the x icon and plus icon should be the exact same besides the rotation), but to ensure there are no issues before we have time to consider all of the places in which the x appears, I will update it to take up the whole space.

Attached is the updated icon with each line 5px by 30px. It's not perfectly flush, but I wanted to keep the sizing within our 5px base system (i.e. 30px is the biggest multiple of 5 that still fits the icon frame).
x

ravi-morbia:
As for the animated icon, I will defer to @bendansby and @staceyadams as I am not familiar with that yet.

bendansby:
Let's hold off on animated icon since it's a whole other kettle of fish with the transition between the two states.

tesk9:
image

bendansby:
Still looks a bit non-flush? But I'll let @ravi-morbia  take it from here in terms of giving the go ahead or doing more revisions.

ravi-morbia:
This is strange. So it isn't supposed to be totally flush, that part is okay. But it also does not look completely centered. Here is what it looks like in Figma. Also attached the svg one more time in case that helps.

Screenshot 2023-03-17 at 5 19 33 PM

Any idea why it might look different here?

ravi-morbia:
Interesting, brought the svg export back into Figma and it is off for some reason. I will investigate and get back to you.

ravi-morbia:
Okay! After much toil, and Figma weirdness, I think the icon should export correctly now. In short Figma does not like objects not having a starting position exactly on a whole pixel. So, to maintain about the same visual weight while making the icon bigger, I basically had to make it fit exactly 23x23px within the 25x25px frame. Took some fun and janky techniques to get there, but I think this should work.

Attached is the (hopefully) final version of the icon for now. But let me know if you have any questions or issues.

x

ravi-morbia:
Yes, that looks good to me!

omnibs (approved):
I was concerned too about the change, mostly bc I think we scale these icons to match designs in a bunch of places, but it seems the difference is very subtle in the last iteration:

## First iteration
2023-03-21 at 10 18

## Current
2023-03-21 at 10 19

PR #1336, opened 2023-03-28
Adjust sidenav mobile behavior

tesk9:
## Context

Fixes notMobile, notQuizEngineMobile, etc media queries so that there's no overlap when using the query and the notQuery together (for instance, the current version of ClickableSvg has its icon disappear at a 1000 pixel viewport width when the mobileIcon is set -- not great!)

Also, adjusts the sidenav mobile behavior so that there's no tooltips for mobile and the selected category shows without needing to open the category list.

Based on [this antetype](https://cloud.antetype.com/exports/link/ed719e8c-1d8c-44f5-be5b-f1fdde859b7e?screen=id5182589l), although the mobile hamburger isn't pixel-perfect (I didn't want to mess with the general padding on the sidenav container).

Screen Shot 2023-03-28 at 1 55 34 PM
Screen Shot 2023-03-28 at 1 55 48 PM

cc @NoRedInk/design 

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog-app/App.elm
+++ b/component-catalog-app/App.elm
@@ -127,7 +127,11 @@ update action model =
                 route =
                     Routes.fromLocation model.moduleStates location
             in
-            ( { model | route = route, previousRoute = Just model.route }
+            ( { model
+                | route = route
+                , previousRoute = Just model.route
+                , isSideNavOpen = False
+              }
             , Maybe.map FocusOn (Routes.headerId route)
                 |> Maybe.withDefault None
             )
--- a/src/Nri/Ui/MediaQuery/V1.elm
+++ b/src/Nri/Ui/MediaQuery/V1.elm
@@ -15,6 +15,7 @@ module Nri.Ui.MediaQuery.V1 exposing
   - remove min-width:1 from media queries in order to support better composibility
   - adds narrowMobileBreakpoint and deprecates narrowMobileBreakPoint
   - adds withViewport for convenience when matching specific viewport size ranges
+  - fix `not` queries to not overlap with the regular breakpoint queries
 
 Standard media queries for responsive pages.
 
@@ -112,7 +113,7 @@ mobile =
 -}
 notMobile : MediaQuery
 notMobile =
-    only screen [ minWidth mobileBreakpoint ]
+    Css.Media.not screen [ maxWidth mobileBreakpoint ]
 
 
 {-| 1000px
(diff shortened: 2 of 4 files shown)


bendansby:
1. What’s the issue with changing the padding?
2. What do you think about collapsing the menu after an item is clicked? That way, the user won’t either have to scroll down to see the new content or scroll up to collapse the menu to see the new content. Also makes it clearer that new content has in fact loaded.
3. Happy about the tooltip behavior, but to be clear if it wasn’t, I was imagining this as a general tooltip parameter (perhaps packaged with tooltip type?)

tesk9:
> 1. What’s the issue with changing the padding?

Not a huge issue, but if I remove the padding from the sidenav container, then I have to add it to every entry instead so that the entry spacing doesn't change, and it seems a bit mistake-prone (cough or maybe _I'm_ mistake-prone)

> 2. What do you think about collapsing the menu after an item is clicked? That way, the user won’t either have to scroll down to see the new content or scroll up to collapse the menu to see the new content. Also makes it clearer that new content has in fact loaded.

Sounds good, but it might be hard to only do this while on the mobile view. I think it would be annoying if the sidenav on the side of the page closed after you clicked something, right? If it's not annoying, then this is easy.

> 3. Happy about the tooltip behavior, but to be clear if it wasn’t, I was imagining this as a general tooltip parameter (perhaps packaged with tooltip type?)

I think what's tricky is that what shows when the menu is closed is _not_ the tooltip content (which is "Open categories"/"Close categories") -- it's the name of the currently selected category. So I'm not sure how to generalize to other use cases. I like the idea of a clickable svg with a tooltip changing to a clickable text on mobile, but I don't think that is what's actually happening here, you know?

bendansby:
Why not just use a media query for some mobile container padding?

bendansby:
Since we’ll be using a collapsible sidebar much more (as in always) on mobile versus on desktop, let’s err toward optimizing for the mobile experience (first time for everything) and have it collapse if that makes sense to you.

tesk9:
Adjusted padding:

Screen Shot 2023-04-12 at 3 23 28 PM
Screen Shot 2023-04-12 at 3 23 34 PM

bendansby:
Looks good, thank you!

charbelrami (approved):
Looks good!

PR #1337, opened 2023-03-28
add Tabs v8: stickiness and background color

BrianHicks:

## Context

For ADM-705, we need to be able to have the tabs and table headers sticky on the screen. This is the first step there!

Note on the last checklist item here: this is true in V8 where it wasn't in V7. I know this is a bigger change but it opens up the possibilities!

There are no visual changes from V7; only behavior changes.

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -58,7 +58,7 @@ Please link to any relevant context and stories.
         - if you want to test the component directly, add tests under `tests/Spec`. Historically, this has been the more popular Elm testing strategy for noredink-ui.
 - [ ] Component API follows standard patterns in noredink-ui
     - e.g., as a dev, I can conveniently add an `nriDescription`
-    - and adding a new feature to the component will _not_ require major API changes to the comopnent
+    - and adding a new feature to the component will _not_ require major API changes to the component
 - [ ] Please assign [team-accessibilibats-a11ybats](https://github.com/orgs/NoRedInk/teams/team-accessibilibats-a11ybats) as a reviewer in addition to assigning a reviewer from your team
 
 
@@ -94,4 +94,4 @@ Please link to any relevant context and stories.
 - [ ] Changes to the component are tested/the new version of the component is tested
 - [ ] Component API follows standard patterns in noredink-ui
     - e.g., as a dev, I can conveniently add an `nriDescription`
-    - and adding a new feature to the component will _not_ require major API changes to the comopnent
+    - and adding a new feature to the component will _not_ require major API changes to the component
--- a/component-catalog-app/Examples/Tabs.elm
+++ b/component-catalog-app/Examples/Tabs.elm
@@ -132,7 +132,6 @@ example =
                                         , Maybe.map (\title -> moduleName ++ ".title " ++ Code.string title) settings.title
                                         , Maybe.map (\spacing -> moduleName ++ ".spacing " ++ String.fromFloat spacing) settings.customSpacing
                                         , Maybe.map (\color -> moduleName ++ ".pageBackgroundColor" ++ colorToCode color) settings.pageBackgroundColor
-                                        , Maybe.map (\color -> moduleName ++ ".highContrastPageBackgroundColor" ++ colorToCode color) settings.highContrastPageBackgroundColor
                                         , Maybe.map
                                             (\sticky ->
                                                 case sticky of
@@ -170,7 +169,6 @@ example =
                     , Maybe.map Tabs.title settings.title
                     , Maybe.map Tabs.spacing settings.customSpacing
                     , Maybe.map (Tabs.pageBackgroundColor << colorToCss) settings.pageBackgroundColor
-                    , Maybe.map (Tabs.highContrastPageBackgroundColor << colorToCss) settings.highContrastPageBackgroundColor
                     , Maybe.map
                         (\stickiness ->
                             case stickiness of
--- a/component-catalog-app/Examples/Tabs.elm
+++ b/component-catalog-app/Examples/Tabs.elm
@@ -144,7 +144,6 @@ example =
                                                             ++ Code.recordMultiline
                                                                 [ ( "topOffset", String.fromFloat stickyConfig.topOffset )
                                                                 , ( "zIndex", String.fromInt stickyConfig.zIndex )
-                                                                , ( "includeMobile", Code.bool stickyConfig.includeMobile )
                                                                 ]
                                                                 2
                                             )
@@ -338,7 +337,6 @@ initSettings =
                       , Control.record Tabs.TabListStickyConfig
                             |> Control.field "topOffset" (values [ 0, 10, 50 ])
                             |> Control.field "zIndex" (values [ 0, 1, 5, 10 ])
-                            |> Control.field "includeMobile" (Control.bool False)
                             |> Control.map Custom
                       )
                     ]
--- a/component-catalog-app/Examples/Tabs.elm
+++ b/component-catalog-app/Examples/Tabs.elm
@@ -122,13 +122,18 @@ example =
                         let
                             code =
                                 [ moduleName ++ ".view"
-                                , "    { title = " ++ Code.maybeString settings.title
-                                , "    , alignment = " ++ moduleName ++ "." ++ Debug.toString settings.alignment
-                                , "    , customSpacing = " ++ Code.maybeFloat settings.customSpacing
-                                , "    , focusAndSelect = identity"
+                                , "    { focusAndSelect = identity"
                                 , "    , selected = " ++ String.fromInt model.selected
-                                , "    , tabs = " ++ Code.listMultiline (List.map Tuple.first tabs) 2
                                 , "    }"
+                                , Code.listMultiline
+                                    (List.filterMap identity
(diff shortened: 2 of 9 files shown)


src/Nri/Ui/Tabs/V8.elm
```diff
@@ -192,6 +224,33 @@ defaultConfig =
     , spacing = Nothing
     , tabListBackgroundColor = Nothing
     , highContrastTabListBackgroundColor = Nothing
+    , tabListStickyConfig = Nothing
+    }
+
+
+{-| Configure how the top bar is sticky.
```
tesk9, line 232: Thanks for the super clear documentation here!

component-catalog-app/Examples/Tabs.elm
```diff
@@ -287,20 +293,31 @@ initSettings =
...
-        |> Control.field "withTooltips" (Control.bool True)
-        |> Control.field "tabListBackgroundColor" (Control.maybe False colorChoices)
-        |> Control.field "highContrastTabListBackgroundColor" (Control.maybe False colorChoices)
+
+
+values : List a -> Control a
+values nums =
+    nums
+        |> List.map (\n -> ( Debug.toString n, Control.value n ))
+        |> Control.choice
```
tesk9, line 353: Could you make this more specific (so it doesn't use Debug.toString) and move it to `Debug.Control.Extra` where the other extensions to Debug.Control live? Thanks!

.github/pull_request_template.md
```diff
@@ -58,7 +58,7 @@ Please link to any relevant context and stories.
         - if you want to test the component directly, add tests under `tests/Spec`. Historically, this has been the more popular Elm testing strategy for noredink-ui.
 - [ ] Component API follows standard patterns in noredink-ui
     - e.g., as a dev, I can conveniently add an `nriDescription`
-    - and adding a new feature to the component will _not_ require major API changes to the comopnent
+    - and adding a new feature to the component will _not_ require major API changes to the component
```
tesk9, line 61: thanks!

src/Nri/Ui/Tabs/V8.elm
```diff
@@ -0,0 +1,473 @@
...
+    [ alignmentStyles
+    , Css.margin Css.zero
+    , Css.fontSize (Css.px 19)
+    , Css.displayFlex
+    , Css.flexGrow (Css.int 1)
+    , Css.padding Css.zero
+    , maybeStyle Css.backgroundColor config.tabListBackgroundColor
+    , maybeStyle
+        (\color -> MediaQuery.highContrastMode [ Css.backgroundColor color ])
+        config.highContrastTabListBackgroundColor
```
tesk9, line 391: So I don't think this is will actually do anything -- I think high contrast mode will coerce the background color no matter what you pass in (unless you set another style that prevents high contrast mode from working, which I don't recommend). 

Screen Shot 2023-03-29 at 9 14 35 AM

I think it's probably fine without 🤷‍♀️

src/Nri/Ui/Tabs/V8.elm
tesk9: I wonder about the mobile attribute -- do we know we need stickiness for mobile at this point? Could we _not_ include this customization yet, and disallow stickiness for mobile?

(There's talk of having tabs collapse down to a different structure entirely for mobile)

tesk9 (commented):
Looking good! Thank you for all the careful attention to docs and code examples! Very appreciated.

One thing I'm wondering is from an API standpoint how to incorporate Juan's coming changes. He's going to need to pass in the background color for the page. Do you expect the tab list background color to ever differ from the background color for the page? If not, I wonder if the API should be generalized a bit -- maybe `tabListBackgroundColor` to `pageBackgroundColor`?

src/Nri/Ui/Tabs/V8.elm
```diff
@@ -128,24 +128,68 @@ type Alignment
     | Right
 
 
-{-| -}
-view :
+type Attribute id msg
```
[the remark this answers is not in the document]
BrianHicks: Yeah, I'll do that. I did this here but used the other pattern in the table work for comparison and it changed my mind 👍

src/Nri/Ui/Tabs/V8.elm
BrianHicks: I just made it configurable because you mentioned it *might* not be a good idea. If we're going to avoid it entirely, I'll just remove the option.

src/Nri/Ui/Tabs/V8.elm
BrianHicks: does it still stop the panel contents from showing through? The point here is for it to be opaque, not necessarily a specific color.

BrianHicks:
> Do you expect the tab list background color to ever differ from the background color for the page?

Nope!

> If not, I wonder if the API should be generalized a bit -- maybe tabListBackgroundColor to pageBackgroundColor?

Yep! Good call, thanks.

src/Nri/Ui/Tabs/V8.elm
tesk9: Ohh I see. I thought the default meant there was design buy-in on not stickifying for mobile 🤔

src/Nri/Ui/Tabs/V8.elm
tesk9: Yup!

tesk9: You can test it on assistivlabs (there's a free trial that doesn't require a card or anything)

src/Nri/Ui/Tabs/V8.elm
BrianHicks: oh, no, but it'd be good to ask… @bendansby what do you think about having headers in tab components and tables not sticky by default on mobile?

bendansby:
This is maybe a question for @juanedi but what do we think about just being able to set a color for the active tab in a set of tabs to fade to rather than programmatically connect it to a page background?

tesk9:
I think Juan is OOO today, but I think that's what he was originally trying to do, but ran into trouble with the bottom border of the selected tab element (which needs to be the same color as the tabpanel background in order for the tab to look selected). Making the border bottom transparent doesn't work (I believe because the tablist also sets border bottom, and we're using negative margin to scootch the tab over the line). Juan left some [implementation thoughts here](https://noredink.slack.com/archives/C02NVG4M45U/p1679876351872079?thread_ts=1679876069.832909&cid=C02NVG4M45U)

bendansby:
Oh yeah, I would assume the fade to background color parameter would also set the border color

src/Nri/Ui/Tabs/V8.elm
BrianHicks: and that is how it is in the latest version of this branch, so yay!

BrianHicks:
I think the tab color is going to be a slightly different issue from the background color. He and I talked through this a little before he was OOOing. It seemed like he was fine for us to proceed and Quetz will figure out how to integrate later.

src/Nri/Ui/Tabs/V8.elm
BrianHicks: thanks. I couldn't find a case where this made any difference, so I removed it. Looks like the regular BG will be OK for what we're doing 💯

BrianHicks:
ok @tesk9 @charlieshanley, this is ready for another review or merge!

component-catalog-app/Examples/Tabs.elm
```diff
@@ -122,13 +123,35 @@ example =
                         let
                             code =
                                 [ moduleName ++ ".view"
-                                , "    { title = " ++ Code.maybeString settings.title
-                                , "    , alignment = " ++ moduleName ++ "." ++ Debug.toString settings.alignment
-                                , "    , customSpacing = " ++ Code.maybeFloat settings.customSpacing
-                                , "    , focusAndSelect = identity"
+                                , "    { focusAndSelect = identity"
```
charlieshanley, line 126: I see this was `identity` before your changes, but would it make sense to change it to `FocusAndSelectTab`, or something like that?

charlieshanley:
It seems the title property messes up the masking of content when the tabs list is sticky:

https://user-images.githubusercontent.com/8228888/228837524-15c11d7b-f12c-4f13-bc9b-7bf47237d9ac.mov

charlieshanley:
> title property ... masking of content...

I think moving the `background-color` up one element should fix it.

component-catalog-app/Examples/Tabs.elm
BrianHicks: this is in the example code, right? Probably not, because then we'd need to define that

charlieshanley (approved):
Looks great!
giphy - 2023-03-30T125620 647

charlieshanley:
> I did this and it didn't; the problem is still that little strip of white border. It wouldn't be the end of the world to move the background color into TabInternal, though, so I think we're moving in the right direction here.

It looks like you fixed the thing I was aware of. I'm not sure what the problem with the white border is.

tesk9 (approved):
Fabulous! Thank you, Brian!

krinhorn:
oh hey this is officially noredink-ui's [1337est](https://www.wikiwand.com/en/Leet) PR 🏆

BrianHicks:
zero-wing-art-gif-by-bitwave-games

PR #1338, opened 2023-04-03
add sticky headers to sortable tables

BrianHicks:

API design note: unlike tabs, we don't currently know any use-case aside from stickiness where we'd need to set the page background color separate from stickiness, so I combined them!

## Context

Part of ADM-705

No changes to the table component except that now the header can be made sticky.

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog-app/Examples/AnimatedIcon.elm
+++ b/component-catalog-app/Examples/AnimatedIcon.elm
@@ -88,7 +88,7 @@ example =
                         ]
                 }
             , Heading.h2 [ Heading.plaintext "Example" ]
-            , Table.view
+            , Table.view []
                 [ Table.custom
                     { header = text "Rendered"
                     , view =
--- a/component-catalog-app/Examples/AnimatedIcon.elm
+++ b/component-catalog-app/Examples/AnimatedIcon.elm
@@ -19,7 +19,7 @@ import Nri.Ui.AnimatedIcon.V1 as AnimatedIcon
 import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Heading.V3 as Heading
 import Nri.Ui.Svg.V1 as Svg
-import Nri.Ui.Table.V6 as Table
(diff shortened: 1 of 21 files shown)


BrianHicks:
@bendansby do the Percy changes look OK to you? A couple sub-5px changes that are going to be very tricky to eliminate.

bendansby:
Looks like table header height got reduced a bit? Why's that?

BrianHicks:
It now has to use an inset box shadow instead of a border. We need this because we collapse border tables, making the header bottom border stick to the table instead of the header. I tried adding a bit of padding to the bottom to fix this, but it doubles the border height. If the 3px difference matters for us, I can spend some more time on it.

charbelrami (approved):
Looks great!

bendansby:
Can we just use a fixed table header height?

BrianHicks:
I can try. What height would you like?

bendansby:
45px plz

tesk9:
(Fixed _minimum_ table height, please! The height will still need to expand if users adjust their font size settings or if the text in the heading is quite long)

BrianHicks:
turns out the header is 47.5px tall right now, and 46 in this branch. I'll try to make it 45, though.

BrianHicks:
well it turns out the margin was messed up—the text was much closer to the border, despite the container being smaller.

Here's how I've adjusted it (15px from baseline to border)

image

and the currently-released version (also 15px from baseline to border):

image

We can shrink that to 45px default height if you want. No big deal at this point. Since it turns out it was a padding thing, much easier to adjust.

bendansby:
Oh great, that looks better and sure let's go for 45. Thanks!

BrianHicks:
45ified!

image

bendansby:
image

charlieshanley (approved):
Looks good!
giphy - 2023-04-04T133728 104

PR #1347, opened 2023-04-05
Tessa/test suggestions

tesk9:
Suggestions for https://github.com/NoRedInk/noredink-ui/pull/1254:

- adds additional assertions for Block tests
- improves test coverage for Highlighter (was only testing overlapping highlighter views, there are 6 other highlighter views that are all more commonly used)
- adds failing assertions for Highlighter rendering the tags visually as markdown correctly

---

> I'm not quite sure I know what you mean by "sometimes-visually-displayed" start tag

If using `Highlighter.staticWithTags` or `Highlighter.staticMarkdownWithTags`, names for the highlights will be displayed (if they exist):

Screen Shot 2023-04-05 at 3 01 56 PM

(Note that the markdown is being rendered raw in the screenshot -- I adjusted the name in the "claimMarkerWithBorder" helper to get into this state.)

Additionally, if users are in high contrast mode, this tag will **always render** if the name is present for every highlighter view:

Screen Shot 2023-04-05 at 3 03 45 PM

Screen Shot 2023-04-05 at 3 03 57 PM

The tag is meaningfully equivalent to the label used in Block.

---

I wasn't thinking about these before, but while adding these tests a couple of points on concern occurred to me:

1. If we add semantics (e.g., strong, em) to the label, that information will not be conveyed to AT users. I'm curious why we need markdown support in the labels? is it mostly for visual affordances or is it going to be meaningful? This is something I will take back to the a11ybats too.
2. Adding the paragraph wrapper to the label content is a bit funky since it introduces a paragraph tag _inside another paragraph tag_. The next version of WCAG is going to remove the HTML parsing validity requirement, so I don't know that it matters a ton. So mostly just a heads up that you can do with what you will.

### DIFF
--- a/tests/Spec/Nri/Ui/Block.elm
+++ b/tests/Spec/Nri/Ui/Block.elm
@@ -1,11 +1,13 @@
 module Spec.Nri.Ui.Block exposing (spec)
 
+import Accessibility.Aria as Aria
 import Browser.Dom exposing (Element)
 import Dict
 import Expect
 import Html.Attributes as Attributes
 import Html.Styled
 import Nri.Ui.Block.V4 as Block
+import Spec.PseudoElements exposing (hasAfter, hasBefore)
 import Test exposing (..)
 import Test.Html.Query as Query
 import Test.Html.Selector as Selector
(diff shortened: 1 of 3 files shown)


micahhahn:
> If we add semantics (e.g., strong, em) to the label, that information will not be conveyed to AT users. I'm curious why we need markdown support in the labels? is it mostly for visual affordances or is it going to be meaningful? This is something I will take back to the a11ybats too.

It's entirely possible curriculum would never have need of markdown in block labels - I was just trying to support markdown **everywhere** they can change content in scaffolding for a consistent editor experience. 

> Adding the paragraph wrapper to the label content is a bit funky since it introduces a paragraph tag inside another paragraph tag. The next version of WCAG is going to remove the HTML parsing validity requirement, so I don't know that it matters a ton. So mostly just a heads up that you can do with what you will.

I didn't realize we were already wrapping in a paragraph! I'm perfectly happy to remove it.

micahhahn:
I've asked curriculum if they need markdown support in labels [in this Slack message](https://noredink.slack.com/archives/C02TBGTURFX/p1680731488086399). If they don't, we can just remove it and undo a lot of this code :)

tesk9:
Awesome, thank you!

PR #1351, opened 2023-04-06
Switch label font size

bendansby:
Makes it 15px like similar labels checkbox and radio button.

image


tesk9 (approved):
Awesome, thanks!

PR #1353, opened 2023-04-10
Adjust tooltip example

tesk9:
ClickableText now automatically adds external link visual indication, so it doesn't make sense to add the icon twice 😅 

Fixes A11-2616

image

Also, adjusts the copy a bit to give different information than the information that is already conveyed by the SVG.


krinhorn:
🤔 hmmm, I feel like the new copy makes this more confusing and isn't actually copy we'd recommend using in the app. "Opens in a new tab/window" was more clear and consistent, IMO.

(Also, the CC isn't "noredink-ui," so we're not leaving noredink-ui anyway.)

charbelrami (approved):

PR #1354, opened 2023-04-11
Remove extra padding from Header

bendansby:
Napkin 17 04-11-23, 8 48 27 AM


tesk9 (commented):
I think this makes the content too wide when not on mobile:

image

I think maybe using `Spacing.centeredContentWithCustomWidth` instead might work?

PR #1358, opened 2023-04-17
move component-catalog-app back under component-catalog/src

BrianHicks:
This rearranges some code in the component catalog to be more readily-understood by someone coming to the project for the first time.

It also conveniently rearranges the code in a way that it's more amenable to analysis by the build tools I'm trying out on noredink-ui. 😆

It looks like the code was this way originally for the purposes of the npm package? But I ran `npm publish` (without actually publishing) and nothing from `component-catalog` is included, so I'm not sure why else we would have done this. 🤷


tesk9 (approved):
Looks good, thanks!

PR #1359, opened 2023-04-17
upgrade npm and component-catalog Elm packages

BrianHicks:
This gets us to the latest everything for build tooling and compatible upgrades in component-catalog.

At very least, this should avoid the versioning warnings in Percy... maybe even take care of a flake or two!

### DIFF
--- a/component-catalog/elm.json
+++ b/component-catalog/elm.json
@@ -28,10 +28,10 @@
             "elm-community/string-extra": "4.0.1",
             "myrho/elm-round": "1.0.5",
             "pablohirafuji/elm-markdown": "2.0.5",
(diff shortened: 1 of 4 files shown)


BrianHicks:
ugh, this was passing CI the first time I pushed it, but after I reverted a commit it flaked. I hope that's not a symptom of a bigger problem. I'm going to leave this as reviewable for now in the hopes that it'll pass on retry—and because I have to leave for the day—but it may need more work before it can actually merge.

BrianHicks:
marking this a draft again. I don't want to end up having this be flakier!

tesk9 (approved):
Looks good to me! Thanks!

PR #1363, opened 2023-04-19
buck2 in CI

BrianHicks:
TODO


BrianHicks:
Sharing some vision! This is a "mega" PR. Make sense, as it's the second proof of concept (the first was local; I never pushed it to GH.) I'm going to break this down into a couple of smaller and easier-to-review "mini" PRs. Each of those will get us closer to using Buck for CI. Once every Shake target is replaced by a Buck one, we'll shut down the Shake pipeline steps.

We'll probably always use the special Netlify configuration for this repo, though—it seems like you have to use Netlify's CD to get deploy previews and those are too useful to abandon just to eliminate an edge case.

Sound OK @NoRedInk/team-accessibilibats-a11ybats?

tesk9:
Oh yeah I _love_ the deploy previews! (They're persistent, so I've used old ones months after a PR has merged, too!)

Small PRs sounds good to me! Thanks!

PR #1367, opened 2023-04-25
replace old apple, add appleOutline

erik-nri:
This simply replaces the existing `apple` UiIcon and adds a new one for `appleOutline` and it did not match up with any of the PR templates 😅. I am replacing the old one instead of adding the new apple alongside per the guidance from [this](https://noredink.slack.com/archives/CB5ERR51P/p1678815470369669) slack convo. Also I removed the `title`s that were present in the versions of this in monolith as requested 😄.


src/Nri/Ui/UiIcon/V1.elm
```diff
@@ -2031,26 +2031,56 @@ graduateCap =
...
+                    , Attributes.fill "#707070"
```
tesk9, line 2050: This is setting the fill color to gray, which I don't think is what's actually desired. In the customizable view, for instance, the apple should turn green when I select green:

image

Also if you hover over the icons, their color is supposed to change, and the apple's color doesn't:

image

src/Nri/Ui/UiIcon/V1.elm
```diff
@@ -2031,26 +2031,56 @@ graduateCap =
...
+                    , Attributes.stroke "#707070"
+                    , Attributes.strokeWidth "2"
+                    , Attributes.fill "#707070"
```
tesk9, line 2080: Same problem with fill and stroke color here -- they should change depending on the context, rather than being hardcoded to gray.

tesk9 (commented):
Thanks for the PR! 

I think the relevant section of the PR template is ["Modifying a Component"](https://github.com/NoRedInk/noredink-ui/blob/master/.github/pull_request_template.md#wrench-modifying-a-component), which includes a section asking for screenshots.

This is the kind of change where it's a good idea to take careful screenshots and ping design because it's actually really easy for the svg viewport/sizing settings to end up a bit off (because of figma, because of human mistakes -- for whatever reason, it can happen). Check out the amount of back and forth we ended up having when I [changed out the X icon](https://github.com/NoRedInk/noredink-ui/pull/1322) for a slightly less chonky version a few weeks ago, for example 😅 (In this particular case, apart from the color not changing correctly, the viewport does look fine to me -- just as a heads up for future icon changes)

src/Nri/Ui/UiIcon/V1.elm
erik-nri: Oh jeez, it took me embarrassingly long to realize that customization view was just a scroll down away 😅. But I think I got it now, setting the fill to `"currentcolor"` as the others gets the desired behaviors.

erik-nri:
Tessa! Thanks so much for the review and also for providing the additional context above!

erik-nri:
@bendansby I underestimated this! Tagging you now based on Tessa's comment above. Outside of the screenshot Tessa added is there anything else I can hunt down that would be helpful? Or would that be better served by checking the differences on the apple when it gets updated over on monolith?

bendansby (approved):
Looks good to me, thanks! When we bump noredink-ui, we can do a standard Percy/QA regression sweep.

tesk9 (approved):
Wooo apples!

PR #1372, opened 2023-04-27
build Elm docs and the component catalog Elm app in Buck

BrianHicks:
This adds the Elm docs and component catalog app to the Buck 2 build.

It's doing things in a slightly heavier way than I would want if this code was only ever going to live in this repo, but:

1. I'd like it to work in the monorepo eventually, so I've avoided making assumptions that only apply to noredink-ui that would simplify the build code.
2. I'd *really* like it if we could eventually publish `prelude-nri` as a separate package so other folks could build Elm apps with this code.

Oh and an implied third reason: there aren't any Elm rules for Buck, so we're blazing a trail here and sometimes build rules get a little hairy like this. It's OK! In an ideal world, you would just do the configuration you see in `toolchains/BUCK`, `BUCK`, and `component-catalog/BUCK` and that would be the end of it! And that's pretty minimal!


ArthurJordao:
@BrianHicks I didn't have time yet to see this because of the refactoring that I am doing, I was thinking if we could take our pairing time to pair-review your PR (as I don't have too many context on this), WDYT?

BrianHicks:
that'd be just fine. 🙂

ArthurJordao (commented):
Hey we tried it locally on my machine and for some reason changing the contents of a file inside source does not modify the output of the compilation for docs.

BrianHicks:
following up with context from our Slack conversation: `docs.json` only changes if you change one of the exposed modules. I feel silly that I didn't remember that earlier. 😆

ArthurJordao (approved):
Awesome work Brain. Thank you for walk me through on the changes :D

PR #1373, opened 2023-04-27
[PHX-675] Fix occasionally overlapping labels

micahhahn:

## Context

We've had many many reports from curriculum that labels were overlapping in places where they shouldn't be, but it was almost impossible to reproduce!  Sometimes it would happen in previews, other times only when debugging the QE.  

Finally I made the connection while looking into this that the "flicker" I see while resizing the QE shouldn't be there! I though it was due to labels temporarily overlapping while calculations were being made - but that's not the case.  @tesk9 's code should never temporarily produce an overlap. 

I think this is flaky due to floating point inconsistencies and weird measurement behavior.  The fix is simple - instead of grouping by exact `element.y + element.height`, we use a much more forgiving test to see if any of the range from `(element.y, element.y + element.height)` is overlapping with other labels. We can do this because our labels are grouped into different bands of vertical space that are separated by the sentence.

### Before

https://user-images.githubusercontent.com/12899207/235009505-248ddd5c-1ee2-4b0a-8568-7d377df2f4bf.mov

### After

https://user-images.githubusercontent.com/12899207/235009519-791c4f46-995a-4996-b315-461a1f801fd7.mov

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/src/Nri/Ui/Block/V3.elm
+++ b/src/Nri/Ui/Block/V3.elm
@@ -196,7 +196,9 @@ splitByOverlaps =
     -- consider the elements from top to bottom
     groupWithSort (\( _, a ) -> a.label.element.y + a.label.element.height)
         (\( _, a ) ( _, b ) ->
-            (a.label.element.y + a.label.element.height) == (b.label.element.y + b.label.element.height)
+            isRangeOverlapping
+                ( a.label.element.y, a.label.element.y + a.label.element.height )
+                ( b.label.element.y, b.label.element.y + b.label.element.height )
         )
         >> List.concatMap
             -- consider the elements from left to right
@@ -205,6 +207,11 @@ splitByOverlaps =
                     (a.label.element.x + xOffsetPx a.label + a.label.element.width) >= (b.label.element.x + xOffsetPx b.label)
                 )
             )
+
+
+isRangeOverlapping : ( Float, Float ) -> ( Float, Float ) -> Bool
+isRangeOverlapping ( a1, a2 ) ( b1, b2 ) =
+    (a1 <= b2 && a2 >= b1) || (a1 <= b2 && a2 >= b1)
 
 
 groupWithSort : (a -> comparable) -> (a -> a -> Bool) -> List a -> List (List a)
--- a/src/Nri/Ui/Block/V4.elm
+++ b/src/Nri/Ui/Block/V4.elm
@@ -185,7 +185,9 @@ splitByOverlaps =
     -- consider the elements from top to bottom
     groupWithSort (\( _, a ) -> a.label.element.y + a.label.element.height)
         (\( _, a ) ( _, b ) ->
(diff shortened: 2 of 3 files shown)


micahhahn:
@tesk9 If you want an example of a weird measurement I observed in Chrome that causes overlapping labels,  here is one that I managed to capture: 

Notice how the  `label.element.y` and then `labelContent.element.y` don't line up.  I don't understand how that happens - maybe we're catching the browser in a weird render state? 

```
h : Dict String { label : MockElement, labelContent : MockElement }
h =
    Dict.fromList
        [ ( "173-display-element-labeled-2-5"
          , { label =
                { scene =
                    { width = 633
                    , height = 748
                    }
                , viewport =
                    { x = 0
                    , y = 0
                    , width = 633
                    , height = 748
                    }
                , element =
                    { x = 350.984375
                    , y = 212.5
                    , width = 150
                    , height = 53
                    }
                }
            , labelContent =
                { scene =
                    { width = 627
                    , height = 782
                    }
                , viewport =
                    { x = 0
                    , y = 0
                    , width = 618
                    , height = 733
                    }
                , element =
                    { x = 350.984375
                    , y = 261.5
                    , width = 150
                    , height = 45
                    }
                }
            }
          )
        , ( "173-display-element-labeled-6-12"
          , { label =
                { scene =
                    { width = 627
                    , height = 782
                    }
                , viewport =
                    { x = 0
                    , y = 0
                    , width = 618
                    , height = 733
                    }
                , element =
                    { x = 477.2421875
                    , y = 212.5
                    , width = 150
                    , height = 102
                    }
                }
            , labelContent =
                { scene =
                    { width = 627
                    , height = 782
                    }
                , viewport =
                    { x = 0
                    , y = 0
                    , width = 618
                    , height = 733
                    }
                , element =
                    { x = 477.2421875
                    , y = 212.5
                    , width = 150
                    , height = 45
                    }
                }
            }
          )
        ]
```

tesk9 (commented):
Dang what a tricky bug! Great work figuring it out! 

`Spec.Nri.Ui.Block` has some positioning-related specs set up, if you wanted to add that example as a regression test. 

My more real question is should this fix be on Block/V4?

micahhahn:
> My more real question is should this fix be on Block/V4?

🤦‍♂️ - why yes of course ... in fact can we remove `Block.V3`? It doesn't appear to be used anymore

micahhahn:
> Spec.Nri.Ui.Block has some positioning-related specs set up, if you wanted to add that example as a regression test.

Interestingly enough my changes had broken 3 unit tests! And CI was still green - are those not getting run as part of CI?

tesk9:
> can we remove Block.V3? It doesn't appear to be used anymore

We do periodic deletions of old components based on what's in use and what isn't, and we'll remove it then. (We avoid doing lots of major upgrades, which is what a deletion requires!)

PR #1375, opened 2023-05-01
[PHX-709] Borders are not getting blended when markdown is directly underneath emphasis

micahhahn:
This PR fixes a bug that was introduced in where markdown **inside** of an emphasis block would have left + right borders between each element if it was at the start or the end of the emphasis.

- https://github.com/NoRedInk/noredink-ui/pull/1254

@tesk9 I'm not sure that just applying the styles on the the parent element is ideal - while there are no visual differences, the styling for individual words is different than for other words not in markdown.  But the API between Block and Mark didn't give me much wiggle room to do anything else.  

I considered extending the type of `viewWithBalloonTags` `renderSegment` function from

```elm
renderSegment : c -> List Style -> Html msg
```

to

```elm
renderSegment : c ->  { startStyles : List Style, styles : List Style, endStyles : List Style }  -> Html msg
```

which gives the caller the power to delegate which descendant node should receive the `startStyles` / `endStyles`.  But there are a handful of `renderSegment` looking functions in the `Mark.V2` module and I didn't want to break the mold. 

## Context

Please link to any relevant context and stories.

| Before | After |
| --- | --- |
| image | image |

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog/src/Examples/Block.elm
+++ b/component-catalog/src/Examples/Block.elm
@@ -210,6 +210,13 @@ example =
                     , Block.labelId editorsNoteId
                     , Block.labelPosition (Dict.get editorsNoteId offsets)
                     ]
+                , Block.view 
+                    [ Block.plaintext " "
+                    ]
+                , Block.view
+                    [ Block.content [ Block.italic (Block.phrase "Emphasized Italics.") ]
+                    , Block.emphasize
+                    ]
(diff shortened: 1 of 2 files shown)


tesk9 (approved):
Looks reasonable to me!

PR #1376, opened 2023-05-01
build project JavaScript with Buck

BrianHicks:
Builds `lib/lib.js` with Buck.

As in #1372, this PR is bigger than I'd strictly like—but if you ignore adding more prelude stuff, it's actually pretty decently sized!


BrianHicks:
yeah, I think so. "Bad CPU type in executable" says all we need to know. Wanna hop on a quick pair and fix this for you?

ArthurJordao (approved):

PR #1381, opened 2023-05-02
reorganize Elm deps to be a bit neater

BrianHicks:
I found a nicer pattern for handling language downloads in #1376, so I'm porting it to Elm too!

### DIFF
--- a/prelude-nri/elm/isolated_compile.py
+++ b/prelude-nri/elm/isolated_compile.py
@@ -22,7 +22,7 @@ def symlink_if_necessary(source, target):
      - If the target already exists and points somewhere else, remove it
        and relink.
     """
-    if os.path.exists(target):
+    if os.path.lexists(target):
         if os.path.isfile(target):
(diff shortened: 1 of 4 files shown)


prelude-nri/elm/isolated_compile.py
```diff
@@ -22,7 +22,7 @@ def symlink_if_necessary(source, target):
      - If the target already exists and points somewhere else, remove it
        and relink.
     """
-    if os.path.exists(target):
+    if os.path.lexists(target):
```
ArthurJordao, line 25: Didn't know the different between lexists vs exists, but makes sense for this case, nice catch!

ArthurJordao (approved):

prelude-nri/elm/isolated_compile.py
BrianHicks: I didn’t either, until it failed for a broken symlink locally. 😆

PR #1387, opened 2023-05-04
adds `hoverZIndex` to SortableTable.StickyConfig

celsobonutti:

## Context

We need a tooltip inside a tab header to be displayed over a tab bar.
The suggestion @tesk9 gave us was lowering the tab bar zIndex, which worked quite well, except now the tab header is being displayed [_over_ the tab bar](https://github.com/NoRedInk/NoRedInk/pull/43950#issuecomment-1534852849). One solution we thought about was lifting the tab header z-index only when it's being hovered, since that would put it in a stacking context above the tab bar, while allowing it to be under it when the user is scrolling.

image
image

- [X] I've gone through the relevant sections of the [Development Accessibility guide](https://paper.dropbox.com/doc/Accessibility-guide-4-Development--BiIVdijSaoijjOuhz3iTCJJ1Ag-rGoHpC91pFg3zTrYpvOCQ) with the changes I made to this component in mind
- [X] Changes are clearly documented
    - [X] Component docs include a changelog
    - [X] Any new exposed functions or properties have docs
- [X] Changes extend to the Component Catalog
    - [X] The Component Catalog is updated to use the newest version, if appropriate
    - [X] The Component Catalog example version number is updated, if appropriate
    - [X] Any new customizations are available from the Component Catalog
    - [X] The component example still has:
        - an accurate preview
        - valid sample code
        - correct keyboard behavior
- [X] Changes to the component are tested/the new version of the component is tested
- [X] Component API follows standard patterns in noredink-ui

### DIFF
--- a/component-catalog/src/Examples/SortableTable.elm
+++ b/component-catalog/src/Examples/SortableTable.elm
@@ -154,6 +154,14 @@ example =
                                                         [ ( "topOffset", String.fromFloat stickyConfig.topOffset )
                                                         , ( "zIndex", String.fromInt stickyConfig.zIndex )
                                                         , ( "pageBackgroundColor", "Css.hex \"" ++ stickyConfig.pageBackgroundColor.value ++ "\"" )
+                                                        , ( "customZIndex"
+                                                          , case stickyConfig.hoverZIndex of
+                                                                Nothing ->
+                                                                    "Nothing"
+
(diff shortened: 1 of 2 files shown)


bendansby (approved):

charbelrami (approved):
Looks good! Thanks!

PR #1388, opened 2023-05-05
enforce elm-format with buck2

BrianHicks:
This enforces elm-format styling in Buck. It does it in a new way! If it's building a push event, it'll do the normal thing. BUT if it's building against a pull request, it'll lift the elm-format diffs to a set of suggestions. It ends up looking like this to apply:

image


prelude-nri/elm/BUCK
```diff
@@ -26,3 +26,33 @@ genrule(
     executable = True,
     visibility = ["PUBLIC"],
 )
+
+ELM_FORMAT_URL = select({
```
stoeffel, line 30: We can't use nix for this?

script/elm_format.py
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env python
```
stoeffel, line 1: One of the things that I dislike about using buck that it introduces so many different languages. I'd much rather do this in haskell than in python (maybe nix-script-haskell?).

.github/workflows/buck2-ci.yml
```diff
@@ -24,3 +24,19 @@ jobs:
         run: script/buck2 test //... -- --env PERCY_TOKEN=$PERCY_TOKEN
         env:
           PERCY_TOKEN: '${{ secrets.PERCY_TOKEN }}'
+
+      - name: buck2 elm-format (with comments)
+        if: github.event.pull_request
+        run: |
+          set -x
+          script/buck2 run //:elm_format -- \
```
stoeffel, line 32: Would be cool if we could make this run `//:format` and that would just format everything that is to be formatted? Like I guess we also want prettier for md and js.

stoeffel (approved):
Nothing blocking only some thoughts.

prelude-nri/elm/BUCK
BrianHicks: I’m not sure how we’re going to use Nix for this. I’d like to eventually but it’s a bigger job to figure out how it works with Buck’s download cache. For now, it seems fine to download the right binaries.

script/elm_format.py
BrianHicks: They recommend using Python scripts because there’s an easy Python bootstrap path. We can certainly move these to a different language if we’re willing to add the right compiler tool chains! I just haven’t been so far. This particular script is generally useful, and we’ll probably end up rewriting it in something else.

.github/workflows/buck2-ci.yml
BrianHicks: Yes! That’s my vision for it too. There’s very little in the script that’s specific to elm-format. It was a bigger change than makes sense for this PR though!

PR #1390, opened 2023-05-09
A11-2658 Checkbox tests

charbelrami:
Adds tests to the checkbox component.

### DIFF
--- a/tests/Spec/Nri/Ui/Checkbox.elm
+++ b/tests/Spec/Nri/Ui/Checkbox.elm
@@ -2,13 +2,11 @@ module Spec.Nri.Ui.Checkbox exposing (..)
 
 import Accessibility.Aria as Aria
 import Accessibility.Role as Role
-import Expect
 import Html.Styled exposing (..)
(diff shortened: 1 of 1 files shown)


tests/Spec/Nri/Ui/Checkbox.elm
```diff
@@ -0,0 +1,131 @@
...
+    ]
+
+
+ensureHasRoleCheckbox : TestContext -> TestContext
+ensureHasRoleCheckbox context =
+    context
+        |> ensureView
+            (Query.findAll [ Selector.attribute Role.checkBox ]
+                >> Query.count (Expect.equal 1)
+            )
```
tesk9, line 53:
```suggestion
        |> ensureViewHas [ Selector.attribute Role.checkBox ]
```

I think for this helper and the 2 helpers below it, `ensureViewHas` can make the assertion a bit more compact and remove the need for `Query`. (At which point, I wonder if the helpers should actually be inlined, instead of being separate functions?)

tesk9 (approved):
Thanks!

PR #1391, opened 2023-05-09
enforce Prettier in all JavaScript files with Buck

BrianHicks:
Does the same enforcement comments as `elm-format`, but with `prettier` driving the diffs.

image


stoeffel (approved):
Looking great!

PR #1392, opened 2023-05-10
enforce Prettier formatting in Markdown and HTML

BrianHicks:
Does what it says on the tin!


BrianHicks:
@charbelrami requesting review from a11ybats because this is enforcing a new rule in the library. Is that OK with y'all?

charbelrami (approved):
Looks great! Thank you

ArthurJordao (approved):

PR #1395, opened 2023-05-16
Fix line-breaking by going full `whitespace: pre` and using `<wbr/>`

micahhahn:

## Context

We've had lots of reports of line breaking that is not ideal - especially when punctuation is emphasized.  I've tried various different approaches to fix this including: 
- Detecting when segments should not have space between them, splitting off the first/last word from plaintext, and wrapping the elements in a "nowrap" span (see https://github.com/NoRedInk/NoRedInk/pull/44103)
- Injectecting zero width joiners (`&NoBreak;`) between the span elements that shouldn't break

Nothing has worked completely. 

I think the only way we can get this breaking behavior exactly correct is to go full `whitespace: pre` on the entire sentence container, and then manually insert break point opportunities (`<wbr/>`) after every space.  Note here that being intentional about putting these **after** spaces means that we fix another annoyance where you can witness a space get bumped to the next line causing it to not line up correctly. 

https://github.com/NoRedInk/noredink-ui/assets/12899207/6f1eb7eb-b1f2-4ae4-b5dc-e85f82f10326

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog/src/Examples/Block.elm
+++ b/component-catalog/src/Examples/Block.elm
@@ -270,6 +270,8 @@ example =
                 , Block.view
                     [ Block.plaintext ", that should not break before this comma!"
                     , Block.label "Dont break"
+                    , Block.labelId dontBreakId
+                    , Block.labelPosition (Dict.get dontBreakId offsets)
                     ]
                 ]
             , Table.view []
@@ -576,6 +578,11 @@ longId =
     "long-label-id"
 
 
+dontBreakId : String
+dontBreakId =
+    "do-not-break-label-id"
+
+
 blocksWithLabelsIds : List String
 blocksWithLabelsIds =
     [ ageId
@@ -591,6 +598,7 @@ blocksWithLabelsIds =
     , goalId
     , shortId
     , longId
+    , dontBreakId
     ]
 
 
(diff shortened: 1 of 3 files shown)


tesk9:
This approach makes sense to me! I pushed a commit that should fix the label repositioning

micahhahn:
> I pushed a commit that should fix the label repositioning

🤦‍♂️ Ah my bad there. 

@tesk9 Requested review for realsies. 

Additional thought, should we try to encode the `whitespace: pre` in the interface somehow? Maybe have `Block.view` return an opaque type that only the `Block` module can assemble into a `whitespace: pre` paragraph?

tesk9:
>  Maybe have Block.view return an opaque type that only the Block module can assemble into a whitespace: pre paragraph?

You and your team have more context than I do, but I think some of the current uses/upcoming uses might be more direct than this would allow? E.g. for guided tutorials, where the Curriculum team writes something like `<block>content</block>` or something like it?

I like the idea of encoding the `whitespace: pre` requirement somehow, though. Maybe in the docs? And in text on the example page?

tesk9 (approved):
Looks great, thanks!

tesk9:
Actually -- on the tutorials front... @micahhahn are you confident that this change won't break tutorials? Is it likely true that the tutorials are typically rendering the blocks in pre containers?

micahhahn:
> @micahhahn are you confident that this change won't break tutorials?

Yes! It won't break without a `pre` wrapper because we're just introducing extra break recommendations. But `pre` will be required to get the breaking to **only** happen on the `<wbr/>` tags. 

Regardless I'll give tutorials a once over as part of incorporating the noredink-ui bump

tesk9:
Oh yeah that makes sense! Great! 👍

PR #1399, opened 2023-05-24
test with elm-test in buck2

BrianHicks:
This tests the package with elm-test in buck2. It temporarily adds elm-test (the package) to Node's package set, just because it probably doesn't make sense to figure out how to grab binaries from Nix at the same time (but WAT-600 will do that.)

### DIFF
--- a/BUCK
+++ b/BUCK
@@ -86,6 +86,6 @@ genrule(
 
 prettier_diffs(
     name = "prettier_diffs",
-    srcs = glob(["**/*.md"]),
+    srcs = glob(["**/*.md", "script/**/*.js"]),
     prettier = "//:prettier",
 )
--- a/BUCK
+++ b/BUCK
@@ -14,6 +14,7 @@ elm_docs(
 elm_test(
     name = "elm_test",
     test_srcs = glob(["test/**/*.elm"]),
+    elm_test = ":elm_test_binary",
 )
 
 filegroup(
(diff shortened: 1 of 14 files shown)


BUCK
```diff
@@ -1,11 +1,20 @@
...
     elm_json = "elm.json",
     src = "src",
+    tests = [
+        ":elm_test",
+    ],
+)
+
+elm_test(
+    name = "elm_test",
+    test_srcs = glob(["test/**/*.elm"]),
```
caiquefigueiredo, line 16:
```suggestion
    test_srcs = glob(["tests/**/*.elm"]),
```

caiquefigueiredo (approved):

BrianHicks:
@NoRedInkBot r=caiquefigueiredo

BrianHicks:
yay, it worked! Ready for re-review @caiquefigueiredo!

script/puppeteer-tests.js
```diff
@@ -33,6 +46,12 @@ describe("UI tests", function () {
     server.close();
   });
 
+  const handlePageErrors = function (page) {
+    page.on("pageerror", (err) => {
```
caiquefigueiredo, line 50: should we exit with error when this happens?

caiquefigueiredo (approved):

script/puppeteer-tests.js
BrianHicks: I think it’s OK to not for now.

PR #1400, opened 2023-05-25
:wrench: Make the toggle button more flexible

tesk9:
## Context

Currently, the formatting tutorial features toggle buttons for setting capitalization, italicization, and quotation marks for selected text.

screenshot of formatting tutorial toggle buttons

Because the toggle button API differed from the rest of the Button api, and therefore didn't support adding an explicit aria-label, the accessible names of these buttons are literally "CAPS", "I", and "“ ”" -- not helpful and not clear!

This PR marks the current toggle button implementation as deprecated and adds a new Button attribute that fits in to the rest of the nice-and-flexible Button API. This way, it will be straightforward to replace the accessible names for these buttons.

Fixes A11-885 partially

image

(no change intended in the above screenshot. Including just in case you see something I missed!)

image

@NoRedInk/design 

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -756,12 +756,7 @@ renderButton : ButtonOrLink msg -> Html msg
 renderButton ((ButtonOrLink config) as button_) =
     Nri.Ui.styled Html.button
         (styledName "customButton")
-        [ buttonStyles config
-        , Css.pseudoClass "focus-visible"
-            [ Css.outline3 (Css.px 2) Css.solid Css.transparent
-            , FocusRing.boxShadows []
-            ]
-        ]
+        (buttonStyles config)
         (ClickableAttributes.toButtonAttributes config.clickableAttributes
             { disabled = isDisabled config.state }
             ++ Attributes.class FocusRing.customClass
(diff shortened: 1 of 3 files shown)


src/Nri/Ui/Button/V10.elm
```diff
@@ -926,15 +932,51 @@ buttonStyles :
...
+                        , Css.backgroundColor Colors.glacier
+                        , Css.borderBottomWidth (Css.px 1)
+                        ]
+                    , secondaryColors
+                    )
+
+                Just False ->
+                    ( []
+                    , Css.hover [ Css.color Colors.navy ]
+                    , secondaryColors
```
charlieshanley, line 956: In future, `toggleButtonPressed` could possibly be made to play nicely with themes other than secondary (well, primary/danger/premium might be harder than the secondary/tertiary themes). But no need to do that now.

tesk9: Agreed! I was thinking about doing it, but then it seemed like since Design hadn't asked for it, it was unlikely to be worth it. (We also don't actually have very many toggle buttons in the app to start with! I think just formatting questions and grade selection)

charlieshanley (approved):
Fantastic! 

giphy - 2023-05-25T161134 344

PR #1403, opened 2023-05-30
[PHX-870] Add new characters to `CharacterIcon.V2`

micahhahn:
I'm creating `CharacterIcon.V2` so that we can leave V1 intact as we migrate. 

This PR became large enough without the SVGs for the "headshots", so those will come in a later PR.  I want to get this full height ones merged so we can start incorporating them into scaffolding.

Eventually we will have a celebration full size SVG as well to use instead of the celebration video for reduced motion - but design still needs to pick a frame for that. 

Given that we (will eventually) have a handful of states for each character (supportive / instruction / celebrating / headshot) I've tried to make the SVG code re-use the palettes and layout as much as possible to save on asset sizes.  I also found some opportunities to basically eliminate custom `clip-path` elements by using masking with existing elements (see usage of `maskStyle`). 

## Context

https://linear.app/noredink/issue/PHX-870/add-new-characters-to-noredink-ui

| Theme | Lindy | Sal | Red |
| --- | --- | --- | --- |
| Instructive | image | image | image |
| Supportive | image | image | image |

        - an accurate preview
        - valid sample code
        - correct keyboard behavior


charbelrami (approved):
Looks good!

PR #1404, opened 2023-06-01
correct instructions for testing individual components

BrianHicks:
This just updates some instructions that were wrong after a recent change. Tiny!

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Run tests with
 - `shake test`
 - `elm-test`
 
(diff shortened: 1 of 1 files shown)


tesk9 (commented):
Just to clarify, now the only way to run the tests locally without percy is to not have the percy credentials set up locally?

BrianHicks:
Oh, is that a thing that's causing you problems? Is that common on Accessibilibats? You could probably run `env -u PERCY_TOKEN script/puppeteer-test.sh` to avoid sending screenshots to Percy.

tesk9:
Not causing me problems as such! Just wanted to be sure I understood the change/intent of the change.

tesk9 (approved):
Thanks for updating!

PR #1405, opened 2023-06-12
A11-2653, A11-2664, A11-2669, A11-2670, add tests for Accordion, Panel, SegmentedControl, and Switch

charbelrami:
This adds tests for Accordion, Panel, SegmentedControl, and Switch.

### DIFF
--- a/tests/Spec/Nri/Ui/Panel.elm
+++ b/tests/Spec/Nri/Ui/Panel.elm
@@ -46,12 +46,14 @@ init =
 
 
 type Msg
-    = Toggle Bool
+    = NoOp
 
 
 update : Msg -> Model -> Model
 update msg state =
-    state
+    case msg of
+        NoOp ->
+            state
 
 
 viewMarkdown : Model -> Html Msg
--- a/tests/Spec/Nri/Ui/Panel.elm
+++ b/tests/Spec/Nri/Ui/Panel.elm
@@ -3,7 +3,6 @@ module Spec.Nri.Ui.Panel exposing (..)
 import Html
 import Html.Styled exposing (..)
 import Nri.Ui.Panel.V1 as Panel
-import ProgramTest exposing (..)
 import Test exposing (..)
 import Test.Html.Query as Query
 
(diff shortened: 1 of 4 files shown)


tests/Spec/Nri/Ui/Panel.elm
```diff
@@ -0,0 +1,94 @@
...
+rendersContent : List Test
+rendersContent =
+    [ test "renders markdown correctly" <|
+        \() ->
+            programMarkdown
+                |> ensureView
+                    (Query.contains
+                        [ Html.em [] [ Html.text "italic" ], Html.strong [] [ Html.text "bold" ] ]
+                    )
+                |> done
```
tesk9, line 27: I feel like this test isn't entirely self-contained -- you have to read more of the file and set up to understand what this test and the one after it are actually asserting.

I think inlining the setup, like so:
```
 \() ->
        ProgramTest.createSandbox
        { init = init
        , update = update
        , view =  
            \_ -> 
                 Panel.view
                    [ Panel.header "Header"
                    , Panel.markdown "_italic_ **bold**"
                    ]
                  |> toUnstyled
        }
                |> ProgramTest.start ()
                |> ensureView
                    (Query.contains
                        [ Html.em [] [ Html.text "italic" ], Html.strong [] [ Html.text "bold" ] ]
                    )
                |> done
```

Might help.

But really I think the biggest thing that might help is to _not_ use ProgramTest for these tests. Panel is not stateful, so I don't think we're getting a lot of extra value from ProgramTest, and it does add some boilerplate to set up.

```
test "renders markdown correctly" <|
        \() ->
            Panel.view
                [ Panel.header "Header"
                , Panel.markdown "_italic_ **bold**"
                ]
                |> toUnstyled
                |> Query.fromHtml
                |> Query.contains
                    [ Html.em [] [ Html.text "italic" ]
                    , Html.strong [] [ Html.text "bold" ]
                    ]
```

tests/Spec/Nri/Ui/SegmentedControl.elm
```diff
@@ -0,0 +1,248 @@
...
+ensurePanelDisplayed word testContext =
+    testContext
+        |> ensureView
+            (Query.find [ Selector.attribute Role.tabPanel, Selector.style "display" "block" ]
+                >> Query.has [ Selector.text word ]
+            )
+
+
+ensureOnlyOnePanelDisplayed : List String -> TestContext -> TestContext
+ensureOnlyOnePanelDisplayed panels testContext =
```
tesk9, line 131: I wonder if we could adjust this helper so that it asserts that the specific panels that we pass in are display:none with one panel selected?

Maybe:
```
ensureOnlyOnePanelDisplayed : String -> List String -> TestContext -> TestContext
ensureOnlyOnePanelDisplayed selectedPanel panels testContext =
```

? 

Otherwise, it seems a little odd to me to pass in a list of strings, when what we're using the list for is just the length, you know?

tests/Spec/Nri/Ui/Switch.elm
```diff
@@ -0,0 +1,108 @@
...
+        Toggle selected ->
+            let
+                selected_ =
+                    if selected then
+                        False
+
+                    else
+                        True
+            in
+            { state | selected = selected_ }
```
tesk9, line 67: Maybe `{state | selected = selected}`?

tests/Spec/Nri/Ui/Switch.elm
```diff
@@ -0,0 +1,108 @@
...
+import ProgramTest exposing (..)
+import Test exposing (..)
+import Test.Html.Selector as Selector
+
+
+spec : Test
+spec =
+    describe "Nri.Ui.Switch.V2"
+        [ describe "'switch' role" hasCorrectRole
+        , describe "has aria-disabled=true when disabled" hasCorrectAriaDisabled
```
tesk9, line 16: So these tests don't actually test state at all, which is fine with me, but I think if we're not testing state, we don't need to use ProgramTest -- might as well keep things simpler if we can, you know?

tesk9 (commented):
Looking great! Thanks for improving test coverage!

charbelrami:
Thank you, @tesk9 !

tesk9 (approved):
Thanks!

PR #1407, opened 2023-06-12
use flakes in Nix

BrianHicks:
This switchs us from using shell.nix to a flake. That should mean faster and more repeatable evaluation, as well as making it easier for the Buck 2 experiment to source packages so we can avoid adding (more) stuff to npm.

Whatever reviewer reviews this, please test your normal workflows here with both `nix shell` and `nix-shell`. If you don't have flakes enabled locally, it'd also be great if you could make sure that your `use nix` direnv workflow still works.

Fixes WAT-602


BrianHicks:
@NoRedInk/team-accessibilibats-a11ybats are y'all OK with this change? This should hypothetically be a nice upgrade, but it may be a hassle if you're using an older version of Nix.

PR #1409, opened 2023-06-14
Update pull_request_template.md with new designer pingage info

bendansby:
PR template update, see diff.

Counterpart to https://github.com/NoRedInk/NoRedInk/pull/44429

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -36,8 +36,12 @@ Please link to any relevant context and stories.
 
 ## :framed_picture: What does this change look like?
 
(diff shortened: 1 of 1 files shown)


tesk9 (commented):
I fear that you all might get pinged if people make WIP PRs and then edit the PR after the fact 😅. Maybe worth removing the `@`? If you're not worried about it, though, I'm not worried about it. It's your mentions after all!

bendansby:
Oh whoops, didn't realize that would create an actual @ mention. In fact we all got the ping from your draft PR on the monolith since I similarly updated that template! Will remove the ats for both.

tesk9 (approved):
Great, thank you!

PR #1412, opened 2023-06-21
Growth/minimalist tabs

ap-nri:
Please use the template that's relevant for your situation and delete the other templates. If this is just a noredink-ui repo doc change, you don't need to follow a template.

# :star2: Adding a new component

## About the component

We are adding a minimal Tab component with an understated view and NO CUSTOMIZATION if you want customizations go to Tabs please and thank you

This is to support the redesign of the teacher dashboard, specifically the tabbed interface of the class cards separating assignments by their Grading/Last Due/Up Next status as seen in the below design from figma
image

[full figma design](https://www.figma.com/file/CRqP1UX0J8lXH4GlHzu0mR/Growth-(2023%2B)?type=design&node-id=342-2383&t=FYSOdxPhU4W5MLvh-0) cc @NoRedInk/design 

### normal

image

### highlighting spacing

image

## zoomed way in to show that text ain't move when you change tabs

https://github.com/NoRedInk/noredink-ui/assets/75996561/7d7525c1-938a-45bb-b334-ea20b8323717

https://github.com/NoRedInk/noredink-ui/assets/75996561/17c086b3-881c-4220-974f-2ff11a0383f4

## Component completion checklist

- [x] I've gone through the relevant sections of the [Development Accessibility guide](https://paper.dropbox.com/doc/Accessibility-guide-4-Development--BiIVdijSaoijjOuhz3iTCJJ1Ag-rGoHpC91pFg3zTrYpvOCQ) with this component in mind
- [x] Component has clear documentation
- [x] Component is in the Component Catalog
    - [x] Component is categorized reasonably (see [Category](https://github.com/NoRedInk/noredink-ui/blob/master/component-catalog-app/Category.elm) for all the currently available categories). The component can be in multiple categories, if appropriate.
    - [x] Component has a representative preview for the Component Catalog cards (bonus points for making it delightful!)
    - [x] Component has a customizable example. Aim for having _every_ possible supported version of the component displayable through the configuration on this page. (Protip: This is handy for testing expected behavior!)
      - It helps that it is not in fact customizable.
    - [x] The customizable example produces sample code
    - [x] The component's example page includes keyboard behavior, if any
    - [x] The component's example page includes guidance around how to use the component, if necessary
- [x] Component is tested
[...]

Issue #1330, opened 2023-03-24
Enforce 20px of space below tabs


bendansby:
Perhaps with 20px of top padding on tab panel?

We'll then want to zero out any custom padding/margin around the site.

bendansby:
This was done for TabsMinimal

PR #1414, opened 2023-06-22
Tessa/minimal tab suggestions

tesk9:
Suggestions for https://github.com/NoRedInk/noredink-ui/pull/1412 and the minimalist tab example on the component catalog.

Mostly, I want the example to be more minimal! So I've cut out the Panel pangrams and made the `tabHtml` example more closely matched to the intended use of `tabHtml`.

I've also fixed some of the spacing in the example code by using `Code` helpers (I know the Tabs component that we started out copying doesn't provide great examples -- sorry about that!

## After

Screen Shot 2023-06-22 at 5 04 16 PM
Screen Shot 2023-06-22 at 5 04 12 PM

### DIFF
--- a/component-catalog/src/Code.elm
+++ b/component-catalog/src/Code.elm
@@ -5,9 +5,10 @@ module Code exposing
     , commentInline
     , list, listMultiline
     , tuple
+    , pipelineMultiline
     , record, recordMultiline
     , newlineWithIndent, newlines
-    , withParens
+    , withParens, withParensMultiline
     , anonymousFunction, always
     , caseExpression
     , browserElement, unstyledView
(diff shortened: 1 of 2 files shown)


bendansby:
I wonder if we should rename the component to TabsMinimal so they're grouped in various alphabetical situations?

tesk9:
Oh dang! I love that idea! 😍 (Will leave to @ap-nri to actually implement though, if he's on board with the change)

bendansby:
I'd also like to do this https://github.com/NoRedInk/noredink-ui/issues/1330 for TabsMinimal as well (except 15px based on Ravi's design) so we have that standardized off the bat.

ap-nri:
yes and yes and yes I love all of these you are strong and wise collaborators tysm @bendansby and @tesk9

PR #1415, opened 2023-06-26
Growth/un dogfood id generator

ap-nri:

## Context

Last hackday I created Nri.Ui.Html.Attributes.V2.safeId and safeIdWithPrefix. This was fine except I used it internally in a lot of our components. This broke way too many cypress tests and also verifiably broke the Guided Drafts intro tour. In order to un-block adoption of newer NRI-UI I have done three things in this PR:

1. stop using this ID generator anywhere in the UI and just exports it for gradual adoption here and in the monolith
1. lowercase the generated ID since most of our tests assume that will happen
1. only add the "id-" prefix when the leading character of the ID is not an ASCII letter

The monolith revert PR in question: https://github.com/NoRedInk/NoRedInk/pull/44536

No visual changes

* there are no visual or API changes to any components

### DIFF
--- a/src/Nri/Ui/Checkbox/V7.elm
+++ b/src/Nri/Ui/Checkbox/V7.elm
@@ -65,6 +65,7 @@ import Nri.Ui.FocusRing.V1 as FocusRing
 import Nri.Ui.Fonts.V1 as Fonts
 import Nri.Ui.Html.Attributes.V2 as Extra
 import Nri.Ui.Svg.V1 exposing (Svg)
+import Nri.Ui.Util as Util
 
 
 {-| This disables the input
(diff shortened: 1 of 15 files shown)


ap-nri:
I do not understand how it can fail to build Accessible HTML

-- PROBLEM BUILDING DEPENDENCIES -----------------------------------------------

I ran into a compilation error when trying to build the following package:

tesk9/accessible-html-with-css 5.1.0

This probably means it has package constraints that are too wide. It may be
possible to tweak your elm.json to avoid the root problem as a stopgap. Head
over to https://elm-lang.org/community to get help figuring out how to take this
path!

Note: To help with the root problem, please report this to the package author
along with the following information:

elm/core 1.0.5
elm/html 1.0.0
elm/json 1.1.3
rtfeldman/elm-css 18.0.0

tesk9:
I don't think this is necessary anymore. Think we got it allll figured out!

PR #1425, opened 2023-07-14
Remove internal padding

bendansby:
Removes redundant padding for the content of Header

image

### DIFF
--- a/src/Nri/Ui/Header/V1.elm
+++ b/src/Nri/Ui/Header/V1.elm
@@ -173,7 +173,7 @@ view attrs { breadCrumbs, isCurrentRoute } =
         ]
         [ Html.div
             (css
-                [ Spacing.centeredContent
+                [ Spacing.centeredContentWithCustomWidth config.pageWidth
                 , Css.alignItems Css.center
                 , Css.displayFlex
                 , Media.withMedia [ MediaQuery.mobile ] [ Css.flexDirection Css.column ]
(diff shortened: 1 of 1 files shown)


bendansby:
No particularly urgency here, just a re-ping @tesk9

tesk9 (commented):
Oopsies!  Thanks for the re-ping 🙏 

Yes, I think I meant to ask about/follow-up on the `pageWidth` getting dropped. 

So right now, I can set a custom page width and the header's contents won't expand past that point:
image

On this branch, the change to `centeredContent` means I can't. I think we want to use `Spacing.centeredContentWithCustomWidth config.pageWidth` instead, unless you are intentionally removing this functionality?

bendansby:
I don't think we want that content area to ever have a custom width, so I think that's good to drop. Curious if you suspect otherwise though!

tesk9:
I know we use ` Header.customPageWidth (Css.px 1350)` on `Nri.Writing.Grading.Layout`/the writing layouts -- I think mostly because of inline comments for QWs and GDs? That's my only hesitation. (If we're dropping functionality, though, I'd also want to note it in the changelog and docs, and make a story for removing the custom width in the next version of Heading, so we don't lose track of it.)

bendansby:
Ohh gotcha, OK, made that change (ngl I bet my eyes originally just glazed over all those camel humps and just chose the simplest one)

tesk9 (approved):
Thank you!

PR #1438, opened 2023-07-27
remove buck2

BrianHicks:
Well, this was a fun experiment but for reasons documented in Paper we won't be moving forward. Maybe we'll see you again in the future, Buck!


tesk9 (approved):
README change looks great!

The rest looks like so many deletions! 
Screen Shot 2023-07-27 at 3 16 27 PM
Love to see it!

PR #1443, opened 2023-08-03
Block: add labelState for label fade-out

omnibs:
Adds an explicit state for fading out `Block` and necessarily `Mark` too. Creates `Mark.V3`.

We used to rely on missing measurements as a way to imply a fade-out is desired, and this caused unmeasured labels in the catalog page for `Block` to fade-out state.

The change also helps our app stop relying on failed measurements to detect when a label should fade out, which was triggering errors unnecessarily.

```

    Nri.Ui.Mark.V3


---- Nri.Ui.Block.V5 - MINOR ----

    Added:
        type LabelState  = Visible | FadeOut
        labelState : Nri.Ui.Block.V5.LabelState -> Nri.Ui.Block.V5.Attribute msg
```

https://github.com/NoRedInk/noredink-ui/assets/1060410/ce58044c-27cf-4c6f-9dfe-eb71e8d98c99

ps: not requesting design approval bc this exact look and feel is already in prod.

        - an accurate preview
        - valid sample code
        - correct keyboard behavior
- [ ] Changes to the component are tested/the new version of the component is tested
  - I'm not sure how to test the presence of an animation
  - `Mark` was not tested before

### DIFF
--- a/component-catalog/src/Examples/Block.elm
+++ b/component-catalog/src/Examples/Block.elm
@@ -486,6 +486,24 @@ initControl =
             )
         |> ControlExtra.optionalListItem "labelId"
             (CommonControls.string ( Code.fromModule moduleName "labelId", Block.labelId ) fruitId)
+        |> ControlExtra.optionalListItem "labelState"
+            (Control.choice
+                [ ( "Visible"
+                  , Control.value
+                        ( Code.fromModule moduleName "labelState "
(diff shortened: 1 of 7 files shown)


tesk9 (approved):
Love it! So great! The example is so easy to work with now!

Also, the animations look so good! 😍 I love them.

Thank you for all these great improvements.

PR #1450, opened 2023-08-07
Link to ATN guide

tesk9:
Fixes A11-3324

Links to ATN guide from:
- Balloon 
- Loading 
- Message 
- Modal 
- Page
- RadioButton
- Select
- TextArea
- TextInput

Also, moves the `Fonts` background info to the new "About" section.

Screen Shot 2023-08-07 at 5 29 28 PM
Screen Shot 2023-08-07 at 5 31 17 PM


bendansby:
I wonder if this should be combined with the usage tables like we have for tooltip into an implementation section?

bendansby:
I think having every h2-headed section in a container like we have for keyboard support would add some structure to the page.

charbelrami (changes requested):
There are nested `p` elements. 
nested p elements in an html code

It would be nice to enclose the `h2` and `p` elements within a `section` element for better landmark navigation.

tesk9:
Screen Shot 2023-08-10 at 4 39 20 PM
Screen Shot 2023-08-10 at 4 39 32 PM

It's going to be too much to put all the content in containers in this PR, but @bendansby is this along the lines of what you wanted?

tesk9:
> I wonder if this should be combined with the usage tables like we have for tooltip into an implementation section?

This is an interesting idea! I think it's a bigger question than this PR can tackle. I'm adding to bat chat

bendansby:
That's nicer, thank you!

bendansby (approved):

charbelrami (approved):
Thanks! Looks great!

PR #1456, opened 2023-08-08
Tooltip text styles

bendansby:
- Remove text decoration declaration so that it can be set as needed (otherwise the default always wins the specificity war)
- Set border color to white so clickable texts don't show with a blue underline
- Reduce body copy to 15px

image


charbelrami (approved):
Looks good!

PR #1458, opened 2023-08-10
[PHX-880] Support specifying the blank block width in characters

micahhahn:

## Context

https://linear.app/noredink/issue/PHX-880/support-variable-length-blanks-in-tutorials

Discussions with design and curriculum have revealed that having 3 sizes for blank widths (`SingleCharacter`, `ShortWordPhrase`, `LongWordPhrase`) is not sufficient for curriculum's needs.  [See Slack Thread](https://noredink.slack.com/archives/C02TBGTURFX/p1691528956508349).  Instead, we want to offer them dynamic blanks widths that are approximately as wide as the text it covers. For this purpose this PR adds a new `CharacterCount` width (which is a minor update) with a long term goal of removing the other `BlankLength` values in the next major release.

a11y concerns about communicating more information to sighted users with the visible width are going to be handled by curriculum when addressing the blank (e.g. "which **character** should go in the blank?" or "which **phrase**"... etc). 

image

        - an accurate preview
        - valid sample code
        - correct keyboard behavior


micahhahn:
Hey @tesk9, it seems to me that the discussion has settled on [the Slack thread](https://noredink.slack.com/archives/C02TBGTURFX/p1691528956508349).   I still think this PR is the right approach as it is the most flexible.  If other usages of blanks want a specific set of 5 or so sizes, they can just choose to fix them at whatever width they want - no need to push those decisions into the API.

src/Nri/Ui/Block/V5.elm
```diff
@@ -284,6 +284,7 @@ type BlankLength
     = SingleCharacter
     | ShortWordPhrase
     | LongWordPhrase
+    | CharacterCount Int
```
tesk9, line 287: This addition causes a major bump:

```
---- Nri.Ui.Block.V5 - MAJOR ----

    Changed:
      - type BlankLength  = SingleCharacter | ShortWordPhrase | LongWordPhrase
      + type BlankLength
            = SingleCharacter
            | ShortWordPhrase
            | LongWordPhrase
            | CharacterCount Basics.Int

```

This is one of the reasons why in most noredink-ui components, we don't expose types directly, and instead will do something like:

```
type Customization = ThemeA | ThemeB

themeA : Attribute msg
themeA =
    Attribute (\config -> {config | theme = ThemeA})
```

tesk9 (changes requested):
I've made a suggestion PR over here https://github.com/NoRedInk/noredink-ui/pull/1466, primarily around ensuring that the Component Catalog example stays as usable as possible.

This can't be merged as a major bump -- not sure how you want to handle that. You could make a new version, and undo the changes the V5. That's probably simplest.

To confirm: are design, the writing team, and curriculum all aware that the current width definitions will change in the new version?

Could you also please update the documentation?

And finally, a small implementation problem that I think needs fixing:

image

I think we need to limit the max width to the viewport width or the container width or to a set value. The blank doesn't split across lines, so its maxwidth should be the line width in practice. Although it does look pretty bad in practice.

src/Nri/Ui/Block/V5.elm
micahhahn: If my hand is forced to do a major version update, I will just remove the `SingleCharacter`, `ShortWordPhrase`, and `LongWordPhrase` options to clarify the API

src/Nri/Ui/Block/V6.elm
```diff
@@ -16,6 +16,8 @@ module Nri.Ui.Block.V6 exposing
 
 {-| Changes from V5:
 
+    - Remove `wordWithId` and `blankWithId` as we no longer are trying to point to words or blanks with the question box.
```
tesk9, line 19: OHHH 💡 nice, thanks

tesk9: Heads up, though, there are uses still:
image

https://github.com/NoRedInk/NoRedInk/blob/master/lib/scaffolding/elm/src/Nri/Scaffolding/Sentence/Render.elm#L665

micahhahn: Yup, but we don't use the IDs for anything (those usages should have been pulled when we switched to the `CharacterStage`).

micahhahn:
> To confirm: are design, the writing team, and curriculum all aware that the current width definitions will change in the new version?

Yes! (And also the size is determined by the consumer of the API anyways - so they might not necessarily change)

component-catalog/src/Examples/Block.elm
```diff
@@ -406,6 +411,104 @@ example =
...
+                            ]
+                  }
+                ]
+            , Heading.h2
+                [ Heading.plaintext "Blank width guidance"
+                , Heading.css [ Css.marginTop Spacing.verticalSpacerPx ]
+                ]
+            , Text.smallBody [ Text.markdown "The accessible name of all blanks, regardless of character width used, is \"blank.\"" ]
+            , Text.smallBody [ Text.markdown "If we're looking for a specific length of content to put in the blank, that _must_ be communicated elsewhere on the page to provide an equitable experience." ]
+            , Text.smallBody [ Text.markdown "The `characterWidth` parameter uses the number of characters expected in the blank to calculate a rough monospace-based width that visually looks like it _could_ match." ]
```
tesk9, line 455: I LOVE having this content moved up above the table! Such a great change -- way more readable, and totally fixes a _very_ funky use of the Table 😅

tesk9 (commented):
I'm so happy with this! I love the API simplifications. This is so much easier to understand!

I pushed a couple of commits, if you could please take a look at them and make sure they're okay. The last one you might want to revert -- I just wanted to make a potential problem that I'm seeing a little bit more obvious to get your thoughts on it. 

So what I'm seeing is that the `Block.blank` with emphasis ends up too tall:
image

Maybe I'm missing something? Or we're just not expecting this use case?

---

Separately, small API concern that you should totally feel free to ignore: I wonder if `characterWidth` reads like you're setting the width of a character, rather than passing in a representation of # of characters? I don't feel strongly about this at all -- just wanted to flag for your consideration.

---

Thanks for all this awesome work!

micahhahn:
Hey @tesk9 

> Separately, small API concern that you should totally feel free to ignore: I wonder if characterWidth reads like you're setting the width of a character, rather than passing in a representation of # of characters? I don't feel strongly about this at all -- just wanted to flag for your consideration.

I think you're right! Name is hard... I've changed the property to `widthInChars` - I think that's clearer? 

> So what I'm seeing is that the Block.blank with emphasis ends up too tall:

You're right! I did a bit of refactoring that made things simpler and fixed this issue if you would kindly take a look!

micahhahn:
Oh also Juliano has some small changes he wants to make to `Block` and `Mark` (just pulling out `labelState` to remove animations from the noredink-ui) - I think it makes sense for me to just incorporate them as part of this major version bump instead of having him do another one after me.

micahhahn:
Also, big thought here, but how would you feel about just removing the `plaintext` and empty `view` cases all together and making `content` required as an explicit argument? 

The use cases for not using `content` are getting smaller and smaller (no markdown, no interleaved blanks, no blank widths, etc...) and it's a little bulky to have to always wrap the contents in an extra function call. I think originally a call to `plaintext` generated better html, but I'm not so sure that's even the case anymore as both `plaintext` and `phrase` go through `parseString`

tesk9:
> Also, big thought here, but how would you feel about just removing the plaintext and empty view cases all together and making content required as an explicit argument?

I think this is a good idea! More and more, it seems like `Block.view []` is not how the Block is used in practice, and it makes the API more confusing to have multiple ways of constructing blanks. And then when you're using `content`, like you always end up doing, it's a real confusion of brackets 😖

(I do think if you do this, you might have more upgrade work to do on behalf of the Writing team as well as on behalf of the Phoenix team, but if you're up for it, I think this is a great idea!)

micahhahn:
> And then when you're using content, like you always end up doing, it's a real confusion of brackets 😖

I'm going to save this for the near future as this variable length blank stuff is relatively high priority to unblock curriculum.  

FWIW, I also want to tweak `phrase` to be `phrase : String -> Content msg` so that usage will be much cleaner: 

```elm
Block.view
  [ Block.phrase "Hello world"
  , Block.blank { widthInChars = 1 }
  , Block.bold [ Block.phrase "says the man on the moon" ]
  ]
  [ Block.emphasize
  , Block.label "Is this a full sentence"
  ]
```

micahhahn:
@tesk9 Ready for final review!

tesk9 (approved):
Looks great! Thanks for all your work on this!

If you could add story links for "add your story links here OR just write this is not a new major version" in the PR description, that would be great! (Since we ended up with new major versions of things 😅)

Also, if you could please review the last commit I pushed (updating puppeteer tests) before you merge, that would be great too! 

Thanks!

micahhahn:
Looks great! Thanks for all the word reviewing this!

Issue #1454, opened 2023-08-08
Improve placement of checkbox guidance


bendansby:
Guidance should be coupled with the checkbox label in a vertically aligned group as in the [original PR](https://github.com/NoRedInk/noredink-ui/pull/1144)

![](https://user-images.githubusercontent.com/13528834/198643323-8b69fa2e-7e7a-4f95-81b5-63e64760db4d.png)

PR #1460, opened 2023-08-10
:wrench: Fix checkbox guidance alignment

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/1454

- adds new ClickableText.caption size for use with Text.caption-size text
- adjusts the Checkbox implementation to reposition the guidance appropriately

Screen Shot 2023-08-10 at 4 04 50 PM

I added a new section to the Checkbox example that always shows the guidance, because I feel like we've had trouble with alignment 🐛 s on the guidance that Percy could have caught.

Does this alignment look okay to you?

        - an accurate preview
        - valid sample code
        - correct keyboard behavior
  - this is not a new major version

### DIFF
--- a/component-catalog/src/Examples/Checkbox.elm
+++ b/component-catalog/src/Examples/Checkbox.elm
@@ -113,7 +113,7 @@ example =
                 [ Table.custom
                     { header = text "Attribute"
                     , view = .name >> text
-                    , width = Css.px 150
+                    , width = Css.pct 10
                     , cellStyles = always [ Css.padding2 (Css.px 14) (Css.px 7), Css.verticalAlign Css.middle ]
                     , sort = Nothing
                     }
@@ -128,7 +128,7 @@ example =
                                 [ Checkbox.id (safeIdWithPrefix "guidance-and-error-example" name)
                                 , attribute
                                 ]
-                    , width = Css.px 150
+                    , width = Css.pct 50
                     , cellStyles = always [ Css.padding2 (Css.px 14) (Css.px 7), Css.verticalAlign Css.middle ]
                     , sort = Nothing
                     }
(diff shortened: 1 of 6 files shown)


bendansby:
Oh, that's great, thanks for the quick fix! And for the new ClickableText size, I had that on my mental to-do list. No more [this](https://github.com/NoRedInk/NoRedInk/pull/45169/files#diff-b0cf00c7e131218a6e7e7fd496189a1e3993b9af60a89825561df42da2244f4dR329)!

It looks sliiightly misaligned sans ClickableText, but that could be attributable to the caprices of letterforms or oddly numbered pixels
Napkin 15 08-10-23, 3 19 16 PM

tesk9:
Okay! So I think the problem is that the ClickableText ends up making the entire guidance line 17.5px tall instead of 13px.

The extra height seems to be coming from two places:
- it uses line-height normal, which for 13px ends up (in chrome) at 16.5px
- appearsInline adds the border bottom, which adds 1px to the bottom

image

I don't want to end up in the CSS weeds of ClickableText, so I think what I'm going to do is try to get the plain `guidance` view's top alignment slightly higher, and not worry as much about the guidanceHtml alignment. (Especially since `guidance` is much more used.)

bendansby:
Awesome, thank you!

tesk9:
Okay! I did remove `line-height: normal` from the ClickableText, so hopefully that doesn't cause any issues! I improved the example for the inline appearance of ClickabelText, which will also help us identify problems sooner rather than later.

Screen Shot 2023-08-11 at 4 21 19 PM

--

For the checkbox icon, I think that the transparent border around the icon (which is there, as I recall, for high contrast mode) was causing the misalignment. I adjusted the label container up to account for the margin. Look okay?

Screen Shot 2023-08-11 at 4 21 30 PM

bendansby:
I like it, appreciate the fine-tuning!

mfonism (approved):
This is fantastic.

Thank you Tessa!

aligned-carla-delaney

PR #1464, opened 2023-08-14
:wrench: Remove checkbox input from Switch implementation

tesk9:
Fixes A11-3200

Changes the Switch component to not be backed by a checkbox input.

Also improves the Switch component catalog example:

image

        - an accurate preview
        - valid sample code
        - correct keyboard behavior
  - https://linear.app/noredink/issue/A11-3391/upgrade-to-switchv3

### DIFF
--- a/component-catalog/src/Examples/Switch.elm
+++ b/component-catalog/src/Examples/Switch.elm
@@ -10,11 +10,14 @@ import Accessibility.Styled.Key as Key
 import Category
 import Code
 import CommonControls
+import Css
 import Debug.Control as Control exposing (Control)
 import Debug.Control.Extra as ControlExtra
 import Debug.Control.View as ControlView
 import Example exposing (Example)
 import KeyboardSupport exposing (Key(..))
+import Nri.Ui.Heading.V3 as Heading
+import Nri.Ui.Spacing.V1 as Spacing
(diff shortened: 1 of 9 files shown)


BrianHicks (approved):
This looks fine, but how come? I thought it was preferable to use "real" DOM elements for interactive elements instead of adding a bunch of aria-thingies to a div? What makes this an exceptional case?

switch-coin-gif-by-millions

tesk9:
It's to fix https://linear.app/noredink/issue/A11-3200. If you play with a screenreader, you'll find that checkboxes will change state on you when aria-disabled is present and you've explicitly set the state. We prefer not to use use `disabled` since it makes elements disappear for SR users, and we think if the control is present but disabled, it's present for a reason, so should be apparent but non-interactive for SR users too.

BrianHicks:
Ah, ok. Thank you!On Aug 15, 2023, at 4:05 PM, Tessa ***@***.***> wrote:﻿
It's to fix https://linear.app/noredink/issue/A11-3200. If you play with a screenreader, you'll find that checkboxes will change state on you when aria-disabled is present and you've explicitly set the state. We prefer not to use use disabled since it makes elements disappear for SR users, and we think if the control is present but disabled, it's present for a reason, so should be apparent but non-interactive for SR users too.

PR #1468, opened 2023-08-18
Canonize sideNav width

bendansby:
Rather than rely on setting it per instance, add it to the component.

image


bendansby:
ping @tesk9

tesk9 (approved):
Looks good! Sorry for the slow response

PR #1477, opened 2023-08-22
Tooltip tail affixation

caseyWebb:

## Context

This PR makes `Nri.Tooltip` tails always affix to the trigger element. While the API remains the same, this is a breaking change with regards to visual behavior.

The changes are best displayed via video, see below: 

### V3 (Current)

https://github.com/NoRedInk/noredink-ui/assets/5419074/21dbf312-e947-4684-b1aa-44dde90d77c3

### V4 (New)

https://github.com/NoRedInk/noredink-ui/assets/5419074/0b7c5fb8-c348-49f9-b3a2-81509620d131

- [ ] I've gone through the relevant sections of the [Development Accessibility guide](https://paper.dropbox.com/doc/Accessibility-guide-4-Development--BiIVdijSaoijjOuhz3iTCJJ1Ag-rGoHpC91pFg3zTrYpvOCQ) with the changes I made to this component in mind
- [ ] Changes are clearly documented
    - [ ] Component docs include a changelog
    - [ ] Any new exposed functions or properties have docs
- [ ] Changes extend to the Component Catalog
    - [ ] The Component Catalog is updated to use the newest version, if appropriate
    - [ ] The Component Catalog example version number is updated, if appropriate
    - [ ] Any new customizations are available from the Component Catalog
    - [ ] The component example still has:
        - an accurate preview
        - valid sample code
        - correct keyboard behavior
- [ ] Changes to the component are tested/the new version of the component is tested
- [ ] Component API follows standard patterns in noredink-ui
- [ ] If this is a new major version of the component, our team has stories created to upgrade all instances of the old component. Here are links to the stories:

### DIFF
--- a/.direnv/nix-profile-pre452927.6ccc4a59c3f
+++ b/.direnv/nix-profile-pre452927.6ccc4a59c3f
@@ -1,0 +1,1 @@
+/nix/store/r52q0abiwslzbfyb8sx577yb20qy638r-nix-shell-env
--- a/.direnv/nix-profile-pre452927.6ccc4a59c3f.rc
+++ b/.direnv/nix-profile-pre452927.6ccc4a59c3f.rc
@@ -1,0 +1,1885 @@
+unset shellHook
+PATH=${PATH:-}
+nix_saved_PATH="$PATH"
+XDG_DATA_DIRS=${XDG_DATA_DIRS:-}
+nix_saved_XDG_DATA_DIRS="$XDG_DATA_DIRS"
+AR='ar'
+export AR
+AS='as'
+export AS
+BASH='/nix/store/f2fi9x1fnwly702yqdlsjhy46pcgz5vi-bash-5.2-p15/bin/bash'
+CC='clang'
+export CC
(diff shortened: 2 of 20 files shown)


tesk9:
@caseyWebb just to double-check, you're aware that if Kraken makes this change, Kraken will be responsible for updating _all_ tooltip uses to the new version? The upgrade path here seems a little challenging, so I just want to be sure the team has enough bandwidth to get everything done!

caseyWebb:
Yep, trust me it's a bit daunting 😅. We don't have _too_ many uses of alignStart/alignEnd, but afaict the upgrade path with involve checking each one so definitely thank you for raising early.

The upside is this isn't terribly high priority, and I can work asynchronously to upgrade existing ones before needing to merge this, and if it truly is a complete headache this isn't strictly necessary, just very nice to have. The other slight saving grace is that the custom offset is _most likely_ (I hope) accomplishing what this PR does automatically, so should generally involve setting it to zero (I think).

tesk9:
If you do end up going for it, I recommend testing your changes with `Nri.Writing.UITour` as early as you can. It's a weird/non-standard use of tooltips (and a pattern that shouldn't be replicated anywhere else), but it's the area where I think this change might be particularly tricky. I could be wrong, but that's my guess!

PR #1482, opened 2023-08-22
[PHX-1071] :wrench: Add `enabled`/`disabled` attributes to `RadioButtonDotless`

micahhahn:
I'm rather frustrated in the corner that I was backed in to by Elm in this PR.  

The only way to do disablement in a a11y friendly manner is to use `aria-disabled` in combination with `preventDefault`.  However, we simply **cannot** use `preventDefault` in Elm without producing a `msg`.  Which is exactly the opposite behavior that we want in this case because we are *preventing* the user from taking an action.   This leads us to require a `NoOp` message for **every** consumer of `RadioButtonDotless` if they want to take advantage of the disabled state.  

- https://github.com/elm/html/issues/96
- https://github.com/elm/html/issues/137

Judging by Evan's comment in the thread above, this won't be fixed on the Elm side anytime soon.

> That said, I'd like to keep it closed. I think it is fine to have a NoOp for now, and if you think of the scenarios you were running into, please open a new issue that describes why it would be helpful.

This doesn't bode well for our existing `RadioButton` and `Button` components - which similarly have broken disabled behavior. 

## Context

[PHX-1071 Add disabled style to RadioButtonDotless](https://linear.app/noredink/issue/PHX-1071/add-disabled-style-to-radiobuttondotless)

image

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/component-catalog/src/Examples/RadioButtonDotless.elm
+++ b/component-catalog/src/Examples/RadioButtonDotless.elm
@@ -132,12 +132,14 @@ sizeControl =
         , ( "large", Control.value ( "RadioButtonDotless.large", RadioButtonDotless.large ) )
         ]
 
-enablementControl : Control ( String, RadioButtonDotless.Attribute ControlSelection Msg)
+
+enablementControl : Control ( String, RadioButtonDotless.Attribute ControlSelection Msg )
 enablementControl =
-    Control.choice 
-        [ ("enabled", Control.value ( "RadioButtonDotless.enabled", RadioButtonDotless.enabled ))
-        , ("disabled", Control.value ("RadioButtonDotless.disabled", RadioButtonDotless.disabled))
+    Control.choice
+        [ ( "enabled", Control.value ( "RadioButtonDotless.enabled", RadioButtonDotless.enabled ) )
+        , ( "disabled", Control.value ( "RadioButtonDotless.disabled", RadioButtonDotless.disabled ) )
         ]
+
 
 moduleName : String
 moduleName =
--- a/component-catalog/src/Examples/RadioButtonDotless.elm
+++ b/component-catalog/src/Examples/RadioButtonDotless.elm
@@ -70,6 +70,7 @@ controlAttributes =
         |> ControlExtra.optionalListItem "textAlign" textAlignControl
         |> ControlExtra.optionalListItem "width" widthControl
         |> ControlExtra.optionalListItem "size" sizeControl
+        |> ControlExtra.optionalListItem "enablement" enablementControl
         |> ControlExtra.optionalListItem "containerCss"
             (Control.choice
                 [ ( "max-width with border"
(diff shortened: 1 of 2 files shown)


charbelrami:
Hey @micahhahn!

I noticed you're using the `disabled` attribute. Have you considered using `aria-disabled` instead? It's generally more accessible because `disabled` prevents focusable elements from receiving focus, which can be problematic for keyboard-only users.

When using `aria-disabled`, you would also need to disable the default behavior of the radio buttons programmatically. You can do this using `event.preventDefault()` for the `"click"` and `"keydown"` events.

Here's a quick [example](https://codesandbox.io/s/morning-morning-p5wm4k?file=/src/index.mjs).

micahhahn:
Hey @charbelrami 

I've been generally following the lead of the `RadioButton` component here, but the behavior around being disabled appears to be wrong... it still allows me to select an element with the keyboard if it is disabled: 

https://github.com/NoRedInk/noredink-ui/assets/12899207/15f8ed01-dd82-41e7-a477-b7f2e2142975

What's the "right" behavior for a radio button? Should the user be able to arrow through the buttons, even the disabled ones, but **not** able to select one that's disabled?

charbelrami:
Hey @micahhahn !

>  but the behavior around being disabled appears to be wrong

Yes, sorry, the behavior in the `RadioButton` component is wrong.

> What's the "right" behavior for a radio button? Should the user be able to arrow through the buttons, even the disabled ones, but not able to select one that's disabled?

Yes. I've written a quick [example](https://codesandbox.io/s/morning-morning-p5wm4k?file=/src/index.mjs). Using `preventDefault` for the `"click"` event seems sufficient.

micahhahn:
@charbelrami, alright that makes sense. Let's make this better than `RadioButton`! 

Due to the only [method we have in Elm to set `preventDefault`](https://package.elm-lang.org/packages/elm/html/latest/Html-Events#custom) I would need to have a `NoOp` message passed in with the `disable` attribute - which is awkward.

Simple not setting an event handler (in combination with the radio button not being visible) seems to do the trick.  At least the correct selection is preserved in Elm-land. Can you check if the solution I've pushed works?

charbelrami:
Thank you @micahhahn!

It doesn't work as intended because the `checked` attribute is still updated in the DOM, so the screen reader announces the checkbox as "dimmed selected".

https://github.com/NoRedInk/noredink-ui/assets/5647344/16c8eff9-00ad-4127-8cf2-31f218b44403

micahhahn:
@charbelrami After much investigation, it looks like we will just have to force the consumer to provide a `NoOp` message to get disabled states. Ready for review again.

charbelrami (approved):
Looks great! Thank you

caseyWebb:
I'm confused; I understand why Charbel's demo needed preventDefault since it was using a global event listener, but is the omittance of the normal event handler (i.e. `AttributesExtra.none` instead of `Events.onClick ...`) not sufficient to achieve the no-op behavior? I could be missing something, but playing with it I'm not really able to tell a difference in behavior.

Issue #1473, opened 2023-08-21
Accordion custom icon (example?) is broken


bendansby:
<img width="1552" alt="image" src="https://github.com/NoRedInk/noredink-ui/assets/13528834/0edff018-e487-4b0b-80b0-dd6bb1be60f4">

PR #1485, opened 2023-08-23
Fix custom accordion icon example

tesk9:
Fixes https://github.com/NoRedInk/noredink-ui/issues/1473

Screen Shot 2023-08-23 at 10 09 42 AM

(Only touches the component catalog example.)


bendansby (approved):
I like the non-enormous aspects of this 👌

PR #1487, opened 2023-08-23
Usage Examples

tesk9:
Experimental PR adding new routing for usage examples. The point of these new routes is to demonstrate how components work _together_.

For example:
- demonstrate a clickable card that also contains a link or button and an explanatory tooltip
- demonstrate forms with many inputs and the validation pattern that we want

### Usage Examples
The new usage examples are linked in their own section in the catalog:
Screen Shot 2023-09-01 at 5 10 08 PM

### Clickable Card with Tooltip
Previously, the Tooltip example included a button-within-a-button example of clicking on a parent container. This PR reverted the tooltip example to be basic, and added links to the new "Clickable Card with Tooltip" example usage page.
Screen Shot 2023-09-01 at 5 10 26 PM

Screen Shot 2023-09-01 at 5 10 15 PM

### Form
I'd like to add all the inputs to this form example. But as a starting point, I've added Select and TextInput on a single line:

Screen Shot 2023-09-01 at 5 10 01 PM


bendansby:
What do you think about splitting the CC into tabs?

tesk9:
Oh! That sounds great to me! I'd probably want to do it in a separate PR though

tesk9:
Partly just... is this an okay idea to move forward on? Is this about what you were thinking as far as how a form inputs example would work (but like, with more inputs?)

bendansby:
Yeah, having usage examples or molecules or whatever is definitely something we want, so this is a great step. Yeah for a form we'd want more inputs, a submit button, various forms of validation, etc.

bendansby (approved):

charbelrami (approved):
Looks good!

PR #1488, opened 2023-08-23
Implement new carousel types

ArthurJordao:

## Context

Implement new carousels types (previous/next and combined) using [w3c guided](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/)
This component doesn't have styles attached to it, so no visual changes.

        - an accurate preview
        - valid sample code
        - correct keyboard behavior
  - [Growth issue to migrate landing](https://linear.app/noredink/issue/GRO-1375/migrate-landing-page-carousel-to-nriuicarouselv2)
  - [Quokka issue for short response](https://linear.app/noredink/issue/WR-208/adapt-and-use-nriuicarousel-to-view-supporting-materials)

### DIFF
--- a/component-catalog/src/Examples/Carousel.elm
+++ b/component-catalog/src/Examples/Carousel.elm
@@ -107,12 +107,10 @@ controlStyles isSelected =
 
 
 type Msg
-    = FocusAndSelect { select : Int, focus : Maybe String }
-    | AnnounceAndSelect { select : Int, announce : String }
+    = Select { select : Int, announce : Maybe String, focus : Maybe String }
     | SelectTip { select : Tip, announce : String }
     | SelectTestimonial { select : Testimonial, focus : Maybe String }
-    | SelectPackageAndFocus { select : Package, focus : Maybe String }
-    | SelectPackageAndAnnounce { select : Package, announce : String }
+    | SelectPackage { select : Package, announce : Maybe String, focus : Maybe String }
     | Focused (Result Dom.Error ())
     | SetSettings (Control Settings)
 
@@ -120,8 +118,17 @@ type Msg
 update : Msg -> State -> ( State, Cmd Msg )
 update msg model =
     case msg of
-        FocusAndSelect { select, focus } ->
-            ( { model | selected = select }
+        Select { select, announce, focus } ->
+            ( { model
+                | selected = select
+                , announcement =
+                    case announce of
+                        Just announcement ->
+                            Just announcement
+
+                        Nothing ->
+                            model.announcement
+              }
             , focus
                 |> Maybe.map (Dom.focus >> Task.attempt Focused)
                 |> Maybe.withDefault Cmd.none
@@ -132,14 +139,6 @@ update msg model =
 
         SetSettings settings ->
             ( { model | settings = settings }, Cmd.none )
-
-        AnnounceAndSelect { select, announce } ->
-            ( { model
-                | selected = select
-                , announcement = Just announce
-              }
-            , Cmd.none
-            )
 
         SelectTip { select, announce } ->
             ( { model
@@ -156,19 +155,20 @@ update msg model =
                 |> Maybe.withDefault Cmd.none
             )
 
-        SelectPackageAndFocus { select, focus } ->
-            ( { model | package = select }
+        SelectPackage { select, announce, focus } ->
+            ( { model
+                | package = select
+                , announcement =
+                    case announce of
+                        Just announcement ->
+                            Just announcement
+
+                        Nothing ->
+                            model.announcement
+              }
             , focus
                 |> Maybe.map (Dom.focus >> Task.attempt Focused)
                 |> Maybe.withDefault Cmd.none
-            )
-
-        SelectPackageAndAnnounce { select, announce } ->
-            ( { model
-                | package = select
-                , announcement = Just announce
-              }
-            , Cmd.none
             )
 
 
@@ -360,7 +360,13 @@ viewCustomizableWithPreviousAndNextControls model =
                 , name = "Items"
                 , visibleLabelId = Nothing
                 , role = Carousel.Group
-                , announceAndSelect = AnnounceAndSelect
+                , announceAndSelect =
+                    \{ select, announce } ->
+                        Select
+                            { select = select
+                            , announce = Just announce
+                            , focus = Nothing
+                            }
                 }
     in
     ( Code.pipelineMultiline
@@ -424,8 +430,7 @@ viewCustomizableWithCombinedControls model =
                 , role = Carousel.Group
                 , name = "Items"
                 , visibleLabelId = Nothing
-                , focusAndSelect = FocusAndSelect
-                , announceAndSelect = AnnounceAndSelect
+                , select = Select
                 }
     in
     ( Code.pipelineMultiline
@@ -487,7 +492,13 @@ viewCustomizableWithTabControls model =
                 , role = Carousel.Group
                 , name = "Items"
                 , visibleLabelId = Nothing
-                , focusAndSelect = FocusAndSelect
+                , focusAndSelect =
+                    \{ select, focus } ->
+                        Select
+                            { select = select
+                            , announce = Nothing
+                            , focus = focus
+                            }
                 }
     in
     ( Code.pipelineMultiline
@@ -806,8 +817,7 @@ viewPackages selected =
                 , role = Carousel.Group
                 , name = "Packages"
                 , visibleLabelId = Nothing
-                , focusAndSelect = SelectPackageAndFocus
-                , announceAndSelect = SelectPackageAndAnnounce
+                , select = SelectPackage
                 }
     in
     Container.view
--- a/component-catalog/src/Examples/Carousel.elm
+++ b/component-catalog/src/Examples/Carousel.elm
@@ -409,7 +409,7 @@ viewCustomizableWithCombinedControls model =
         allItems =
             settings.items
                 |> indicesForItemCount
-                |> List.map toTabbedCarouselItem
+                |> List.map toCombinedCarouselItem
 
         { tabControls, slides, viewPreviousButton, viewNextButton, containerAttributes } =
             Carousel.viewWithCombinedControls
@@ -560,13 +560,46 @@ toTabbedCarouselItem :
         ( String
         , { id : Int
           , idString : String
+          , slideHtml : Html msg
+          , tabControlHtml : Html Never
+          }
+        )
+toTabbedCarouselItem id =
+    let
+        idString =
+            Attributes.safeIdWithPrefix "slide" <| String.fromInt id
+
+        humanizedId =
+            String.fromInt (id + 1)
+    in
+    ( Code.recordMultiline
+        [ ( "id", Code.int id )
+        , ( "idString", Code.string (String.fromInt id) )
+        , ( "tabControlHtml", "Html.text " ++ Code.string ("Slide " ++ humanizedId) )
+        , ( "slideHtml", "Html.text " ++ Code.string ("Contents for slide " ++ humanizedId) )
+        ]
+        3
+    , { id = id
+      , idString = idString
+      , tabControlHtml = Html.text ("Slide " ++ humanizedId)
+      , slideHtml = Html.text ("Contents for slide " ++ humanizedId)
+      }
+    )
+
+
+toCombinedCarouselItem :
+    Int
+    ->
+        ( String
+        , { id : Int
+          , idString : String
           , name : String
           , visibleLabelId : Maybe String
           , slideHtml : Html msg
           , tabControlHtml : Html Never
           }
         )
-toTabbedCarouselItem id =
+toCombinedCarouselItem id =
     let
         idString =
             Attributes.safeIdWithPrefix "slide" <| String.fromInt id
@@ -677,22 +710,16 @@ viewTestimonials selected =
                 , slides =
                     [ { id = GreatService
                       , idString = "great-service"
-                      , name = "Great Service"
-                      , visibleLabelId = Nothing
                       , slideHtml = text "Great service!"
                       , tabControlHtml = span Style.invisible [ text "Testimonial 1" ]
                       }
                     , { id = GreatProduct
                       , idString = "great-product"
-                      , name = "Great Product"
-                      , visibleLabelId = Nothing
                       , slideHtml = text "Great product!"
                       , tabControlHtml = span Style.invisible [ text "Testimonial 2" ]
                       }
                     , { id = GreatMission
                       , idString = "great-mission"
-                      , name = "Great Mission"
-                      , visibleLabelId = Nothing
                       , slideHtml = text "Great mission!"
                       , tabControlHtml = span Style.invisible [ text "Testimonial 3" ]
                       }
--- a/component-catalog/src/Examples/Carousel.elm
+++ b/component-catalog/src/Examples/Carousel.elm
@@ -421,8 +421,8 @@ viewCustomizableWithCombinedControls model =
             Carousel.viewWithCombinedControls
                 { selected = model.selected
                 , slides = List.map Tuple.second allItems
-                , tabControlStyles = \_ -> []
-                , tabControlListStyles = []
+                , tabStyles = \_ -> []
+                , tabListStyles = []
                 , previousButton =
                     { attributes = [], icon = UiIcon.arrowLeft, name = "Previous" }
                 , nextButton =
@@ -438,8 +438,8 @@ viewCustomizableWithCombinedControls model =
             ++ Code.record
                 [ ( "selected", Code.int model.selected )
                 , ( "slides", Code.listMultiline (List.map Tuple.first allItems) 2 )
-                , ( "tabControlStyles", "(\\_ -> [])" )
-                , ( "tabControlListStyles", Code.list [] )
+                , ( "tabStyles", "(\\_ -> [])" )
+                , ( "tabListStyles", Code.list [] )
                 , ( "previousButton"
                   , Code.recordMultiline
                         [ ( "name", Code.string "Previous" )
@@ -487,8 +487,8 @@ viewCustomizableWithTabControls model =
             Carousel.viewWithTabControls
                 { selected = model.selected
                 , slides = List.map Tuple.second allItems
-                , tabControlStyles = \_ -> []
-                , tabControlListStyles = []
+                , tabStyles = \_ -> []
+                , tabListStyles = []
                 , role = Carousel.Group
                 , name = "Items"
                 , visibleLabelId = Nothing
@@ -506,8 +506,8 @@ viewCustomizableWithTabControls model =
             ++ Code.record
                 [ ( "selected", Code.int model.selected )
                 , ( "slides", Code.listMultiline (List.map Tuple.first allItems) 2 )
-                , ( "tabControlStyles", "(\\_ -> [])" )
-                , ( "tabControlListStyles", Code.list [] )
+                , ( "tabStyles", "(\\_ -> [])" )
+                , ( "tabListStyles", Code.list [] )
                 , ( "role", Code.fromModule moduleName "Group" )
                 , ( "name", Code.string "Items" )
                 , ( "visibleLabelId", Code.maybe Nothing )
@@ -735,8 +735,8 @@ viewTestimonials selected =
                       , tabControlHtml = span Style.invisible [ text "Testimonial 3" ]
                       }
                     ]
-                , tabControlStyles = tabControlStyles
-                , tabControlListStyles = tabControlListStyles
+                , tabStyles = tabStyles
+                , tabListStyles = tabListStyles
                 , role = Carousel.Group
                 , name = "Testimonials"
                 , visibleLabelId = Nothing
@@ -788,8 +788,8 @@ viewPackages selected =
                       , tabControlHtml = span Style.invisible [ text "Enterprise tier" ]
                       }
                     ]
-                , tabControlStyles = tabControlStyles
-                , tabControlListStyles = tabControlListStyles
+                , tabStyles = tabStyles
+                , tabListStyles = tabListStyles
                 , previousButton =
                     { attributes =
                         [ ClickableSvg.withBorder
@@ -846,8 +846,8 @@ type Package
     | EnterpriseTier
 
 
-tabControlStyles : Bool -> List Style
-tabControlStyles isSelected =
+tabStyles : Bool -> List Style
+tabStyles isSelected =
     [ borderRadius (pct 100)
     , overflow hidden
     , padding zero
@@ -870,8 +870,8 @@ tabControlStyles isSelected =
     ]
 
 
-tabControlListStyles : List Style
-tabControlListStyles =
+tabListStyles : List Style
+tabListStyles =
     [ displayFlex
     , property "gap" "15px"
     , justifyContent center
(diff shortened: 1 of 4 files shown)


src/Nri/Ui/Carousel/V2.elm
```diff
@@ -126,7 +126,6 @@ viewWithCombinedControls :
             , idString : String
             }
     , focusAndSelect : { select : id, focus : Maybe String } -> msg
-    , viewTabControl : id -> Html msg
```
ArthurJordao: I am not sure if this could be deleted, but I deleted from the original draft

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,175 @@
...
+        , slides : Html msg
+        }
+viewWithCombinedControls config =
+    let
+        { controls, slides } =
+            viewWithTabControls config
+    in
+    { tabControls = controls
+    , slides = slides
+    , previousAndNextControls = div [] [ config.viewPreviousButton, config.viewNextButton ]
```
ArthurJordao, line 174: Not sure if there was any attribute that we should add to the buttons 🤔

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,175 @@
...
+    }
+    ->
+        { controls : Html msg
+        , slides : Html msg
+        , containerAttributes : List (Attribute msg)
+        }
+viewWithPreviousAndNextControls config =
+    { controls =
+        div [ css config.controlListStyles ]
+            [ config.viewPreviousButton, config.viewNextButton ]
```
ArthurJordao, line 49: Not sure if there was any attribute that we should add to the buttons 🤔

juanedi: I didn't see anything in the W3 guide!

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,175 @@
...
+    , controlListStyles : List Style
+    }
+    ->
+        { controls : Html msg
+        , slides : Html msg
+        , containerAttributes : List (Attribute msg)
+        }
+viewWithPreviousAndNextControls config =
+    { controls =
+        div [ css config.controlListStyles ]
```
juanedi, line 48: Growth's use case requires rendering one button on each side of the panel, and for our case we'll want to add the "__ of __" label somewhere in between them. Should we return both buttons separately rather than in a container rendered by ourselves?

Screenshot 2023-08-25 at 15 17 48

Screenshot 2023-08-25 at 15 15 51

juanedi: Would definitely be weird to receive the buttons' html just to return it untouched, though? 😆 

💭  Kind of cheating, but this made me wonder if we want to turn `viewPreviousButton` and `viewNextButton` into something like:

```elm
{  ...
  , previousButton : { name : String, attributes: List Button.Attribute }
  , nextButton : { name : String, attributes: List Button.Attribute } 
  , ...
}
```

We could then call `Button.button` ourselves, ensuring that actual buttons are being used (which is also suggesded by the spec). We could even add some default-but-overridable "Previous/Next slide" labels too, while we are at it.

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,175 @@
...
+                          else
+                            Css.display Css.none
+                        ]
+                    ]
+                    [ panel.slideHtml ]
+            )
+            config.panels
+            |> Html.div [ Attrs.attribute "atomic" "false", Attrs.attribute "live" "polite" ]
+    , containerAttributes =
+        [ Attrs.attribute "role" "region"
```
juanedi, line 70: 💭 I think we may want to support role=group here, too:

> A carousel container element that encompasses all components of the carousel, including both carousel controls and slides, has either role [region](https://w3c.github.io/aria/#region) or role [group.](https://w3c.github.io/aria/#group) The most appropriate role for the carousel container depends on the information architecture of the page. See the [Landmark Regions Practice](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/) to determine whether the carousel warrants being designated as a landmark region.

Worth checking with the bats, but after reading a bit i suspect region makes sense for our use case but may not be the case for Growth.

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,175 @@
...
+
+        StringLabel l ->
+            Aria.label l
+
+
+{-| Type which represents the type of aria label which will be used
+`IdLabel` will point to an existing element id on the DOM
+`StringLabel` will be a label of the element
+-}
+type AriaLabel
```
juanedi, line 91: 🎨 thoughts on renaming to this...?

```elm
type LabelledBy
    = LabelledByIdOfVisibleLabel String
    | LabelledByAccessibleLabelOnly String
```

No big deal but I don't see us use "aria" anywhere in our APIs, and I think `AriaLabel` gives the impression that whatever we provide will end up in an `aria-label` attribute.

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,175 @@
...
+                , slideHtml : Html msg
+                , tabControlHtml : Html Never
+                , idString : String
+                }
+        , focusAndSelect : { select : id, focus : Maybe String } -> msg
+        , tabControlStyles : Bool -> List Style
+        , tabControlListStyles : List Style
+    }
+    ->
+        { controls : Html msg
```
juanedi, line 116: I think we still need a top level container with appropriate attributes as in the function above!

src/Nri/Ui/Carousel/V2.elm
juanedi, line 174: 🎨 there's a little inconsistency in that `viewWithPreviousAndNextControls`receives a list of styles for the prev/next container, but this one doesn't. But maybe we won't need to add one anyways as per my other comment of returning the buttons separately!

component-catalog/src/Examples/Carousel.elm
```diff
@@ -44,15 +45,32 @@ type alias Settings =
     { items : Int
     , controlListStyles : ( String, List Style )
     , controlStyles : ( String, Bool -> List Style )
+    , carouselType : CarouselType
     }
 
 
+type CarouselType
```
juanedi, line 52: I didn't check in detail so there may be other things missing but just FYI the code snippet for combined controls actually has a call to `viewWithTabControls`:

Screenshot 2023-08-25 at 15 34 32

Totally fine to leave this all for the end of the code review once we have an API we are all happy with :-)

juanedi (commented):
Thanks a lot!! Left a bunch of comments but feeling pretty good about this overall 🤘

src/Nri/Ui/Carousel/V2.elm
ArthurJordao: I don't know if we should use Nri.UI.Button, they have some styles by default, and the whole component is unystlyed by default.

juanedi: Interesting!

Curious to know what the bats think, but I think the value we are shooting for is **not imposing** anything wrt styles. I personally don't see a practical issue with the button having _some_ style by default (as long as it can be customize to accommodate all use cases we care about).

On the other hand, I just looked at how we the Carousel in the landing page is implemented currently, and we are [using clickable divs with svgs inside](https://github.com/NoRedInk/NoRedInk/blob/49b58a5d4534925c60872f666f98cb9c8f259485/contento/ui/src/Nri/LandingPage.elm#L2931-L2948) (when I'm pretty sure it would be better to [use customized buttons as we do in GDs](https://github.com/NoRedInk/NoRedInk/blob/49b58a5d4534925c60872f666f98cb9c8f259485/monolith/ui/src/Page/Learn/GuidedDrafts/ShortResponseWriting/WritingSample.elm#L763-L776)). Makes me think that there actually is value in enforcing buttons in the API.

WDYT? Happy to let a bat weigh in before we make any decisions!

ArthurJordao: I don't have strong opinions on this one, but I think it's worth to have a opinion from someone on the accessibility bats team first. Maybe @tesk9 could help us on this point?

src/Nri/Ui/Carousel/V2.elm
ArthurJordao: I didn't find any references to this container on the [docs](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/) actually, do we really need it here?

juanedi: Soooo the description for Tabbed carousels says "The structure of a tabbed carousel is the same as a basic carousel except that: [...]".

Since the basic one has that container and there is no explicit instruction that we shouldn't have it in the tabbed version, I assumed it was reasonable to assume that we should keep it in the tabbed version.

ArthurJordao: Ohhh, I totally ignored the "The structure of a tabbed carousel is the same as a basic carousel except that: [...]".
thing and my eyes moved directly to the bullet points haha

ArthurJordao:
Hey @NoRedInk/team-accessibilibats-a11ybats! I tagged you as a review, but we already have some points in discussion here, so I would recommend to go through the comments that Juan and I already made, so you could give us some inputs, specially on this comment:
- https://github.com/NoRedInk/noredink-ui/pull/1488#discussion_r1306002259

component-catalog/src/Examples/Carousel.elm
```diff
@@ -192,47 +208,205 @@ example =
...
+
+            else
+                settings.items - 1
+
+        nextId =
+            if model.selected + 1 < settings.items then
+                model.selected + 1
+
+            else
+                0
```
tesk9, line 306: If possible, I think the "next" and "previous" selections should happen _inside_ the carousel component, rather than out here (for `viewWithCombinedControls` and `viewWithPreviousAndNextControls`)

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,206 @@
...
+type Role
+    = Group
+    | Region
+
+
+{-| Builds a carousel with previous and next controls
+Returns:
+`controls`: previous and next buttons element
+`slides` the container with the carousel contents
+`containerAttributes` attributes that should be used on the parent div of both the button and slides elements
```
tesk9, line 42: I don't think this doc comment is quite correct -- the return value is a record with `viewPreviousButton`, `viewNextButton`, `slides`, and `containerAttributes`.

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,206 @@
...
+        , containerAttributes : List (Attribute msg)
+        }
+viewWithPreviousAndNextControls config =
+    { viewPreviousButton = config.viewPreviousButton
+    , viewNextButton = config.viewNextButton
+    , slides =
+        List.map
+            (\panel ->
+                Html.div
+                    [ Attrs.attribute "role" "group"
```
tesk9, line 70: Not a big deal, so feel free to ignore, but you should be able to use `Role.group`

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,206 @@
...
+                            Css.display Css.block
+
+                          else
+                            Css.display Css.none
+                        ]
+                    ]
+                    [ panel.slideHtml ]
+            )
+            config.panels
+            |> Html.div [ Attrs.attribute "atomic" "false", Attrs.attribute "live" "polite" ]
```
tesk9, line 84: I know the authoring practices guide recommends putting the slides in a live region, but I think we want to instead manage announcements ourselves more directly. Using live regions, you can end with up content getting way over-announced, and it's difficult to fix once it starts happening. 

For example, go to a regular Guided Draft, turn on VoiceOver, and click into a component -- because the sidebar content changes, the tips and tricks carousel that currently uses aria-live gets announced in a confusing and unhelpful way.

Maybe we can expose a standard announcement from this component for when the selected slide changes, and document that it should be used when the slide changes? Or maybe adjust the api such that it takes a field like `selectAndAnnounce : { select : id, announce : String } -> msg`?

tesk9 (commented):
This is looking good! I left some comments, but before you address them, heads up that I'm also going to push a commit adjusting how the example code is generated.

component-catalog/src/Examples/Carousel.elm
ArthurJordao: Do you think it's reasonable to accept a new field like:
```
    , onRotateCarousel: Prev/Next -> msg
```

Or do you have something else in mind?

src/Nri/Ui/Carousel/V2.elm
ArthurJordao: I don't know if I quite get what this selectAndAnnouce would do, could you give me an example for it?

ArthurJordao:
Hey @tesk9 I made some changes to the PR, would you mind taking a look again later?

src/Nri/Ui/Carousel/V2.elm
tesk9: GitHub isn't letting me upload a video right now, annoying!

There are basically 2 ways of communicating to SR users that something has changed elsewhere on the page:
- move focus to the changed item (for example, when navigating between SPA pages, we move focus to the new page's heading)
- use a live region

For our purposes, we _only_ want our AT announcements center to include live regions. Everywhere else, we don't want live regions. So anything that would _otherwise_ be something we'd use a live region to communicate, we instead want to send a (potentially shortened) message to the ATAC instead (with Nri.Effect.announcePolitely and/or Nri.Effect.announceAssertively).

I do see that you're currently trying to move focus to the newly-selected slide. This works in theory (although I'm not sure if it's ergonomic for keyboard users?). However, for the slide to actually _accept_ focus, you'll need to add `Accessibility.Styled.Key.tabbable False` (which adds tabindex=-1).

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,307 @@
...
+                    (\index ->
+                        List.Extra.getAt
+                            (if index + 1 < List.length config.panels then
+                                index + 1
+
+                             else
+                                0
+                            )
+                            config.panels
+                    )
```
tesk9, line 102: Could you add tests, please? I'm thinking that this logic in particular would be nice to have covered, by way of some program-tests with varying numbers of slides.

tesk9:
Took a look! I didn't give this a very thorough review yet, though. Let me know if you'd like me to review thoroughly before you start making changes and I'll do so. Otherwise, I'll await your next ping! 🙏

src/Nri/Ui/Carousel/V2.elm
juanedi: Ended up enforcing ClickableSvg! Marking this thread as resolved.

tests/Spec/Nri/Ui/Carousel.elm
```diff
@@ -146,6 +147,15 @@ viewWithPreviousAndNextControlsSpec =
                     |> clickButton "Previous"
                     |> ensureSlideIsVisible "slide-0"
                     |> done
+        , test "does not crash with 0 slides" <|
```
tesk9, line 150: When 0 slides are passed, `findPreviousAndNextSlides` breaks with a divide by zero error. Maybe an approach like is used in FocusLoop might work here?

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+import Css exposing (..)
+import Html.Styled as Html exposing (..)
+import Html.Styled.Attributes as Attrs exposing (id)
+import List.Extra
+import Nri.Ui.ClickableSvg.V2 as ClickableSvg
+import Nri.Ui.Svg.V1 exposing (Svg)
+import TabsInternal.V2 as TabsInternal
+
+
+{-| `Role`, which can be either [Group](https://w3c.github.io/aria/#group) or [Region](https://w3c.github.io/aria/#region)
```
tesk9, line 31: I haven't looked back at the authoring practices guide or experimented at all, and I'm noticing that looking at the docs/examples, I don't know which Role to use.

If we want devs/designers to make a choice here, can we provide some more guidance and add the options to the configurable settings on the example page? Or alternatively, can we pick an approach?

(LMK if you want me to do more research and provide a suggestion here! Out of time for today, but perfectly willing to look into this more if desired.)

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
+module Nri.Ui.Carousel.V2 exposing
+    ( viewWithPreviousAndNextControls
+    , viewWithTabControls
+    , viewWithCombinedControls
+    , Role(..)
+    )
+
+{-| Patch changes:
```
tesk9, line 8:
```suggestion
{-| Changes from V1:
```

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+
+                          else
+                            Attrs.style "display" "none"
+                        ]
+                        [ slide.slideHtml ]
+                )
+                config.slides
+            )
+    , containerAttributes =
+        [ Attrs.attribute "role" (roleToString config.role)
```
tesk9, line 125: 🎨 picky, so feel free to ignore, but maybe `roleToString` could change type and name to be `role -> Attribute msg` and use `Role.group` and `Role.region`

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+
+-}
+viewWithTabControls :
+    { selected : id
+    , slides :
+        List
+            { id : id
+            , idString : String
+            , name : String
+            , visibleLabelId : Maybe String
```
tesk9, line 146: These properties don't seem used. 💀 ?

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+                , tabListStyles = config.tabControlListStyles
+                }
+    in
+    { controls = tabList
+    , slides = tabPanels
+    , containerAttributes =
+        [ Attrs.attribute "role" (roleToString config.role)
+        , Aria.roleDescription "carousel"
+        , labelAttribute config
+        ]
```
tesk9, line 185: So the previous version of Carousel didn't return `containerAttributes` from `Carousel.view` -- it just returned `controls` and `slides`.

Do we need this return value? If so, could you add it to the documentation please?

If not, then I think there are some values in the API for this function that aren't necessary anymore:
- role
- name
- visibleLabelId

tesk9: Oh, I see -- this is the same topic as https://github.com/NoRedInk/noredink-ui/pull/1488/files#r1306015638.

Let's add the return value to the documentation then, if you would! 🙏

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+    , previousButton : { name : String, icon : Svg, attributes : List (ClickableSvg.Attribute msg) }
+    , nextButton : { name : String, icon : Svg, attributes : List (ClickableSvg.Attribute msg) }
+    , role : Role
+    , name : String
+    , visibleLabelId : Maybe String
+    , focusAndSelect : { select : id, focus : Maybe String } -> msg
+    , announceAndSelect : { select : id, announce : String } -> msg
+    }
+    ->
+        { tabControls : Html msg
```
tesk9, line 220: 🎨 also super picky, but `viewWithTabs` returns a record with `controls` and this funciton returns a record with `tabControls`. Could we standardize? 🙏

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+
+{-| Builds a carousel with tab buttons
+Returns:
+
+  - `tabControls`: tabs control buttons
+  - `slides` container with the carousel contents
+  - `viewPreviousButton` previous button
+  - `viewNextButton` next button
```
tesk9, line 195: ✏️ Could you please add `containerAttributes` to these docs? 🙏

src/Nri/Ui/Carousel/V2.elm
```diff
@@ -0,0 +1,344 @@
...
+            }
+    , tabControlStyles : Bool -> List Style
+    , tabControlListStyles : List Style
+    , previousButton : { name : String, icon : Svg, attributes : List (ClickableSvg.Attribute msg) }
+    , nextButton : { name : String, icon : Svg, attributes : List (ClickableSvg.Attribute msg) }
+    , role : Role
+    , name : String
+    , visibleLabelId : Maybe String
+    , focusAndSelect : { select : id, focus : Maybe String } -> msg
+    , announceAndSelect : { select : id, announce : String } -> msg
```
tesk9, line 217: What do you think about combining the msg type used here:

`select : { select : id, announce : String, focus : Maybe String} -> msg`

?

It struck me as a little but inconvenient to have 2 msg constructors to manage the state when I was adding an example.

Although I suppose if folks want to track click events for the tab controls vs the <> buttons separately, having 2 constructors might be more convenient 🤔

tesk9 (commented):
This is getting so close! 😍 

Thank you for all your awesome work on this! 

I pushed a couple of commits expanding the examples (partly to ensure that I fully understood the API for each view function, and what the different pieces were used for). While working on those additions, I noticed that empty slides cause a JS error, so I also added some failing tests for 0 slide cases. Finally, I implemented a live region fake "announcement center" similar to the one in Nri.Spa/Nri.Program in order to make the example more testable/authentic with an actual screenreader.

One other small thing -- could you finish out the PR description? Thanks! LMK if you need explanation or support on any of the checklist items or on how to use the template

tesk9:
Oh right! This is small, and frankly I don't care at all if you fix it or ignore it.

But thought I'd mention it!

So I grabbed the "dot" styles from contento's use of Carousel.V1. What they're _trying_ to do is to set the focus styles to be explicitly `FocusRing.boxShadows []`. This is the version of the focus ring that _doesn't_ change the border radius when an element is focused. [See the example page for FocusRing](https://noredink-ui.netlify.app/#/doodad/FocusRing).

However, unless the `FocusRing.customClass` is on the element that receives focus (e.g., the button), **we will use `FocusRing.styles`**, which _does_ change the border radius.

Which is to say, keyboard users see a very square dot:
image

I don't have a strong opinion on what should be done here. If you add `FocusRing.customClass`, please also either add some default focus ring behavior or make it clear to devs using the package that they'll need to specify the focus ring behavior.

If you ignore it, and design doesn't mind, sounds fabulous to me -- less complexity to worry about!

src/Nri/Ui/Carousel/V2.elm
ArthurJordao: I added more context about both of them in the docs!

src/Nri/Ui/Carousel/V2.elm
ArthurJordao: Sure, just did it, the unique thing is that `Role.region` doesn't exist in this case

juanedi:
@tesk9 thank you so much for the review!

This should be ready for another look. We took the liberty to do some extra renames for consistency, and ended up addressing the focus ring issue by letting users of the API attach attributes to the tabs (so anyone in need of customizing the focus ring via `FocusRing.customClass` can chose to do so). We actually did it in the example page:

Screenshot 2023-09-06 at 16 02 59

tesk9 (approved):
Thanks!



I pushed a commit adding the Role configuration to the examples page. If you're fine with it, please feel free to merge! 🎉

PR #1489, opened 2023-08-24
HTML links

bendansby:
Adds a link to the sample HTML content and improves link styles for Message.html

image


charbelrami (approved):
Looks good!

PR #1497, opened 2023-08-31
🔧 Support ATAC in new Modal.V12

tesk9:
towards A11-2116

Adds Modal.V12 which:
- removes use of FocusTrap type alias
- returns ids to focus on instead of cmds (in order to make modals more testable via the effect pattern with program-test)
- uses tesk9/accessible-html-with-css for SR-only content
- simplifies internal attributes pattern
- adds Modal.atac to make a hole for the ATAC to render inside the modal

Screen Shot 2023-08-31 at 11 57 12 AM
(Not pinging designers, as the visual changes are constrained to this example and are not particularly interesting.)

        - an accurate preview
        - valid sample code
        - correct keyboard behavior
  - [A11-3504](https://linear.app/noredink/issue/A11-3504/update-modalv11-to-modalv12-everywhere)

### DIFF
--- a/component-catalog/src/Examples/Modal.elm
+++ b/component-catalog/src/Examples/Modal.elm
@@ -17,7 +17,7 @@ import Debug.Control.View as ControlView
 import EventExtras exposing (onClickStopPropagation)
 import Example exposing (Example)
 import Guidance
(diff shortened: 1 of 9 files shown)


charbelrami (approved):
Looks great!

PR #1513, opened 2023-09-11
[PHX-1127] Fix vertical spacing above labels

micahhahn:

## Context

- https://github.com/NoRedInk/NoRedInk/pull/45661

I've noticed for a while (and we've had mild complaints from curriculum) that labels in scaffolding appear to consume more vertical space than they need to.  I finally tracked down the root cause... see the image below (where I've tweaked the child of the `balloon-spacer` element to have `width: 10px`.

The spacers line height is greater than than of the text it is beside!  This is due to us setting the `line-height` to `unset` which opens the child up to inherit an ancestors line height.  If we set the `line-height` to `initial` instead (which defaults to `normal`) then all is well in the world. 

This change doesn't really affect the example catalog as much as scaffolding due to the ancestor line height issue. 

| Before | After |
| --- | --- |
| image | image |

(Note: these are what the hack fix looks like on scaffolding)

        - an accurate preview
        - valid sample code
        - correct keyboard behavior

### DIFF
--- a/src/Nri/Ui/Mark/V5.elm
+++ b/src/Nri/Ui/Mark/V5.elm
@@ -7,6 +7,11 @@ module Nri.Ui.Mark.V5 exposing
 {-|
 
 
+### Patch changes:
+
+    - Fix line-height for spacer element reducing the amount of vertical space consumed by the balloons
+
+
 ### Changes from V4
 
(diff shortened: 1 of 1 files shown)


charbelrami (approved):
```
* {
    all: initial;
}
```

fixes all your CSS issues 😆

PR #1520, opened 2023-09-13
A11-3357 Include `Nri.Effect.focus` in the Modal docs

charbelrami:
Adds documentation to inform engineers that they should use `Nri.Effect.focus` instead of `Dom.focus` when working in the monorepo.


BrianHicks (approved):
focusing-season-8-gif-by-paramount

@NoRedInkBot r+

PR #1526, opened 2023-09-15
A11-3361 Add docs and expose keyboard and mouse test helpers

charbelrami:
Adds docs and exposes keyboard and mouse test helpers for use within the monolith.


bcardiff (approved):

elm.json
```diff
@@ -6,7 +6,8 @@
     "version": "24.8.2",
     "exposed-modules": [
         "Browser.Events.Extra",
-        "Nri.Ui",
```
caseyWebb, line 9: @charbelrami was this intentional?

PR #1529, opened 2023-09-18
Added html function to QuestionBox

nunocf:

## Context

[PHX-1151](https://linear.app/noredink/issue/PHX-1151/support-passing-html-to-noredink-ui-questionbox) requires us to add a html function to the QuestionBox to allow us to render DisplayElements within the question box.

Here's a screenshot of the component catalogue using the new `html` function
image

    - an accurate preview
    - valid sample code
    - correct keyboard behavior
  - [x]  If there are user-facing changes, a designer:

### DIFF
--- a/component-catalog/src/Examples/QuestionBox.elm
+++ b/component-catalog/src/Examples/QuestionBox.elm
@@ -349,10 +349,10 @@ initAttributes =
                                 [ p []
                                     [ text "Let's remember the rules "
                                     , ul []
-                                        [ li [] [text "Always capitalize the first word of a title"]
-                                        , li [] [text "Capitalize all words except small words like “and,” “of” and “an.”"]
+                                        [ li [] [ text "Always capitalize the first word of a title" ]
+                                        , li [] [ text "Capitalize all words except small words like “and,” “of” and “an.”" ]
                                         ]
-                                    , strong [] [text "How should this be capitalized?"]
+                                    , strong [] [ text "How should this be capitalized?" ]
                                     ]
                                 ]
                             )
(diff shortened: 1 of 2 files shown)


micahhahn (changes requested):
The code generation for the markdown is not right... (it should be `QuestionBox.markdown` instead of `content`).

image

ravi-morbia:
@nunocf the spacing for the text elements inside seem to be different between markdown and html.
Can we match the html spacing to be the same as markdown?
Also are display elements part of this PR, or is this a separate, pre-work ticket for adding display elements to the QuestionBox?

micahhahn:
> Also are display elements part of this PR, or is this a separate, pre-work ticket for adding display elements to the QuestionBox? 

This is pre-work! We will be rendering the display elements in a different place, and then handing off the HTML directly to the QuestionBox component

nunocf:
@ravi-morbia 

> Can we match the html spacing to be the same as markdown?
Do you mean for the example or in general?
The container accepts any kind of html, with its own CSS. I can change the CSS to match the markdown example but it would only be useful for this example page so I feel it is unnecessary, but I can change it if you think it will help!

ravi-morbia:
Hey @nunocf I meant for all instances when html is being used. So there is no way for us to control the spacing between paragraphs with html, is that right?

charbelrami (approved):
Looks good from a11y perspective!

nunocf:
@ravi-morbia We can force some styles, but whoever is calling this new function will most likely be adding styles to it, so I think us allowing the default CSS styles defined globally make the behavior more predictable.

My opinion is that we shouldn't override styling because of the above.

ravi-morbia:
Gotcha, so to confirm, when we use html from the other side in QuestionBox, we will be able to define spacing, correct? I think that's fine then.

The only reason I would say it might be better to override here is because we have defined spacing for paragraph text and bullet points. So it seems like it would keep consistency better to define it here. But maybe I am missing something.

In any case, thanks for discussing. I'll go ahead and approve.

ravi-morbia (approved):

PR #1532, opened 2023-09-18
Lazy focus loop

caseyWebb:

## Context

This PR adds `Nri.Ui.FocusLoop.view` and `Nri.Ui.FocusLoop.Lazy.lazy..lazy5`

The rationale behind this stems from work in the student interests page, but the tl;dr:

We **must** use `Html.Lazy` in student interests, else the page becomes wholly un-usable. As it stands, the only way to add `FocusLoop` to this and maintain the laziness is by calling `addEvents` in init/update and keeping a reference to the events in the model so that they can be accurately checked by reference. This is cumbersome, and we'd prefer not to keep event handlers on the model.

With these new helpers, `FocusLoop` becomes responsible for the `view`, allowing it to tap into the first 2 args of `Html.Lazy.lazyN` itself, negating the need for callers to care about the specifics.

  - [ ] Component docs include a changelog
    - an accurate preview
    - valid sample code
    - correct keyboard behavior
- [x] ~If this is a new major version of the component, our team has stories created to upgrade all instances of the old component. Here are links to the stories:~
  - Not a major version
  - [ ]  ~If there are user-facing changes, a designer:~


src/Nri/Ui/FocusLoop/V1.elm
```diff
@@ -1,80 +1,76 @@
...
+        div
+            [ Key.onKeyDownPreventDefault arrowKeyHandlers ]
+            [ text item.name ]
 
+Does your list support adding and removing items? If so, check out `FocusLoop.Lazy` which
+will prevent recalculation of event handlers for every item when the list changes.
 
-addEvents_ :
+-}
+view : Config id msg item -> List item -> List (Html msg)
```
bcardiff, line 52: We should avoid type alias of records in exported functions to have better error reporting https://github.com/NoRedInk/NoRedInk/blob/master/docs/sdgg/elm-style-guide.md#prefer-explicit-types-in-type-signatures-to-type-aliases

src/Nri/Ui/FocusLoop/Lazy/V1.elm
```diff
@@ -0,0 +1,158 @@
...
+        }
+        [ FocusLoop.Args2 a1 a2
+        , FocusLoop.Args2 b1 a2
+        ]
+
+    viewFocusableItem (FocusLoop.Args2 arg1 arg2) arrowKeyHandlers =
+        ...
+
+-}
+lazy2 : LazyFocusLoop msg (Args2 a1 a2)
```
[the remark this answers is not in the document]
caseyWebb: It was originally tuples but elm has an arbitrary restriction on the number of items so it only works up to 3, then you have to use a record and doing `{ a1 : a1, a2 : a2, ... }` felt bad

src/Nri/Ui/FocusLoop/V1.elm
caseyWebb: hmm, if we're going to do that we can drop the uncurry from the view and id too and pass those args in separately. the only reason I didn't do that was trying to avoid having a bunch of different config types

src/Nri/Ui/FocusLoop/V1.elm
```diff
@@ -1,80 +1,76 @@
...
-                        else
-                            []
-                in
-                ( Just item
-                , ( item, List.filterMap identity (leftRightEvents ++ upDownEvents) ) :: acc
-                )
-            )
-            ( firstId, [] )
-        |> Tuple.second
+    -> List ( a, List (Key.Event msg) )
```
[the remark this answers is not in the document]
bcardiff: Ok, I missed that.

src/Nri/Ui/FocusLoop/V1.elm
bcardiff: Sounds good to drop the uncurry and have specific record types for view/view2/viewN. If we need to keep the Args type maybe that can be an internal not exported type.

I would not implement all N, just up to the one we need to student interest.

caseyWebb: Playing with this now, I'm not sure if we can drop the uncurry _and_ keep everything contained to a helper fn without higher-kinded types. I'd prefer to avoid having the business logic in the fn for each arity.

> I would not implement all N, just up to the one we need to student interest.

Any particular reason? I've found this can change/fluctuate pretty rapidly as functionality is added, and since we can only have up to 5 anyway idk why we wouldn't have all of them from the get-go.

charbelrami (commented):
Looks great! In the usage example, could you add a submit button for the Complexity input? It gets slow when changing the value dynamically.

src/Nri/Ui/FocusLoop/Lazy/V1.elm
```diff
@@ -0,0 +1,224 @@
+module Nri.Ui.FocusLoop.Lazy.V1 exposing
+    ( lazy, lazy2, lazy3, lazy4, lazy5
+    , Args2, Args3(..), Args4(..), Args5(..)
```
bcardiff, line 3: It seems we can drop ArgsN from the export 🎉

caseyWebb:
> Looks great! In the usage example, could you add a submit button for the Complexity input? It gets slow when changing the value dynamically.

@charbelrami I added onBlur/onMouseout handlers, but can add a button if you'd prefer that specifically.

charbelrami (approved):
that's great! Thank you!

bcardiff (approved):

PR #1533, opened 2023-09-19
A11-3346 Add tests and docs for Helpfully Disabled Switch

charbelrami:
## Context

The Switch already had the correct behavior and attributes and was functioning with a helpfully disabled tooltip, so there was no need to change its implementation. I have added documentation and tests to ensure that it works correctly.

[A11-3346 : Add helpfully disabled version of Switch](https://linear.app/noredink/issue/A11-3346/add-helpfully-disabled-version-of-switch)

helpfully disabled switch with tooltip


omnibs (approved):
Looks good!

PR #1536, opened 2023-09-23
Add elm-review rule for FocusLoop.Lazy

caseyWebb:
This PR introduces an elm-review rule that can be used by consumers of nri-ui to ensure proper usage of `FocusLoop.Lazy` per the comment in `Nri.Ui.FocusLoop.Lazy.V1`.

I'm not sure this is exactly the best place for this to live, but I'm also not sure where else it would go.

### DIFF
--- a/component-catalog/review/elm.json
+++ b/component-catalog/review/elm.json
@@ -1,7 +1,8 @@
 {
     "type": "application",
     "source-directories": [
-        "src"
+        "src",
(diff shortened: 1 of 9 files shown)


charbelrami (approved):
Looks good!

bcardiff:
I think this rule should be moved to monolith to avoid elm-review runtime dependency

PR #1543, opened 2023-09-28
A11-3348 Helpfully disabled select

charbelrami:

## Context

[A11-3348 : Add helpfully disabled version of Select](https://linear.app/noredink/issue/A11-3348/add-helpfully-disabled-version-of-select)

The purpose of this PR is to replace `disabled` with `aria-disabled="true"` to keep the element in the tab order. Unlike `disabled`, `aria-disabled` does not remove event listeners from the element, so we need to remove these interactions programmatically.

The initial attempt was to prevent default on `mousedown` and some `keydown` events when the select element is disabled. However, this would require passing a `NoOp` msg to `Select.disabled` since `Events.preventDefaultOn` needs a msg. This would also require an `Attribute msg` type, which would conflict with the existing `Attribute a` type.

To avoid changes in the API, the current solution is to replace the `select` element with a `div` element with `aria-disabled="true"` and `tabindex="0"`. This change prevents user interactions while keeping the element in the tab order.

    - an accurate preview
    - valid sample code
    - correct keyboard behavior
  - [x]  If there are user-facing changes, a designer:

### DIFF
--- a/src/Nri/Ui/Select/V9.elm
+++ b/src/Nri/Ui/Select/V9.elm
@@ -50,6 +50,8 @@ module Nri.Ui.Select.V9 exposing
 
 import Accessibility.Styled as Html exposing (Html)
 import Accessibility.Styled.Aria as Aria
+import Accessibility.Styled.Key as Key
+import Accessibility.Styled.Role as Role
 import Css
 import Dict
 import Html.Styled.Attributes as Attributes exposing (css)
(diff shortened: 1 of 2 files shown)


lindsaykwardell:
> To avoid changes in the API, the current solution is to replace the `select` element with a `div` element with `aria-disabled="true"` and `tabindex="0"`. This change prevents user interactions while keeping the element in the tab order.

Just making sure I'm following (in case there's been prior discussion I missed), does replacing the `select` with a tabbable `div` still make it clear to screen reader users that this is still usable as a dropdown? I know the goal is to make the content visible to screen readers, so just wanted to confirm this doesn't introduce another problem.

charbelrami:
> Just making sure I'm following (in case there's been prior discussion I missed), does replacing the `select` with a tabbable `div` still make it clear to screen reader users that this is still usable as a dropdown but that it's not enabled right now? I know the goal is to make the content visible to screen readers, so just wanted to confirm this doesn't introduce another problem.

I have added `Role.listBox` to the `div`. Although it is not perfect, since screen readers announce it as "dimmed, list box" instead of "dimmed, select", it is sufficient for users to recognize it as a disabled list of options.

lindsaykwardell (approved):
Thank you for the note! This change looks good to me!

PR #1545, opened 2023-09-28
Revert elm dependencies change due to #1526

bcardiff:
Workaround the strict dependency introduced in #1526 .
Otherwise dependent projects are forced to match `avh4/elm-program-test` and `elm-explorations/test`.

Merging this will require a major release yet only Nri.Test.* modules are changed

It also reverts an unintended change in https://github.com/NoRedInk/noredink-ui/pull/1526/files#diff-b2798733fa48dc75a65aa2d774f9f5e1a294e46f04c1161290ae62755755f20dL5


charbelrami (approved):
Looks great! Liked the config approach!

caseyWebb (approved):

PR #1561, opened 2023-10-26
A11-2660 refactor and add ClickableText tests

charbelrami:
Restructures the ClickableText spec, improves the descriptions for better understanding, adds tests.

### DIFF
--- a/tests/Spec/Nri/Ui/ClickableText.elm
+++ b/tests/Spec/Nri/Ui/ClickableText.elm
@@ -1,4 +1,4 @@
-module Spec.Nri.Ui.ClickableText exposing (..)
+module Spec.Nri.Ui.ClickableText exposing (spec)
 
 import Accessibility.Aria as Aria
 import Accessibility.Role as Role
--- a/tests/Spec/Nri/Ui/ClickableText.elm
+++ b/tests/Spec/Nri/Ui/ClickableText.elm
@@ -36,6 +36,8 @@ elementTests =
     [ testAsButton
     , testAsAnchor
     , testIconAsSvg
+    , testRightIconAsSvg
+    , testLinkExternalIconAsSvg
     ]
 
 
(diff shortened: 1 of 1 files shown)


tests/Spec/Nri/Ui/ClickableText.elm
```diff
@@ -1,72 +1,411 @@
-module Spec.Nri.Ui.ClickableText exposing (spec)
+module Spec.Nri.Ui.ClickableText exposing (..)
```
bcardiff, line 1: Is there a reason to expose everything instead of the top level `spec`?

tests/Spec/Nri/Ui/ClickableText.elm
```diff
@@ -1,72 +1,411 @@
...
+        , describe "attributes" attributeTests
+        , describe "icon accessibility" iconAccessibilityTests
+        , describe "disabled behavior and attributes" disabledStateTests
         ]
 
 
-helpfullyDisabledClickableText : List Test
-helpfullyDisabledClickableText =
-    [ test "does not have `aria-disabled=\"true\" when not disabled" <|
+type Type_
```
bcardiff, line 29: Nothing against this `Type_`, but from I can see having `programLink` and `programButton` functions would be enough today. Are there any plans in the future?

tests/Spec/Nri/Ui/ClickableText.elm
```diff
@@ -1,72 +1,411 @@
...
+    , testAsAnchor
+    , testIconAsSvg
+    , testRightIconAsSvg
+    , testLinkExternalIconAsSvg
+    ]
+
+
+testAsButton : Test
+testAsButton =
+    test "the `button` type renders as a button element" <|
```
bcardiff, line 46: I think we usually nest describes and tests directly. Is there a reason to flatten the specs and organizing them as proposed?

I checked on our elm style guide and there is nothing against but it seems like a deviation from what we usually do, so I wonder the motivations in case I am missing something.

bcardiff (approved):
Feel free to merge directly. My comments are optional/discussions.

tests/Spec/Nri/Ui/ClickableText.elm
charbelrami: that was unintended

tests/Spec/Nri/Ui/ClickableText.elm
charbelrami: no specific reason, I just thought it would look cleaner, but don't feel strongly. Updated!

charbelrami:
thank you @bcardiff !

tests/Spec/Nri/Ui/ClickableText.elm
charbelrami: no plans! Updated!

PR #1594, opened 2024-01-12
Tessa/show underline example

tesk9:
I noticed that https://github.com/NoRedInk/noredink-ui/pull/1582 didn't update the static examples for Block use, so I thought I'd add expand the examples (good to have Percy looking out, after all!)

I think adding these additional cases also makes it apparent that the API now does some presumably undesirable things:

### 1. Double underlines for blank within emphasis

image

Note that the "emphasis" is no longer dashed. The emphasis uses an underline style _and_ the blank uses an underline style.

### 2. Double underline for blanks with emphasis

image

There's a similar problem here -- the blank has 2 underlines underneath it.

### 3. Should non-blanks ever be underlined?

image

Is it intended that the API can support underlining non-blanks?

| | Before | After |
| --- | --- | --- |
| Main complex static example | Screenshot 2024-01-12 at 3 34 32 PM |  Screenshot 2024-01-12 at 3 34 16 PM |
| Simple blank example | image |  image |
| Within emphasis blank example | image | image |
| Blank width guidance | image | image |

### DIFF
--- a/component-catalog/src/Examples/Block.elm
+++ b/component-catalog/src/Examples/Block.elm
@@ -370,7 +370,7 @@ example =
                                 ]
                                 1
                   , description =
-                        """**Labeled blank block**
+                        """**Labeled dashed blank block**
 
 A labelled blank in the sentence.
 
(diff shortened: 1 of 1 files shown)


lindsaykwardell:
At the moment, this API is only being used for blank, non-emphasized blocks, but I agree that it should work for all cases. I've got another PR going to do some cleanup on this that Stacey requested, I can make those adjustments there.

lindsaykwardell (approved):

PR #1595, opened 2024-01-12
Clean up blank underline block

lindsaykwardell:
Change some minor styles based on feedback from Stacey.

## Context

https://github.com/NoRedInk/NoRedInk/pull/46846#issuecomment-1883598283

image

    - an accurate preview
    - valid sample code
    - correct keyboard behavior
  - [x]  If there are user-facing changes, a designer. (You may want to direct your designer to the [deploy preview](https://github.com/NoRedInk/noredink-ui#reviews--preview-environments) for easy review):

### DIFF
--- a/src/Nri/Ui/Block/V6.elm
+++ b/src/Nri/Ui/Block/V6.elm
@@ -704,7 +704,16 @@ render config =
             , backgroundColor = palette.backgroundColor
             , maybeMarker = maybeMark
             , labelPosition = config.labelPosition
-            , labelCss = config.labelCss
+            , labelCss =
+                config.labelCss
+                    ++ (case ( config.borderStyle, config.content ) of
+                            ( Underline, [ Blank _ ] ) ->
+                                [ Css.top (Css.px 6)
+                                ]
+
+                            _ ->
+                                []
+                       )
             , labelId = config.labelId
             , labelContentId = Maybe.map labelContentId config.labelId
             }
@@ -721,14 +730,19 @@ viewBlank : BorderStyle -> BlankHeight -> CharacterWidth -> Html msg
 viewBlank borderStyle blankHeight (CharacterWidth width) =
     let
         heightStyles =
-            case blankHeight of
-                BlankHeightFull ->
-                    [ Css.paddingTop topBottomSpace
-                    , Css.paddingBottom topBottomSpace
-                    , Css.lineHeight Css.initial
-                    ]
+            case borderStyle of
+                Dashed ->
+                    case blankHeight of
+                        BlankHeightFull ->
+                            [ Css.paddingTop topBottomSpace
+                            , Css.paddingBottom topBottomSpace
+                            , Css.lineHeight Css.initial
+                            ]
 
-                BlankHeightInline ->
+                        BlankHeightInline ->
+                            [ Css.lineHeight (Css.num 1) ]
+
+                Underline ->
                     [ Css.lineHeight (Css.num 1) ]
     in
     span
@@ -738,7 +752,7 @@ viewBlank borderStyle blankHeight (CharacterWidth width) =
                     Css.border3 (Css.px 2) Css.dashed Colors.navy
 
                 Underline ->
-                    Css.borderBottom2 (Css.px 2) Css.solid
+                    Css.borderBottom2 (Css.px 1) Css.solid
             , MediaQuery.highContrastMode
                 [ Css.property "border-color" "CanvasText"
                 , Css.property "background-color" "Canvas"
--- a/src/Nri/Ui/Block/V6.elm
+++ b/src/Nri/Ui/Block/V6.elm
@@ -340,7 +340,7 @@ renderContent config content_ styles =
                         BlankHeightFull
             in
             blockSegmentContainer
-                [ viewBlank config.borderStyle blankHeight length ]
+                [ viewBlank config.blankStyle blankHeight length ]
                 styles
 
         Markdown markdown contents ->
@@ -425,14 +425,14 @@ italic =
 -}
 dashed : Attribute msg
 dashed =
-    Attribute (\config -> { config | borderStyle = Dashed })
+    Attribute (\config -> { config | blankStyle = Dashed })
 
 
 {-| Sets the border style to be underlined
 -}
 underline : Attribute msg
 underline =
-    Attribute (\config -> { config | borderStyle = Underline })
+    Attribute (\config -> { config | blankStyle = Underline })
 
 
 
@@ -524,7 +524,7 @@ toMark :
         | emphasize : Bool
         , label : Maybe String
         , content : List (Content msg)
-        , borderStyle : BorderStyle
+        , blankStyle : BlankStyle
     }
     -> Palette
     -> Maybe Mark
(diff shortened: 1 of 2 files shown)


staceyadams (approved):

src/Nri/Ui/Block/V6.elm
```diff
@@ -553,23 +553,34 @@ toMark config { backgroundColor, borderColor } =
...
                     ]
+                , Css.batch
+                    (case config.borderStyle of
+                        Dashed ->
+                            []
+
+                        Underline ->
+                            [ Css.textDecoration Css.underline
+                            , Css.property "text-decoration-thickness" "1px"
+                            , Css.property "text-underline-offset" "4.5px"
```
tesk9, line 578: Perhaps 0.2rem or something like that instead? so that this style works well for 30px fonts and 18px fonts.

4.5px looks a bit far to me for an 18px font:

Screenshot 2024-01-12 at 6 39 04 PM

src/Nri/Ui/Block/V6.elm
```diff
@@ -542,7 +542,7 @@ toMark config { backgroundColor, borderColor } =
                     Underline ->
                         Css.solid
 
-            notBottomBorderColor =
+            borderColor_ =
```
tesk9, line 545: Do we definitely want to use a different color for the underline than the text?

image

src/Nri/Ui/Block/V6.elm
```diff
@@ -704,7 +715,16 @@ render config =
             , backgroundColor = palette.backgroundColor
             , maybeMarker = maybeMark
             , labelPosition = config.labelPosition
-            , labelCss = config.labelCss
+            , labelCss =
+                config.labelCss
+                    ++ (case ( config.borderStyle, config.content ) of
+                            ( Underline, [ Blank _ ] ) ->
+                                [ Css.top (Css.px 6)
```
tesk9, line 722: I don't think adjusting the top px will work consistently for a couple of reasons.

The first reason is variable font size. As far as I am aware, the Block component is used in containers that set the font to 30px _and_ in containers that set the font to 18px. I believe the place the underline is going to be used has 18px font. 

| | 30px | 18px | 12px |
| --- | --- | --- |  --- |
| Original Spacing | image |  image |  image |
| New Spacing | image  |  image |  image |

Users _should_ be able to customize their font size without all our styles breaking. But even if we don't set that as a goal, I'm not sure that the 6px from the top will land in the right spot, you know?

The second reason I think we should change approaches here is that setting the top to 6px will sometimes play badly with repositioning:

| previous behavior | top: 6px |
| --- | --- |
| image | image |

<details>
<summary> Code for the above screenshots
</summary>

On https://github.com/NoRedInk/noredink-ui/pull/1594 or with it merged in, adjust the first repositioning example to use:

```
, Block.view
                    [ Block.plaintext "gifts"
                    , Block.label "direct object collision"
                    , Block.yellow
                    , Block.labelId directObjectId
                    , Block.labelPosition (Dict.get directObjectId offsets)
                    ]
                , Block.view [ Block.plaintext " " ]
                , Block.view
                    [ Block.label "long adjectives shouldn't hit a lil"
                    , Block.underline
                    , Block.purple
                    , Block.labelId adjectiveId
                    , Block.labelPosition (Dict.get adjectiveId offsets)
                    ]
```

</details>

A third reason is that when labels _aren't_ repositioned, the 6px change will make them look a little funky when they're rendered next to normal labels:

image

src/Nri/Ui/Block/V6.elm
lindsaykwardell: The main issue I was running into was getting it aligned with a blank underline, which is why I was moving it at all. Happy to go toward a different solution, I wasn't attached or happy with how this was turning out.

src/Nri/Ui/Block/V6.elm
lindsaykwardell: CC @staceyadams

src/Nri/Ui/Block/V6.elm
tesk9: Ahh I see! This is happening because of the user agent stylesheet.

In the screenshot above, there's a label, so the blank is treated as highlighted by "fruit." Semantically, we're using the `mark` tag for this:
image

If there's no label or emphasis, the styles apply as expected:
image

(I don't have an opinion on whether this is worth addressing as part of this PR, but if you do override the user agent stylesheet, please check that high contrast mode keeps working.)

src/Nri/Ui/Block/V6.elm
staceyadams: @lindsaykwardell let's revert this part.

@tesk9 for context of why we wanted to change this: the labels look more detached in the use in Short Response because we use the blank in a 3 sentences in a column, sometimes with label, sometimes without. In the third sentence frame example, you can see that it appears equidistant between the second and third bullet so not immediately visually clear at a glance that it belongs to the third one.

image

src/Nri/Ui/Block/V6.elm
tesk9: Okay! I think I'm following. So border-bottom is used for adding a "blank" underline and text-decoration is used for adding underlines to actual text?

I think this brings me back to my question from yesterday: "Should non-blanks ever be underlined?" Like, is the intention of the underline just to be a different styling for a blank? If so, then I don't think we'd ever need text-decoration at all? (Although I think an API change could clarify the expected behavior a bit.)

staceyadams: > Like, is the intention of the underline just to be a different styling for a blank?

Yes. In our context, students thought the dashed bordered rectangle versions of the blanks meant that it was interactive

lindsaykwardell: How about this?

| 30px | 18px |
|------|------|
|image|image|

Instead of adjusting the border offset for the text, I adjusted the line height of the blank, bringing it up rather than down. I also set the width to 0.2rem, which at a default zoom is rendering at 1.5px.

tesk9: I guess I'm still not clear on the advantage of allowing the text to be underlined? I don't think there's a time when we would want to underline text in this way. 🤔

staceyadams: Oops, missed the first part of Tessa's question. We don't underline text in Short Response. I'm not 💯 sure but don't think there's a use case for it in scaffolding or Guided Tutorials either? (the other 2 places currently using Block)

lindsaykwardell: I was just aiming for some consistency between the two modes. Taking away the names the component is currently using, we have two display modes that we want to be able to choose between:

- Show blanks with dashes
- Show blanks with underlines

Originally, I felt it made more sense to make that a global thing: show _everything_ with dashes, or show _everything_ as underlined. If we are feeling this is not the case, I can scope that down to just blanks.

tesk9: Personally, I think we should not show everything as underlines. The case that really clinches it for me is a blank within an emphasis, which is the one you've got screenshots of up above in this thread:

image

I would expect the "underline" style to look like this:
image

Not like this:

image

staceyadams: Ah, I get where the other underlines came from. We want to apply underline to just blanks. Or apply to everything, but have the border go all the way around except in blanks, where it should just be on the bottom.

We don't currently have a use case to change the dashes on the other Block elements. The border is there for added contrast in addition to the background color.

lindsaykwardell: Fine with me! Here's the update (which looks just like Tessa's example):

image

tesk9:
Looks good! @lindsaykwardell would you please review/merge https://github.com/NoRedInk/noredink-ui/pull/1594/ so that the changes are a little easier to double check?

lindsaykwardell:
Done!

tesk9 (approved):
Looks great! Thanks! 

Someone may want to update the [corresponding Paper document](https://paper.dropbox.com/doc/Display-Elements-and-Scaffolding-Container-additional-things-to-know--CHvyo5VvA_DrMQ_0QL1Z7mrWAg-6vszpNDLoYIiMyg7Wv66s) still (although I'm not sure who owns that work at this point! Not me, I'm pretty sure 😆)

tesk9:
I thought I'd double-check high-contrast mode, and just FYI the experience is a little confusing when the blank appears within an emphasis. If we're planning to use the new underline blank styles with emphasis, this is something we might want to improve.

| | Standard contrast | High contrast |
| --- | --- | --- |
| Blank | Screenshot 2024-01-16 at 3 40 08 PM |  Screenshot 2024-01-16 at 3 37 55 PM |
| Blank in emphasis | Screenshot 2024-01-16 at 3 40 02 PM |  Screenshot 2024-01-16 at 3 37 36 PM |
| Mixed content | Screenshot 2024-01-16 at 3 40 15 PM | Screenshot 2024-01-16 at 3 38 11 PM |

lindsaykwardell:
I just pushed up a change to support the correct background in high contrast mode, although I'm not 100% certain how to enable it locally to test.

tesk9:
> I just pushed up a change to support the correct background in high contrast mode

Yay, thank you!!

>  I'm not 100% certain how to enable it locally to test.

We have an [instructional guide for testing in high contrast mode](https://paper.dropbox.com/doc/Accessibility-testing-High-contrast-mode--B2KR0U3FuMGr3NHJ5SY~mwa5Ag-UsI7UIOzhdR7RyiOVFIW7)! 🎉

staceyadams:
I can own updating the Paper doc

PR #1607, opened 2024-01-26
Adds example view illustrating Menu regression

tesk9:
Adds an example table to Menu showing all the different base trigger types.

Hopefully this will catch regressions like https://noredink.slack.com/archives/CD7BC4G1J/p1706225659272049 in the Percy stage at noredink-ui.

image

@NoRedInk/design I would love to add guidance to this table in terms of when to use each trigger type. I could also easily add, say, Button.secondary trigger types or anything like that if you think it makes sense or would be helpful. Thanks!


bendansby:
We'd need to think on that and write something up, but thank you for the open offer and for jumping in to help prevent future regressions!

bendansby (approved):

charbelrami (approved):

PR #1609, opened 2024-01-26
Adds About section for icons

tesk9:

For every Icon example page, adds a brief identical About section describing our approach to Icon modeling.

Screenshot 2024-01-26 at 2 50 24 PM

Happy to re-style or re-work copy. LMK!


krinhorn (changes requested):
Suggested changes:
1. Use a semantic list & bullets for these if possible.
2. Alter copy so the "For x images" statements are consistent and up-front.
3. Consistently capitalize SVG.
4. Consistently use "SVG" rather than "image" (don't feel strongly; maybe there is a good reason to not do this!)
5. [Maybe] swap the order of the 2nd and 3rd bullets so the default behavior/decorative case is first. Definitely don't feel strongly about this! There may be a very good reason to keep the existing order.

Example:

- Our icons are Elm SVGs, not separate files or sprites. We use an opaque type to represent them, which enables nice type-safe composability across our components.
- For decorative SVGs (which is the default), we add `aria-hidden=true` to the SVG node.
- For non-decorative SVGs, we use Pattern #5: `<svg> + role='img' + <title>` from [Accessible SVGs: Perfect Patterns For Screen Reader Users](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/).

tesk9:
image

I applied your changes & I added a link to the monolith README on adding new SVGs

krinhorn (approved):
🙌 Thank you!

bendansby (approved):

charbelrami (approved):

PR #1614, opened 2024-01-30
Table header color

bendansby:
Set the table header color to gray20 so it's not browser default black

image


bendansby:
@NoRedInkBot r+

bendansby:
I forgot, do we use the bot to merge in noredink-ui?

bendansby:
Looks like the answer is yes, and even if you delete your r+ message

PR #1618, opened 2024-01-31
Use nri colors for color picker

tesk9:
Component Catalog change only. For all Icon examples, the color picker is now noredink-ui-themed and will default to showing only NRI colors.

Thanks to @ap-nri for the awesome idea! I had no idea this was even possible. HTML is so cool.

| | Before | After |
| --- | --- | --- |
| Chrome |  image | image |
| Safari | image | image |
| Firefox (Mac)\* | image | image  |

\* "list" attribute is not supported on Macs, see [can I use table note](https://caniuse.com/mdn-html_elements_input_type_color_list)


bendansby (approved):
So neat!

charbelrami (approved):
cool!

PR #1621, opened 2024-02-02
Add hot reload via elm-live

bcardiff:
Changes `./script/develop.sh` to use elm-live with hot reload.

You might need to drop the local `.direnv` directory if elm-live is not available, I'm not sure how to force the nix-profile to be rebuilt in other way.

In case you are blocked the old script is in `./script/legacy-develop.sh`, we can clean things later and probably drop python as a dependency.

In order to make hot reload work the Example/UsageExample needs to be tweaked a bit. Before this PR we stored them fully in the App model. This included the view, so the hot reload making its magic restored the old views 🙃 . We now store only the example state in the App model. Additionally examples now define `init` instead of `state` field and the `view` functions receive the state explicitly. Example and UsageExample are now immutable declarations.

🍐 with @tesk9


tesk9 (approved):
AHHHHHHHHHH SO GOOD

The modeling changes make the CC WAY better, and live reload is 💯 !

🏆 Best noredink-ui PR of the year award 🏆

PR #1625, opened 2024-02-05
Add Block resources

tesk9:
Component Catalog changes only -- expands Block resources.

image

@micahhahn I've got a list of a11ybat team demos, but I didn't keep track of demos you've done. Are there other resources you want to add to the Block example?


micahhahn:
@tesk9 Cool thanks for updated this! I never gave a demo on block so nothing to add there

micahhahn (approved):

PR #1622, opened 2024-02-02
Style code output

tesk9:
Component Catalog example code reskinning! 🎉 

| | Before | After |
| --- | --- | --- |
| w/1 example, mobile | image | Screenshot 2024-02-02 at 1 32 43 PM |
| w/1 example, desktop | image | Screenshot 2024-02-02 at 1 32 35 PM |
| w/2 examples, mobile | image | image |
| w/2 examples, desktop | image | image |

Based (roughly) off of https://cloud.antetype.com/exports/link/0499b00d-fe06-465c-a8e7-9e9f8f8e9efe/?starting=0&screen=id519257437l

PR #1626, opened 2024-02-05
Add helpfully disabled links

tesk9:

### Button

image

image
(no changes in the second screenshot. I'm including it because the behavior is a bit different than for Checkbox and Switch).

### Checkbox

image

image

### Select

image

Noticing that this looks bad? Why yes! I agree. I happen to know that we switch from mediumBody -> smallBody in https://github.com/NoRedInk/noredink-ui/pull/1622. I'm adding new text here, so I want it to be smallBody, even though it may look pretty terrible briefly, depending on when things merge.

### Switch

image

image

### RadioButton

image

image

### DIFF
--- a/component-catalog/src/Examples/Checkbox.elm
+++ b/component-catalog/src/Examples/Checkbox.elm
@@ -157,7 +157,7 @@ example =
                   }
                 ]
             , Heading.h2
-                [ Heading.plaintext "Helpfully-Disabled Example"
+                [ Heading.plaintext "Helpfully Disabled Example"
                 , Heading.css [ Css.marginTop (Css.px 30) ]
                 ]
             , Tooltip.view
--- a/component-catalog/src/Examples/Checkbox.elm
+++ b/component-catalog/src/Examples/Checkbox.elm
@@ -15,6 +15,7 @@ import Debug.Control as Control exposing (Control)
 import Debug.Control.Extra as ControlExtra
 import Debug.Control.View as ControlView
 import Example exposing (Example)
+import Guidance
 import Html.Styled exposing (..)
 import Html.Styled.Attributes exposing (css)
 import KeyboardSupport exposing (Key(..))
@@ -52,7 +53,7 @@ example =
     , update = update
     , subscriptions = \_ -> Sub.none
     , preview = preview
-    , about = []
+    , about = [ Guidance.helpfullyDisabled ]
     , view =
         \ellieLinkConfig state ->
             let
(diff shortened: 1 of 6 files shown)


component-catalog/src/Examples/Checkbox.elm
```diff
@@ -156,7 +157,7 @@ example =
                   }
                 ]
             , Heading.h2
-                [ Heading.plaintext "Tooltip Example"
+                [ Heading.plaintext "Helpfully-Disabled Example"
```
krinhorn, line 160:
```suggestion
                [ Heading.plaintext "Helpfully Disabled Example"
```

component-catalog/src/Examples/Switch.elm
```diff
@@ -168,7 +169,7 @@ example =
                   }
                 ]
             , Heading.h2
-                [ Heading.plaintext "Tooltip example"
+                [ Heading.plaintext "Helpfully-Disabled Example"
```
krinhorn, line 172:
```suggestion
                [ Heading.plaintext "Helpfully Disabled Example"
```

component-catalog/src/Guidance.elm
```diff
@@ -12,8 +15,42 @@ useATACGuide moduleName =
...
+            , ClickableText.link "read the docs"
+                [ ClickableText.linkExternal "https://paper.dropbox.com/doc/Helpfully-disabled-components--CI8Ma_KHKL1CcCWpWG~p_RTwAg-2RUPgKnBsBNI7ScGDHS73"
+                , ClickableText.appearsInline
+                ]
+            , text " and "
+            , ClickableText.link "watch Charbel's demo"
+                [ ClickableText.linkExternal "https://noredink.zoom.us/rec/play/fwV3mqsxjvF_95N2au0vAN2PmnH2IHZx2yCoAQ76gvZ0fLlrkNcFIuVL6i7ze7y1ivSxq0f6e2EXE-RJ.kHMKX9CBHI1kFM50?canPlayFromShare=true&from=share_recording_detail&continueMode=true&componentName=rec-play&originRequestUrl=https://noredink.zoom.us/rec/share/YvgK0427ADw42fY2edJ_tmkwwvPxz505Kpfhkz5DqF1_eh8sgj7wVfwBQ5FmieM8.P9YlMkM_XY_Kamm6&autoplay=true&startTime=1696520905000&_x_zm_rtaid=VeLjvOzDToKMf1R0XllC7A.1707171050117.67806369f8182aa5b282c10165d75544&_x_zm_rhtaid=323"
+                , ClickableText.appearsInline
+                ]
+            , text " on the Helpfully Disabled pattern."
```
krinhorn, line 54:
```suggestion
            , text " on the Helpfully Disabled pattern. A sample Helpfully Disabled {component name} can be found below."
```

tesk9: This won't be the case for the `Select`

charbelrami (approved):
Looks great!

PR #1628, opened 2024-02-07
Improve radio button example

tesk9:
Component Catalog improvement only.

Make the RadioButton example more usable by:
- making the customizable example _less_ customizable
- organizing the settings
- adding static examples showcasing states/combinations

### Before

noredink-ui netlify app_ (1)

### After

localhost_8000_ (10)

### DIFF
--- a/component-catalog/src/CommonControls.elm
+++ b/component-catalog/src/CommonControls.elm
@@ -59,10 +59,10 @@ specificColor match =
 premiumDisplay : Control ( String, PremiumDisplay )
 premiumDisplay =
     Control.choice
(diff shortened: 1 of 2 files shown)


bendansby (approved):
Look at all those beautiful states!

A few non-blocking suggestions:

This could just be a bullet point
image

Change this to a vertical layout with 10px gap 
image

Same here (not sure why one is black, feel like that comes up every now and then)
image

tesk9:
Thank you! I won't change the RadioButtonDotless to a bullet in this PR because there's some other work (https://github.com/NoRedInk/noredink-ui/pull/1626) in progress that touches it.

I'll do the other suggestions now, thanks!

charbelrami (approved):
Looks good!

PR #1633, opened 2024-02-09
Modal guidance

tesk9:

Fixes https://linear.app/noredink/issue/A11-4104/move-modal-guidance-from-text-field-to-about-section

Clarifies https://noredink.slack.com/archives/C02NMHB1K55/p1707433597293789?thread_ts=1707432784.790109&cid=C02NMHB1K55

image

### DIFF
--- a/component-catalog/src/Examples/Modal.elm
+++ b/component-catalog/src/Examples/Modal.elm
@@ -186,7 +186,7 @@ example =
     , about =
         Guidance.useATACGuide moduleName
             ++ [ Text.smallBody
-                    [ Text.markdown "Generally, you'll want to pair the `Modal.warning` theme with the `Button.danger` theme and the `Modal.info` theme with the `Button.primary` theme."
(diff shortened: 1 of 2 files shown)


bendansby (approved):
I'm not sure you actually need this since it's more a design-level thing:

> Pair the Modal. warning theme with the Button. danger theme and the Modal. info theme with the Button.primary theme.

tesk9:
I buy that! I was wondering a bit about whether it made more sense to link to the styleguide anyway. Or I can cut it 🤔 

@krinhorn what do you think?

krinhorn:
I like the idea of linking to the style guide if there's something about this in there. Otherwise I think @bendansby  & Design are in a better position to say if this kind of note is needed in the CC or if they specify this in their design specs/mockups.

krinhorn (approved):
Oops, didn't realize I was set as a reviewer! Looks good other than that last comment for Ben to address. Thank you!

tesk9:
✂️  image ✂️

bendansby:
That looks good. I think what modal styles to use will be adequately conveyed in each design.

PR #1634, opened 2024-02-09
Major release - clean up

tesk9:
- Adjusts APIs for Keyboard and Mouse spec Helpers to make them convenient to use. This is possible because https://github.com/NoRedInk/NoRedInk/pull/43978 merged & now monorepo elm apps are on the same version of test and elm-program-test as noredink-ui.
- Removes deprecated & unused modules
- Removes deprecated & unused functions
- Removes `Colors.yellow`. This _is_ used, but it's a direct alias for `Colors.mustard`, so the upgrade path is very simple.
- Removes `narrowMobileBreakPoint`, which is also a direct alias (for `narrowMobileBreakpoint`) with a simple upgrade path.

Fixes A11-3678. As it happens, the keyboard and mouse helpers have been available from noredink-ui this whole time, but they weren't wildly convenient to use. Once this version is out, the helpers will be simple to use. After this releases, I will add some example uses to the monorepo for discoverability.


charbelrami (approved):
wow! :D

PR #1639, opened 2024-02-12
Add demos & other reference to About sections of components

krinhorn:
Fixes A11-4110

Component catalog content changes only (no component/API changes).

Note that this PR gets all of the remaining "Components" demos linked in the CC. It does not link the "Frontend infrastructure" demos, though I made Linear issues in Triage with some ideas for linking those up in the CC.

### Reviewer, please check the following and feel free to skip any that don't seem important:
- [x] Generally, feel free to fix any bad Elm syntax.
- [x] I wasn’t always sure how to concatenate Guidances into the About sections - I just messed with ++ and [] until the compiler was happy. Feel free to fix them up.
- [x] On the **Accordion** page, I would be happy for the `Guidance.communicateState moduleName` to be the last bullet in the bulleted list, but I couldn’t figure out how to make that work.
- [x] Not super important, but on **Colors**, I don’t know how to fix: heading “General Colors” smushes into the about section.
- [x] On **FocusRing**, I had to escape some quote characters. I just guessed at how to do this and the compiler was happy, but please adjust if there is a better way.
- [x] On **FocusRing**, I wanted to link to Focus Loop but I couldn’t get the fancy `Route` stuff to work so I just used a relative URL. Feel free to change this.
- [x] On **Heading**, I had to import import Html.Styled, but I wasn’t sure the best way to do this in terms of what to expose. Please feel free to fix if there’s a better way.
- [x] For the **Message Guidance**, it could be nice to style or at least bold “errorMessage” and “guidance” (in relation to form elements.) I don’t know how to do that.
- [x] Did I miss any **form elements** that use errorMessage and/or guidance?
- [ ] **Focus loop** page could probably use a link to the [focus loop demo ](https://noredink.zoom.us/rec/play/tz3NTCCHbF1O8JE9SM1xwSFHXUcNONfUuSX8slrGOaO334YqZIFsY4AEYoh15WbW70okUdIfB8ux9E3i.56iPsgPTPEyAp9wU?canPlayFromShare=true&from=share_recording_detail&startTime=1691679791000&componentName=rec-play&originRequestUrl=https%3A%2F%2Fnoredink.zoom.us%2Frec%2Fshare%2Fr7Nq3cXaQ8ksO23LsQ14qg47OigfodJZEaKQTH0usMOMSp8ycs6RvixjgJOUgXIQ.fneI7G9r-o86ptDc%3FstartTime%3D1691679791000)but I didn’t know how to do that and I didn’t try. (It's not linked from any other component at the moment.)
- [ ] In **Percy**, for some reason Accordion and FocusRing show missing/fallback fonts. Locally & in the deploy preview, both of these pages are showing the correct font for me.
- Note that the **Tooltip** update is in the usage table, not in an About section.


tesk9 (commented):
Awesome PR, thank you! 🎉 

So far I have gone through the checkboxes that I checked off. I haven't reviewed the copy or links yet but I'll get to it tomorrow. The CC is getting so polished! ✨

component-catalog/src/Examples/FocusRing.elm
```diff
@@ -47,11 +50,48 @@ example =
         , example_ FocusRing.tightStyles
         ]
     , about =
-        [ Text.smallBody
+        [ Text.mediumBody
```
tesk9, line 53: Should this be smallBody?

component-catalog/src/Examples/QuestionBox.elm
```diff
@@ -57,7 +59,18 @@ example =
             , QuestionBox.correct
             ]
         ]
-    , about = []
+    , about =
+        [ Text.smallBody
+            [ Text.html
+                [ text "Learn more about this component from "
+                , ClickableText.link "Tessa's demo"
+                    [ ClickableText.linkExternal "https://www.dropbox.com/preview/NRI%20Engineering/Demos/2022-12-22%20-%20Tessa%20-%20QuestionBox%20and%20Block.mp4?role=work"
```
tesk9, line 67: I'm not sure that we should link to this demo for QuestionBox, as the question box has changed a _lot_ since this point.

In my demo:
image

Currently:
image

tesk9 (commented):
2 small questions/comments, and then I think this is ready to merge! 🎉

component-catalog/src/Examples/QuestionBox.elm
krinhorn: Fine with me - I didn't review all of the demos. (If you are able to make the change to remove it, I would appreciate that!)

component-catalog/src/Examples/FocusRing.elm
krinhorn: Oh I don't know actually! I wasn't sure when we were using one vs. the other. Feel free to change!

tesk9 (approved):

krinhorn:
Woohoo, thank you!

PR #1640, opened 2024-02-13
Improve carousel example

tesk9:
Touches the Component Catalog only.

- improves the Ellie sample code for Carousel
    - we're running into a limit of how long our urls that we send to Ellie can be, so I don't think I can make it much better than it is currently 🤷‍♀️ 
- adds `Tabs` category to the CC for convenience
- adds descriptions of each view type to the Carousel example

### Before

noredink-ui netlify app_ (6)

### After

localhost_8000_ (16)

### DIFF
--- a/component-catalog/src/Category.elm
+++ b/component-catalog/src/Category.elm
@@ -32,6 +32,7 @@ type Category
     | Animations
     | Progress
     | Navigation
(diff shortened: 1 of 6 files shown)


charbelrami (approved):
Looks great!

PR #1641, opened 2024-02-13
Improve Balloon example

tesk9:

### Before

noredink-ui netlify app_ (7)

### After

localhost_8000_ (17)


bendansby (approved):
Quite pneumatic!

charbelrami (approved):
Looks good!

PR #1662, opened 2024-03-29
Make Highlighter views lazy

juanedi:
## Context

Makes `Highlighter.V5` use `Html.Styled.lazy` by default for all its views.

This is motivated by performance issues on the Guided Drafts inline comments highlighter.

Among other things, we noticed we were rendering the highlighter unnecessarily a lot. Making the page generally sluggish even on interactions that didn't touch the highlighter at all. In that page the gain also happens to make the highlighter itself way faster, because we have multiple highlighters in the page (so operating a highlighter now updates just the one you need, not all of them).

We first thought about adding `lazy` just to that use case, but figured there might be value in applying it generally. 

**_Why we would want to make this the default:_**
  - Highlighter views are pretty heavy
  - It's hard to note when things start getting slow, and then dedicate time to changing it
    - For example: the GD highlighter has been super slow for years and we just managed to focus our efforts on making it better now!
  - In practice, there shouldn't be any downsides in doing this (see note below)

**_Why we would NOT want to make this the default:_**
  - The only potential downside we (well, Micah!) could think of is that `lazy` forces elm-css to build its stylesheets and add a `<style>` tag next to the lazy node. This means that if you have N lazy highlighters you will end up with N `<style>` tags, probably with lots of duplicated classes. This is known to be slow in Chrome for large Ns (>100s), but I think it's very unlikely that we'll use the highlighter this way. If we ever do, we can always change the API to allow disabling the default lazy behavior.

**Results**

This change made the GD highlighter WAAAAY more responsive, as you can see in the videos below.

Here are profiler results for hovering out of a word, for example. Note how we went from 82ms to 5ms to render the new state:
Screenshot 2024-03-29 at 08 44 03
Screenshot 2024-03-29 at 08 44 33

These videos were taken on my fast dev laptop with 4x CPU throttling, compiling elm without `--debug`.

### Before `lazy`

https://github.com/NoRedInk/noredink-ui/assets/753421/6440413d-495c-48e8-9c18-f4406ba65275

### After `lazy`

https://github.com/NoRedInk/noredink-ui/assets/753421/f71d1037-0c3a-4eeb-bd91-5fac75c43209

    - an accurate preview
    - valid sample code
    - correct keyboard behavior

### DIFF
--- a/src/Nri/Ui/Highlighter/V5.elm
+++ b/src/Nri/Ui/Highlighter/V5.elm
@@ -859,19 +859,22 @@ highlightLengths model =
 
 {-| -}
 view : Model marker -> Html (Msg marker)
-view model =
-    view_
-        { showTagsInline = False
-        , maybeTool = Just model.marker
-        , mouseOverIndex = model.mouseOverIndex
-        , mouseDownIndex = model.mouseDownIndex
-        , hintingIndices = model.hintingIndices
-        , overlaps = False
-        , viewSegment = viewHighlightable { renderMarkdown = False, overlaps = False } model
-        , id = model.id
-        , highlightables = model.highlightables
-        , sorter = Just model.sorter
-        }
+view =
+    lazy
+        (\model ->
+            view_
+                { showTagsInline = False
+                , maybeTool = Just model.marker
+                , mouseOverIndex = model.mouseOverIndex
+                , mouseDownIndex = model.mouseDownIndex
+                , hintingIndices = model.hintingIndices
+                , overlaps = False
+                , viewSegment = viewHighlightable { renderMarkdown = False, overlaps = False } model
+                , id = model.id
+                , highlightables = model.highlightables
+                , sorter = Just model.sorter
+                }
+        )
 
 
 {-| -}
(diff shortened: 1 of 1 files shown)


bcardiff (approved):

erik-nri (approved):

omnibs (changes requested):
Can we hold off merging? im not seeing the perf gains here

omnibs:
I'll dig into it for a bit with Juan in a few to try to figure out what up

caseyWebb:
> The only potential downside we (well, Micah!) could think of is that lazy forces elm-css to build its stylesheets and add a <style> tag next to the lazy node. This means that if you have N lazy highlighters you will end up with N <style> tags, probably with lots of duplicated classes. This is known to be slow in Chrome for large Ns (>100s), but I think it's very unlikely that we'll use the highlighter this way. If we ever do, we can always change the API to allow disabling the default lazy behavior.

Just throwing out there that `Css.Global` is another option, albeit a more significant departure from existing practices.

juanedi:
> > The only potential downside we (well, Micah!) could think of is that lazy forces elm-css to build its stylesheets and add a <style> tag next to the lazy node. This means that if you have N lazy highlighters you will end up with N <style> tags, probably with lots of duplicated classes. This is known to be slow in Chrome for large Ns (>100s), but I think it's very unlikely that we'll use the highlighter this way. If we ever do, we can always change the API to allow disabling the default lazy behavior.
> 
> Just throwing out there that `Css.Global` is another option, albeit a more significant departure from existing practices.

Thanks! Yeah, we've been experimenting with that too :-)

Even leaving out the effect of `lazy`, we want to see how much time we save by moving some/most/all of the styling to `Css.Global` assuming there's potential gains from:
  - simplifying vdom diffing (changing class name attributes rather than css code in `<style>` nodes)
  - avoiding the hashing for automatic class name generation based on styles

We'll report on results, but it looks like it's kind of a big change so there's a chance we'll drop it if we find that things are already fast enough.

PR #1668, opened 2024-04-17
[PHX-1513] Add fold functions to Highlighter on noredink-ui

micahhahn:

## Context

We have for a long time dreamed of a version of highlighter on the writing team that would allow us to highlight rich text documents with formatting, lists, etc...  The best we could do was some awful hacks to approximate a list like structure which are very broken and most definitely bad for a11y. 

This PR carves a new way forward.  Instead of having a list of `Highlightable` elements that we pass off to the Highlighter library to render into a single `<p>` tag, we now have the control to render the `Highlightable`s one by one if we wish.  This gives the consumer of the Highlighter library the power to render HTML structure on their own and call out to Highlighter only to render individual segments.  

Fixes PHX-1513

I implemented a spike in the following PR to show that an approach where the consumer of the Highlightable library has more power over the rendering of the content.

- https://github.com/NoRedInk/NoRedInk/pull/47982

No direct changes to view from this.  The video presented [on this PR](https://github.com/NoRedInk/NoRedInk/pull/47982) is a good indication for how we plan to use the new functions in team Phoenix. 

    - an accurate preview
    - valid sample code
    - correct keyboard behavior

### DIFF
--- a/src/Nri/Ui/Highlighter/V5.elm
+++ b/src/Nri/Ui/Highlighter/V5.elm
@@ -1140,6 +1140,8 @@ findOverlapsSupport model =
         }
 
 
+{-| A type that contains information needed to render individual `Highlightable`s one at a time
+-}
 type FoldState marker
     = FoldState
         { model : Model marker
@@ -1148,6 +1150,8 @@ type FoldState marker
         }
 
 
+{-| Computes all the mark styles necessary to perform a fold over the `Highlightable` elements
+-}
 initFoldState : Model marker -> FoldState marker
 initFoldState model =
     let
@@ -1189,6 +1193,11 @@ initFoldState model =
         }
 
 
+{-| Render a single `Highlightable` while also returning an updated state.
+
+A list of extraStyles is also accepted if, for example, you want to apply bold / italic / underline formatting to the generated span.
+
+-}
 viewFoldHighlighter : FoldState marker -> List Css.Style -> ( List (Html (Msg marker)), FoldState marker )
 viewFoldHighlighter (FoldState ({ model, overlapsSupport } as foldState)) extraStyles =
     viewFoldHelper
@@ -1202,6 +1211,11 @@ viewFoldHighlighter (FoldState ({ model, overlapsSupport } as foldState)) extraS
         extraStyles
 
 
+{-| Render a single `Highlightable` that is NOT interactive while also returning an updated state.
+
+A list of extraStyles is also accepted if, for example, you want to apply bold / italic / underline formatting to the generated span.
+
+-}
 viewFoldStatic : FoldState marker -> List Css.Style -> ( List (Html msg), FoldState marker )
 viewFoldStatic =
     viewFoldHelper
--- a/src/Nri/Ui/Highlighter/V5.elm
+++ b/src/Nri/Ui/Highlighter/V5.elm
@@ -915,30 +915,8 @@ viewWithOverlappingHighlights =
     lazy
         (\model ->
             let
-                hoveredMarkers =
-                    model.highlightables
-                        |> List.Extra.find (\h -> Just h.index == model.mouseOverIndex)
-                        |> Maybe.map (.marked >> List.map .kind)
-                        |> Maybe.withDefault []
-
                 overlaps =
-                    OverlapsSupported
-                        { hoveredMarkerWithShortestHighlight =
-                            case hoveredMarkers of
-                                [] ->
-                                    Nothing
-
-                                marker :: [] ->
-                                    Just marker
-
-                                first :: second :: rest ->
-                                    Just
-                                        (markerWithShortestHighlight
-                                            model.sorter
-                                            model.highlightables
-                                            ( first, second, rest )
-                                        )
-                        }
+                    findOverlapsSupport model
             in
             view_
                 { showTagsInline = False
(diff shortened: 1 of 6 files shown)


micahhahn:
@charbelrami I tagged you for review if you want to take a look.  Since the new functions re-use the code from `viewWithOverlappingHighlights`, they should no less accessible than `viewWithOverlappingHighlights` is.  But a review wouldn't hurt as you're the only a11ybat left with experience in this code.

juanedi (commented):
SO EXCITING!!!

I'm just beginning to review, but leaving one preliminary thought: would it make sense to add a demo of this new rendering method to the component Catalog? Ideally one that shows how this one is more powerful than what we had before -- like properly rendering lists?.

I think that would really help clarify the value of the new function. Also, having a canonical realistic example of how to use this would probably help us a lot in the future.

micahhahn:
> would it make sense to add a demo of this new rendering method to the component Catalog?

src/Nri/Ui/Mark/V6.elm
```diff
@@ -69,72 +74,100 @@ viewWithOverlaps :
...
+                            Set.dropIf (Set.memberOf currentMarks) state.priorMarks
+                                |> Set.toList
+
+                        startStyles =
+                            case startedMarks of
+                                [] ->
+                                    []
+
+                                _ ->
+                                    tagBeforeContent startedMarks :: List.concatMap .startStyles startedMarks
```
juanedi, line 133: 🎨 super minor but we can get rid of the special treatment for `[]` here if we make `tagBeforeContent` return a List of styles rather than a single style (getting rid of the problematic `Css.batch []`).

src/Nri/Ui/Mark/V6.elm
```diff
@@ -69,72 +74,100 @@ viewWithOverlaps :
...
+
+                ( _, [] ) ->
+                    patchRevStyles
+
+                ( ( prevContent, prevLabel, prevStyles ) :: otherPrevStyles, _ ) ->
+                    ( prevContent, prevLabel, prevStyles ++ (tagAfterContent endedMarks :: List.concatMap .endStyles endedMarks) ) :: otherPrevStyles
+
+        { priorMarks, revStyles } =
+            List.foldl
+                (\( content, marks ) state ->
```
juanedi, line 114: 🎨 maaaaybe rename `marks` to `currentMarksList` or similar? might help understanding with so many collections of marks floating around.

src/Nri/Ui/Mark/V6.elm
```diff
@@ -69,72 +74,100 @@ viewWithOverlaps :
...
+                ( [], _ ) ->
+                    patchRevStyles
+
+                ( _, [] ) ->
+                    patchRevStyles
+
+                ( ( prevContent, prevLabel, prevStyles ) :: otherPrevStyles, _ ) ->
+                    ( prevContent, prevLabel, prevStyles ++ (tagAfterContent endedMarks :: List.concatMap .endStyles endedMarks) ) :: otherPrevStyles
+
+        { priorMarks, revStyles } =
```
juanedi, line 112: `revStyles` isn't actually reversed, or is it?˙ 🤔

src/Nri/Ui/Highlighter/V5.elm
```diff
@@ -1123,6 +1112,155 @@ staticMarkdownWithTags =
...
+
+        markStyles =
+            model.highlightables
+                |> List.map (\highlightable -> ( highlightable, List.map (toMark highlightable) highlightable.marked ))
+                |> Mark.overlappingStyles
+    in
+    FoldState
+        { model = model
+        , overlapsSupport = overlapsSupport
+        , state = markStyles
```
juanedi, line 1192: 💭 reading this line it seems pretty evident to me that `markStyles` and `overlappingStyles` really downplay the importance of we are doing here! like... when i read it my mind went "oh just computes a bunch of styles" when in reality we're building an intermediate representation that is basically the model we'll be rendering.

sharing a thought without a suggestion because TBH i don't know how I would improve it. feel free to ignore this 😛

juanedi (approved):
daaaaaaaaaaang what a work of art! 😻
left a bunch of minor suggestions. take what you want!

src/Nri/Ui/Mark/V6.elm
micahhahn: Good idea!

src/Nri/Ui/Mark/V6.elm
micahhahn: Done!

src/Nri/Ui/Mark/V6.elm
micahhahn: It is! We are folding from the left here over the segments, so `revStyles` is being built up backwards.  We then call `List.reverse` that the very end of the function.

juanedi (approved):

brian-carroll (approved):
Nice! 😍

tests/Spec/Nri/Ui/Mark.elm
```diff
@@ -0,0 +1,377 @@
...
+    else
+        [ Css.after
+            [ cssContent (highlightDescription "end" marks)
+            , invisibleStyle
+            ]
+        ]
+
+
+{-| Mark.overlappingStyles gives us a (Maybe (Html msg)) for the label outputs,
+but we can't test equality on HTML elements very well so we need to drop it.
```
omnibs, line 111: Maybe we could map the `Maybe` to a `Just MarkStartLabel | Nothing`?

There's one thing that's confusing to me about the tests below which is why `startStyles` are sometimes missing, and my guess is sometimes they're on this label, sometimes they're not.

omnibs: ps: I did a lazy attempt at this [here](https://github.com/NoRedInk/noredink-ui/pull/1669)

feel free to take it or leave it

omnibs (approved):
This looks great!

Left one tiny suggestion on tests. Feel free to ignore, since other tests already cover start tag presence (I tried removing the code that adds them and 15 tests failed).

tests/Spec/Nri/Ui/Mark.elm
micahhahn: Ok, so this comment is old and needs to be updated - I had an incorrect understanding of what was happening before... 

If a mark has a name (which they really should for a11y purposes), then that is what gets put into the `Maybe (Html msg)` element as invisible content.  You can make the mark names show by calling `viewWithInlineTags` instead.  

So the start styles are **always** present, but if the mark has a name then they are applied to the invisible prefix element - if there is no name than the styles are attached to the first word itself. 

I'm going to see if I can make this clearer in the comments

PR #1680, opened 2024-05-28
Drop Nri.Ui.Tabs.V8

juanedi:
Tabs.V8 was replaced by Tabs.V9.

V8 won't be used anywhere in our codebase once https://github.com/NoRedInk/NoRedInk/pull/48542 merges. That means we can safely remove it.


ArthurJordao (commented):
Hey, you just forgot to change it here 😬:
https://github.com/NoRedInk/noredink-ui/blob/7d0fcc0d7b951736cb61c9bfc91c2d6b50dea337/component-catalog/src/Examples/Tabs.elm#L42

juanedi:
> Hey, you just forgot to change it here 😬:
> 
> https://github.com/NoRedInk/noredink-ui/blob/7d0fcc0d7b951736cb61c9bfc91c2d6b50dea337/component-catalog/src/Examples/Tabs.elm#L42

Uuuuuu good catch! I wonder how hard it would be to write an elm-review rule to catch these.

ArthurJordao (approved):

PR #1689, opened 2024-07-25
Highlighter.V5: Expose KeyboardMsg

omnibs:
This will allow us to fine-tune behavior at application site.

Specific scenario we want to cover:
- We want to block overlapping/merging/growing higlights with the mouse
- We still want to allow sentence highlights with the keyboard

Sentence highlighting with the keyboard requires expanding a highlight, so we want to be able to detect it and special-case it.

Ps: I think this will also be necessary to add keyboard navigation to highlighter-based components, where we need to bypass default keyboard behavior

ps2: don't be intimidated by the 2k LoC diff, I had to bump elm-review for CI to pass, and for that I had to make an overlay, and node package overlays are big bc so many dependencies


micahhahn (approved):
Looks fine to me! Maybe want to have a 🦊 take a look at the nix changes?

omnibs:
@zwilias tagging you for the elm-review overlay

zwilias (approved):
Overlays looks perfect!

One small caveat - `nix-shell -p` doesn't take packages from the nixpkgs input of a flake, but from the global `<nixpkgs>` ref, and we have no control over that. So we may end up executing different versions of `node2nix` depending on who's executing the process to update `elm-review`. _probably_ not going to be an issue, but it's worth noting, and adding `node2nix` to the shell closure packages would prevent that from happening. Anyway, happy to merge as is!

omnibs:
@micahhahn do you mind merging if you plan on cutting a new version of
nri-ui today?

I can follow up on the node2nix on Monday.

Just do I don't miss the train :S

micahhahn:
Was just about to start the bump! :)

PR #1705, opened 2024-11-01
quo 987 sortabletable consistent default padding for header and data

ArthurJordao:

## Context
This adds more flexibility to the table api and add better table defaults, which includes:
- Disable alternating row columns.
- Add placeholder columns
- Add consistent default padding for header.
- Show sorting indicator next to label.

image
Screenshot 2024-11-01 at 17 55 38

image

    - an accurate preview
    - valid sample code
    - correct keyboard behavior

closes: quo-987
closes: quo-989
closes: quo-990
closes: quo-991

### DIFF
--- a/component-catalog/src/Examples/Table.elm
+++ b/component-catalog/src/Examples/Table.elm
@@ -17,7 +17,7 @@ import Html.Attributes exposing (alt)
 import Nri.Ui.Button.V10 as Button
 import Nri.Ui.Heading.V3 as Heading
 import Nri.Ui.Spacing.V1 as Spacing
-import Nri.Ui.Table.V7 as Table exposing (Column)
+import Nri.Ui.Table.V8 as Table exposing (Column)
 
 
 {-| -}
@@ -32,7 +32,7 @@ moduleName =
 
 version : Int
 version =
-    7
+    8
 
 
 {-| -}
--- a/component-catalog/src/Examples/Table.elm
+++ b/component-catalog/src/Examples/Table.elm
@@ -46,7 +46,7 @@ example =
     , categories = [ Layout ]
     , keyboardSupport = []
     , preview =
-        [ Table.view { additionalStyles = [], alternatingRowColors = True }
+        [ Table.view []
             [ Table.string
                 { header = "A"
                 , value = .a
@@ -125,16 +125,16 @@ example =
                 ]
             , case ( showHeader, isLoading ) of
                 ( True, False ) ->
-                    Table.view { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns data
+                    Table.view [] columns data
 
                 ( False, False ) ->
-                    Table.viewWithoutHeader { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns data
+                    Table.viewWithoutHeader [] columns data
 
                 ( True, True ) ->
-                    Table.viewLoading { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns
+                    Table.viewLoading [] columns
 
                 ( False, True ) ->
-                    Table.viewLoadingWithoutHeader { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns
+                    Table.viewLoadingWithoutHeader [] columns
             ]
     }
 
--- a/component-catalog/src/Examples/Table.elm
+++ b/component-catalog/src/Examples/Table.elm
@@ -169,45 +169,6 @@ example =
                     }
                 ]
                 data
-            , Heading.h3 [ Heading.plaintext "Without placeholderColumn", Heading.css [ Css.marginTop (Css.px 30) ] ]
-            , Table.view []
-                [ Table.rowHeader
-                    { header = text "User ID"
-                    , view = text << String.fromInt << .userId
-                    , width = Css.px 80
-                    , cellStyles = always []
-                    , sort = Nothing
-                    }
-                , Table.string
-                    { header = "First Name"
-                    , value = .firstName
-                    , width = Css.calc (Css.pct 50) Css.minus (Css.px 250)
-                    , cellStyles = always []
-                    , sort = Nothing
-                    }
-                , Table.string
-                    { header = "Last Name"
-                    , value = .lastName
-                    , width = Css.calc (Css.pct 50) Css.minus (Css.px 250)
-                    , cellStyles = always []
-                    , sort = Nothing
-                    }
-                , Table.string
-                    { header = "Submitted"
-                    , value = .submitted >> String.fromInt
-                    , width = Css.px 125
-                    , cellStyles = always [ Css.textAlign Css.center ]
-                    , sort = Nothing
-                    }
-                , Table.custom
-                    { header = text "Actions"
-                    , width = Css.px 250
-                    , view = \_ -> Button.button "Action" [ Button.small ]
-                    , cellStyles = always []
-                    , sort = Nothing
-                    }
-                ]
-                data
             ]
     }
 
--- a/component-catalog/src/Examples/Table.elm
+++ b/component-catalog/src/Examples/Table.elm
@@ -13,6 +13,7 @@ import Css
 import Debug.Control as Control exposing (Control)
 import Debug.Control.View as ControlView
 import Example exposing (Example)
+import Html.Attributes exposing (alt)
 import Nri.Ui.Button.V10 as Button
 import Nri.Ui.Heading.V3 as Heading
 import Nri.Ui.Spacing.V1 as Spacing
@@ -73,7 +74,7 @@ example =
     , view =
         \ellieLinkConfig state ->
             let
-                { showHeader, isLoading } =
+                { showHeader, isLoading, alternatingRowColors } =
                     Control.currentValue state
 
                 ( columnsCode, columns ) =
@@ -89,7 +90,7 @@ example =
                 , extraCode = [ "import Nri.Ui.Button.V10 as Button" ]
                 , renderExample = Code.unstyledView
                 , toExampleCode =
-                    \settings ->
+                    \_ ->
                         let
                             codeWithData viewName =
                                 List.map datumToString data
@@ -100,7 +101,14 @@ example =
                                 { sectionName = moduleName ++ "." ++ viewName
                                 , code =
                                     (moduleName ++ "." ++ viewName)
-                                        ++ " [] "
+                                        ++ " "
+                                        ++ Code.list
+                                            (if alternatingRowColors then
+                                                []
+
+                                             else
+                                                [ "Table.disableAlternatingRowColors" ]
+                                            )
                                         ++ Code.list columnsCode
                                         ++ dataStr
                                 }
@@ -117,16 +125,16 @@ example =
                 ]
             , case ( showHeader, isLoading ) of
                 ( True, False ) ->
-                    Table.view { additionalStyles = [], alternatingRowColors = True } columns data
+                    Table.view { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns data
 
                 ( False, False ) ->
-                    Table.viewWithoutHeader { additionalStyles = [], alternatingRowColors = True } columns data
+                    Table.viewWithoutHeader { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns data
 
                 ( True, True ) ->
-                    Table.viewLoading { additionalStyles = [], alternatingRowColors = True } columns
+                    Table.viewLoading { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns
 
                 ( False, True ) ->
-                    Table.viewLoadingWithoutHeader { additionalStyles = [], alternatingRowColors = True } columns
+                    Table.viewLoadingWithoutHeader { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns
             ]
     }
 
@@ -150,6 +158,7 @@ update msg state =
 type alias Settings =
     { showHeader : Bool
     , isLoading : Bool
+    , alternatingRowColors : Bool
     }
 
 
@@ -158,6 +167,7 @@ controlSettings =
     Control.record Settings
         |> Control.field "visible header" (Control.bool True)
         |> Control.field "is loading" (Control.bool False)
+        |> Control.field "alternatingRowColors" (Control.bool True)
 
 
 type alias Datum =
--- a/component-catalog/src/Examples/Table.elm
+++ b/component-catalog/src/Examples/Table.elm
@@ -166,10 +166,9 @@ example =
                         )
                         columns
             , Heading.h2
-                [ Heading.plaintext "Using placeholderColumn for consistent column alignment"
+                [ Heading.plaintext "Using placeholder columns for consistent column alignment"
                 , Heading.css [ Css.marginTop (Css.px 30) ]
                 ]
-            , Heading.h3 [ Heading.plaintext "With placeholderColumn", Heading.css [ Css.marginTop (Css.px 30) ] ]
             , Text.smallBody
                 [ Text.plaintext
                     "By default tables will distribute any remaining space across columns. You can change this behavior to have fixed column widths and leave any remaining blank space to the right by using empty placeholder columns with placeholderColumn. This is particularly useful when you need to align similar columns across multiple tables."
--- a/component-catalog/src/Examples/Table.elm
+++ b/component-catalog/src/Examples/Table.elm
@@ -125,16 +125,46 @@ example =
                 ]
             , case ( showHeader, isLoading ) of
                 ( True, False ) ->
-                    Table.view [] columns data
+                    Table.view
+                        (if alternatingRowColors then
+                            []
+
+                         else
+                            [ Table.disableAlternatingRowColors ]
+                        )
+                        columns
+                        data
 
                 ( False, False ) ->
-                    Table.viewWithoutHeader [] columns data
+                    Table.viewWithoutHeader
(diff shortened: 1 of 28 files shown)


src/Nri/Ui/Table/V7.elm
```diff
@@ -1,4 +1,4 @@
-module Nri.Ui.Table.V7 exposing
+module Nri.Ui.Table.V8 exposing
```
juanedi: This should be `V7`, right?

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+viewRow columns alternatingRowColors data =
+    tr
+        [ css (rowStyles alternatingRowColors) ]
+        (List.map (viewColumn data) columns)
+
+
+viewColumn : data -> Column data msg -> Html msg
+viewColumn data (Column _ renderer width cellStyles _ cellType _) =
+    cell cellType
+        [ css ([ width, verticalAlign middle, padding4 (px 11) (px 12) (px 14) (px 12), textAlign left ] ++ cellStyles data)
```
juanedi, line 146: This adds top and bottom padding to internal cells. Are you sure we want that? My understanding is that we just wanted to align the start of headers and content cells:

Before:
Screenshot 2024-11-04 at 12 18 40 PM



After:
Screenshot 2024-11-04 at 12 18 44 PM

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+-}
+viewWithoutHeader : { additionalStyles : List Style, alternatingRowColors : Bool } -> List (Column data msg) -> List data -> Html msg
+viewWithoutHeader { additionalStyles, alternatingRowColors } columns =
+    tableWithoutHeader additionalStyles columns (viewRow columns alternatingRowColors)
+
+
+{-| Displays a table of data based on the provided column definitions
+-}
+view : { additionalStyles : List Style, alternatingRowColors : Bool } -> List (Column data msg) -> List data -> Html msg
+view { additionalStyles, alternatingRowColors } columns =
```
juanedi, line 132: 💭 Since we are introducing a new version of the component, did you consider switching it to using attributes like we do in many other modules? Those are a bit more upgrade, friendly. So we could have two to start with:

```elm
css : List Style -> Attribute

-- defaults to True, which is the current behavior
-- could also be `noAlternatingRowColors : Attribute`
alternatingRowColors : Bool -> Attribute
```

Might want to check with the frontend guild before doing it, but I think it might be nice! If we already had this for table, this could have been a minor change to the module without forcing us to introduce a new version.

juanedi: OH, a bit like what you did in the SortableTable wrapper! I'm reviewing commit-by-commit leaving comments as I go :-P

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+rowStyles alternatingRowColors =
+    [ height (px 45)
+    , fontSize (px 14)
+    , color gray20
+    , if alternatingRowColors then
+        pseudoClass "nth-child(odd)"
+            [ backgroundColor gray96 ]
+
+      else
+        Css.batch []
```
juanedi, line 291: Note that when there are no alternating row colors we want to show a thin gray border between rows:

Code and screenshot from insights:
https://github.com/NoRedInk/NoRedInk/blob/b54b50b77be0fc6217c4d754f9b3f4fa3f2953a2/monolith/ui/src/Nri/Writing/ClassResults.elm#L840
Screenshot 2024-11-04 at 12 33 23 PM

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+import Html.Styled.Attributes as Attributes exposing (css)
+import Nri.Ui.Colors.V1 exposing (..)
+import Nri.Ui.Fonts.V1 exposing (baseFont)
+
+
+{-| Closed representation of how to render the header and cells of a column
+in the table
+-}
+type Column data msg
+    = Column (Html msg) (data -> Html msg) Style (data -> List Style) (Maybe SortDirection) CellType Bool
```
juanedi, line 36: All other attributes don't make much sense when representing a placeholder column. Does that mean we should have two separate constructors?

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+    }
+    -> Column data msg
+string { header, value, width, cellStyles, sort } =
+    Column (Html.text header) (value >> Html.text) (Css.width width) cellStyles sort DataCell False
+
+
+{-| Creates a placeholder column which will reserve the space for the column,
+this can be used when multiple tables have similar data and we want to keep
+the size consistent.
+-}
```
juanedi, line 81: 🎨 Can we document that using this attribute will make the whole table switch to table-layout: fixed?

src/Nri/Ui/SortableTable/V4.elm
```diff
@@ -230,6 +241,24 @@ custom config =
...
         , width = config.width
         , cellStyles = config.cellStyles
+        , hidden = False
+        }
+
+
+{-| Creates a placeholder column which will reserve the space for the column,
+this can be used when multiple tables have similar data and we want to keep
+the size consistent.
+-}
```
juanedi, line 251: Same comment re. documenting that this switches to `table-layout: fixed`. Or maybe we can just say `See <link to docs for Table.placeholderColumn>`?

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+custom :
+    { header : Html msg
+    , view : data -> Html msg
+    , width : LengthOrAuto compatible
+    , cellStyles : data -> List Style
+    , sort : Maybe SortDirection
+    }
+    -> Column data msg
+custom options =
+    Column options.header options.view (Css.width options.width) options.cellStyles options.sort DataCell True
```
juanedi, line 101: Does the `True` at the end mean that `custom` will make the column hiden?

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+{-| Creates a placeholder column which will reserve the space for the column,
+this can be used when multiple tables have similar data and we want to keep
+the size consistent.
+-}
+placeholderColumn :
+    { width : LengthOrAuto compatible
+    }
+    -> Column data msg
+placeholderColumn options =
+    Column (Html.text "") (always (Html.text "")) (Css.width options.width) (always []) Nothing DataCell False
```
juanedi, line 87: Shouldn't this have a `True` at the end?

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+tableStyles columns =
+    let
+        hasPlaceholderColumns =
+            List.any (\(Column _ _ _ _ _ _ h) -> h) columns
+    in
+    [ borderCollapse collapse
+    , baseFont
+    , Css.width (Css.pct 100)
+    , if hasPlaceholderColumns then
+        Css.tableLayout Css.fixed
```
juanedi, line 327: I think that there's value in supporting `fixed` layout even if we are not using placeholder columns. Should we just make that a separate attribute?

src/Nri/Ui/Table/V8.elm
juanedi: (Assuming we don't make the layout a separate attribute as suggested in some other comment)

src/Nri/Ui/Table/V8.elm
```diff
@@ -0,0 +1,368 @@
...
+
+{-| Creates a placeholder column which will reserve the space for the column,
+this can be used when multiple tables have similar data and we want to keep
+the size consistent.
+-}
+placeholderColumn :
+    { width : LengthOrAuto compatible
+    }
+    -> Column data msg
+placeholderColumn options =
```
juanedi, line 86: Note that placeholder columns need to add `aria-hidden` to both header and cells to make screenreaders skip those. We're just not doing that yet in the current version versions we needed to change the Table component, but it's requested in QUO-989.

component-catalog/src/Examples/Table.elm
```diff
@@ -13,10 +13,11 @@ import Css
 import Debug.Control as Control exposing (Control)
```
juanedi, line 13: Can we add an example showcasing the use case for placeholder columns?

juanedi (commented):
Yay, thanks for taking care of these! Left a bunch of comments.

src/Nri/Ui/Table/V7.elm
ArthurJordao: Yes, It's outdated already 😬

src/Nri/Ui/Table/V8.elm
ArthurJordao: Hmm, the title of the ticket "consistent default padding for header and data cells", that is why I used the same padding including vertical padding, but I can remove the vertical padding.

src/Nri/Ui/Table/V8.elm
ArthurJordao: Yes, I thought about it, and I actually moved to that at some point, but it was a little strange the API to communicate between modules, but thinking about it again I think would make sense. I will change it.

src/Nri/Ui/Table/V8.elm
ArthurJordao: Noted 👍

src/Nri/Ui/Table/V8.elm
ArthurJordao: Make sense. I will add a new constructor for that.

src/Nri/Ui/Table/V8.elm
ArthurJordao: Suer!

src/Nri/Ui/Table/V8.elm
ArthurJordao: Nice catch.

src/Nri/Ui/Table/V8.elm
ArthurJordao: sure!

src/Nri/Ui/Table/V8.elm
juanedi: @staceyadams can probably confirm what the desired behavior is!

src/Nri/Ui/Table/V8.elm
ArthurJordao: Isn't it already supported with additional styles?

juanedi: Oh, you're right. Feel free to ignore this, then.

nunocf:
Great work doing this.
Working with placeholder columns is a great scenario  for us to use CSS Grid, as we can ensure the layout is consistent in both row and column directions.
It is also widely supported across most browsers and very simple to make changes, and gives our design team the ability to use more than just pixels for laying out tables. For example, we could define that Column A takes twice as much as column B, while Column C takes half as much as Column B. This would keep layout consistent across multiple screen widths.

@juanedi Would this upgrade be something that makes sense to do here?

ArthurJordao:
@nunocf i am not sure how we should integrate grids with the table components. Would you mind explaining how the API would change to add compatibility to grids?
By looking at the current api it would be possible to the client add grids using custom css.

ArthurJordao:
@juanedi I believe this should be ready to review again :D

src/Nri/Ui/Table/V8.elm
```diff
@@ -176,11 +184,18 @@ viewLoadingRow columns alternatingRowColors index =
...
+        Column _ _ width _ _ cellType ->
+            cell cellType
+                [ css (stylesLoadingColumn rowIndex colIndex width ++ [ verticalAlign middle ] ++ loadingCellStyles)
+                ]
+                [ span [ css loadingContentStyles ] [] ]
+
+        PlaceholderColumn width ->
+            cell DataCell
+                [ css (stylesLoadingColumn rowIndex colIndex width ++ [ verticalAlign middle ] ++ loadingCellStyles) ]
+                [ span [ css loadingContentStyles ] [] ]
```
juanedi: 🎨 Shouldn't we skip the loading indicator in placeholder columns? Doesn't seem right to show a loading thingy signaling that content will load soon and then see it disappear.

src/Nri/Ui/Table/V8.elm
```diff
@@ -117,27 +119,86 @@ rowHeader options =
...
+-- Model
+
+
+type alias Config =
+    { css : List Style, alternatingRowColors : Bool }
+
+
+{-| Attribute to configure the table
+-}
+type Attr
```
juanedi, line 132: 🎨 We call this `Attribute` everywhere, can we change it here for consistency?

Screenshot 2024-11-06 at 12 14 29 PM

src/Nri/Ui/Table/V8.elm
```diff
@@ -117,27 +119,86 @@ rowHeader options =
...
+    { css = [], alternatingRowColors = True }
+
+
+{-| Add a CSS style to the table
+-}
+css : List Style -> Attr
+css styles =
+    Attr <|
+        \config ->
+            { config | css = styles ++ config.css }
```
juanedi, line 147: can we make add the new customizations at the end so they take precedence over default styles? that's how `additionalStyles` works in V7, I think.

component-catalog/src/Examples/Table.elm
```diff
@@ -125,16 +125,16 @@ example =
...
-                    Table.viewWithoutHeader { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns data
+                    Table.viewWithoutHeader [] columns data
 
                 ( True, True ) ->
-                    Table.viewLoading { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns
+                    Table.viewLoading [] columns
 
                 ( False, True ) ->
-                    Table.viewLoadingWithoutHeader { additionalStyles = [], alternatingRowColors = alternatingRowColors } columns
+                    Table.viewLoadingWithoutHeader [] columns
```
juanedi: I think we should conditionally add the `disableAlternatingRowColors` attribute here! The toggle isn't doing anything in the demo anymore 😅.

https://github.com/user-attachments/assets/e380244e-0c5f-48ad-883c-4f8a5df30f26

component-catalog/src/Examples/Table.elm
```diff
@@ -135,6 +134,80 @@ example =
...
+                    }
+                , Table.custom
+                    { header = text "Actions"
+                    , width = Css.px 250
+                    , view = \_ -> Button.button "Action" [ Button.small ]
+                    , cellStyles = always []
+                    , sort = Nothing
+                    }
+                ]
+                data
```
juanedi, line 210: What is the purpose of this last table? At a glance it doesn't seem to be any different from the "basic" example at the top, and it seems strange to have a "Without placeholderColumn" example in a section called "Using placeholderColumn [...]".

I think that adding the new "With placeholderColumn" table should be enough. It shows the new behavior clearly, and it also happens to align with the columns from the basic example (that could change in the future, of course, but I think that adding a third table just to show that they align isn't worth the noise).

component-catalog/src/Examples/Table.elm
```diff
@@ -135,6 +134,80 @@ example =
 
                 ( False, True ) ->
                     Table.viewLoadingWithoutHeader [] columns
+            , Heading.h2
+                [ Heading.plaintext "Using placeholderColumn for consistent widths tables"
```
juanedi, line 138: "for consistent column alignment", maybe?

I think it will be hard to condense all the information of the use case in a header, though. How about adapting the very nice docs you added for the new functions and adding it as a little intro blurb right below the header? Something like:

> By default tables will distribute any remaining space across columns. You can change this behavior to have fixed column widths and leave any remaining blank space to the right by using empty placeholder columns with `placeholderColumn`. This is particularly useful when you need to align similar columns across multiple tables.

component-catalog/src/Examples/Table.elm
```diff
@@ -135,6 +134,80 @@ example =
...
+                , Heading.css [ Css.marginTop (Css.px 30) ]
+                ]
+            , Heading.h3 [ Heading.plaintext "With placeholderColumn", Heading.css [ Css.marginTop (Css.px 30) ] ]
+            , Table.view []
```
juanedi, line 142: I love having this in the demo page. Thanks!

juanedi (commented):
This is looking awesome!
Just a few more minor suggested tweaks... thanks!! 🙏

component-catalog/src/Examples/Table.elm
ArthurJordao: lol, thank you for noticing

component-catalog/src/Examples/Table.elm
ArthurJordao: Yeah, make sense, it was more to showcase when to use it.

nunocf:
@juanedi @ArthurJordao I don't think my suggestion above should stop us from merging the PR. Thinking about it, it would probably cause us to have to rewrite the view  part of the `Table` module to use `grid` and that will definitely get this out of scope completely.
It would be an interesting project for the UX Guild to do, in order to make this component a little bit nicer to use, what do you think, @charbelrami ?

ArthurJordao:
@juanedi would you mind to taking another look at this one?

juanedi (approved):
LET'S MEEEEEEEERRRGGGEEEE!!!

(FYI I took the liberty to push one little nitpick before approving)

PR #1730, opened 2025-01-31
Upgrade puppeteer and other npm packages

derek-noredink:
We've got a bug in our continuous integration tests related to our very out of date puppeteer version. This should fix it right up, which our tests will verify. I decided to just upgrade all of the npm packages at once, because why not?

contributes to https://linear.app/noredink/issue/FXN-3732/build-of-noredink-ui-failing

### DIFF
--- a/lib/CustomElement.js
+++ b/lib/CustomElement.js
@@ -111,7 +111,7 @@ function noOp() {}
 exports.create = function create(config) {
   if (customElements.get(config.tagName)) {
     throw Error(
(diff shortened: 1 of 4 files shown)


derek-noredink:
Okay, so that's the exact same error as before 😅 

Perhaps this has to do with the host machine that's running the tests. Let's see how we can address that. 

Will still try and merge this pull request later because it's nice to keep dependencies up to date.

derek-noredink:
closing in favor of https://github.com/NoRedInk/noredink-ui/pull/1731

PR #1740, opened 2025-03-19
faster sortable table

michaelglass:
## What changes does this release include?

Adds a more performant sortable table by caching sorted data in the state.

easiest to review commit-by-commit

## How has the API changed?

Please paste the output of `elm diff` run on latest master in the code block:

```
This is a MINOR change.

    Nri.Ui.SortableTable.V6
```

Rejiggers the API for SortableTable a bit.

- State becomes opaque, and is renamed Model
- because of this, encoders and decoders are included for the Model
- because of caching & opaque model, implements TEA and exposes an `update` function
- in order to initialize the model, dev now needs to pass in all columns
- when the `SortableTable` has no model attribute, it is "Loading", else it is loaded.

### DIFF
--- a/component-catalog/src/Examples/SortableTable.elm
+++ b/component-catalog/src/Examples/SortableTable.elm
@@ -168,6 +168,7 @@ viewWithCode ({ sortModel } as model) =
     in
     ( \indentOffset ->
         ([ Code.fromModule moduleName "view"
+         , Code.record [ ( "model", "model" ), ( "msgWrapper", "identity" ) ]
          , Code.listMultilineFlat
             (List.concat
                 [ case settings.stickyHeader of
@@ -202,9 +203,6 @@ viewWithCode ({ sortModel } as model) =
 
                   else
                     []
-                , [ "SortableTable.msgWrapper identity"
-                  , "SortableTable.model model"
-                  ]
                 ]
             )
             (indentOffset + 2)
@@ -214,11 +212,11 @@ viewWithCode ({ sortModel } as model) =
         )
             |> Code.unstyledViewWithIndent (indentOffset + 1)
     , SortableTable.view
+        { model = sortModel
+        , msgWrapper = SortableTableMsg
+        }
         (List.concat
-            [ [ SortableTable.msgWrapper SortableTableMsg
-              , SortableTable.model sortModel
-              ]
-            , case settings.stickyHeader of
+            [ case settings.stickyHeader of
                 Nothing ->
                     []
 
--- a/component-catalog/src/Examples/SortableTable.elm
+++ b/component-catalog/src/Examples/SortableTable.elm
@@ -212,9 +212,7 @@ example =
                 SortableTable.viewLoading attrs columns
 
               else
-                SortableTable.view attrs
-                    columns
-
+                SortableTable.view attrs columns
             ]
     }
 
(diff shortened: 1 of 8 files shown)


bcardiff:
> @bcardiff do you want me to support pagination here?

Only if you find a nice way to 
* make it optional
* don't make it dependent of a specific pagination control

Another thing I recall is that SortableTable is for client only data, but it has some styling on headers. Is there a way to have a config that would make SortableTable work with a remote data source also? I think that supporting optional pagination and remote data source will allow us to design a better component api.

But, we can always iterate on a V7 with that and leave pagination and remote out of the picture for now.

component-catalog/src/Examples/SortableTable.elm
```diff
@@ -326,15 +314,15 @@ dataWithCode =
 
 {-| -}
 type alias State =
```
michaelglass, line 316: 💭 let's make this type opaque

bcardiff: Yes!

src/Nri/Ui/SortableTable/V6.elm
```diff
@@ -0,0 +1,708 @@
...
+                        ( model_.column, model_.sortDirection )
+                        (List.sortWith
+                            (entriesSorter (Model model_))
+                            entries
+                        )
+        }
+
+
+{-| -}
+init : id -> List (Column id entry msg) -> List entry -> Model id entry
```
bcardiff, line 302: Since the `init`, `rebuild` haves a mandatory `List entry` and the config can have a missing state it seems the user will be forced to do some tricky stuff to implement a "refresh button".

Should we make `init` and `rebuild` have a `Maybe (List entry)` instead?

I imagine:

- the app makes a request to load some entries remotely
- we don't have a table model yet.. ok. In the view I will not pass a model doing List.filerMap or something
- the app get the request
- we can initialize a table model now, with some choice of which column to sort with
- the user changes the sorting
- we have a refresh button to refetch things (but preserving the order) how we do that? 
  - we need to know that the records are in Loading state, and in that case we don't include a model
  - we then rebuild the table model when the records arrived
  - it works but is odd that we need to leave the table model out of sync. we could at least rebuild it with an empty list of entries...

But if the `List entry` becomes a `Maybe (List entry)` we can rebuild things more easily and straight forward from a RemoteData (as anything that is not Succeed is mapped to Nothing).

(Slightly odd that the loading state don't show the arrows in the table header)

Not 100% sure. WDYT?

src/Nri/Ui/SortableTable/V6.elm
```diff
@@ -0,0 +1,708 @@
...
+        }
+
+
+{-| -}
+type Model id entry
+    = Model
+        { column : id
+        , sortDirection : SortDirection
+        , entries : Dict.Dict ( id, SortDirection ) (List entry)
+        , sorter : id -> Sorter entry
```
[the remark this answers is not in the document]
michaelglass: can you elaborate here. I don't totally see what you'd do to make this better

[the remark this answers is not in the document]
bcardiff: If we do
  
```elm
type alias Columns id entry msg = List (Column id entry msg)

rebuild : Config data id -> Model id entry -> List entry -> Model id entry
view : Config data id -> List (Attribute id entry msg) -> Html msg -- flip the order of the first two arguments
update : Config data id -> Msg id -> Model id entry -> Model id entry
```

Then the model no longer needs the sorter. It deviates from other Nri.Ui components a little bit. But the config is usally fixed and bound to a specific "table instance", but we need to wire things manually for each rebuil / view / update.

Otherwise we could give more "essence" to a "table instance"

```elm
myTable = 
  let
    config = ...
  in
  { rebulid = SortableTable.rebuild config
  , view = SortableTable.view config
  , update = SortableTable.update config
  }

view = ...
  myTable.view ...
```

Again, nothing wrong with the current design. Just a thought.

PR #1575, opened 2023-12-02
in AssignmentIconds replace <path>s with <circle>s where applicable

ap-nri:
This is now ready to merge! All changes are approved, old versions of icons dropped, and updated `<svg>`s are in this readme as code (and can be exported with some judicious [Taco Bell programming](http://widgetsandshit.com/teddziuba/2010/10/taco-bell-programming.html)). 

~~HOLD UP NO MERGING YET
As of right now this has two versions of every applicable icon. When the icon name is suffixed with `X` that's actually the original version of the icon using a path. I include those because I want design to be able to compare side by side. Once they approve, I will make a commit removing all of them leaving only the versions with `<circle>`s.~~ So the checklist is:

- [x] design approves these changes
- [x] I make commit dropping all the old icons and their examples (suffixed with X)
- [x] I provide rendered SVG code of everything so that the master SVG files can also contain circles.

# :wrench: Modifying a component

## Context

Sometimes circular icons in in AssignmentIcons look unsharp around the outsides. This is because they do not use Circle or Ellipse elements, but Paths.

## :framed_picture: What does this change look like?

You can see little pixel imperfections in this, not gonna lie. In the old versions a lot of the circles were slightly elliptical or did not touch the sides of their bounding box. Maybe that actually works better because the shape inside of the circle looks more centered in those circles, and I get that. Now the circles touch all four sides of the bounding rectangle.

You can definitely get a better feel with a live page. [The preview](https://deploy-preview-1575--noredink-ui.netlify.app/#/doodad/AssignmentIcon) lets you blow up the icons pretty big (turn on "Show names" to know which is which). I can also send the HTML page that I've screenshot below. There's no good way to serve it to you other than as an image. Stinks. 

_Users_alexp_Developer_noredink-ui_icon-compare_main html (1)

## Rendered SVG code for each

<details><summary>Collapsed because it's a ton of code blocks that make this hard to read</summary>

### startPrimary 
```svg
[...]

PR #1746, opened 2025-04-21
Assignment icon updates

bendansby:
Updates and adds one new icon to AssignmentIcon


bcardiff:
@bendansby many (or all) of the Circled icons that are updated in this PR are loosing the benefits of https://github.com/NoRedInk/noredink-ui/pull/1575 since they are no longer using the `encircle` function.

I also noticed some clipPaths added, but no element with that id is defined. So I am not sure if things are looking as you would expect.

bendansby:
Thanks for the reminder about Alex's PR; I had forgotten that cleanup was needed there. I fixed the slightly off dimensions and replaced the paths with circles. The clipping paths are just an artifact of groups in Figma; looks fine but I went ahead and removed them for cleanliness.

bcardiff (approved):

PR #1753, opened 2025-04-25
Component standardization

bendansby:
This conforms the colors, weights, sizes, and spacing of several components to be more uniform and standardized.

Affected components:
- Button
- Menu
- Dotless radio button
- Radio button
- Segmented control
- Select
- Tabs
- Text input

### DIFF
--- a/src/Nri/Ui/Button/V10.elm
+++ b/src/Nri/Ui/Button/V10.elm
@@ -935,12 +935,12 @@ viewLabel config =
 viewIcon : ButtonSize -> List Css.Style -> NriSvg.Svg -> Html msg
 viewIcon size iconStyles svg =
     let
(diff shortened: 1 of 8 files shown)


micahhahn (approved):
Seems fine to me... are we worried about things falling out of alignment if they relied (perhaps implicitly) on these defaults? 

It's hard, if not impossible, to know all the downstream effects these changes might have.

bendansby:
Good thing to keep an eye out for on Percy etc. next version bump. I imagine it will be OK since our tolerances aren't generally that tight, but hard to know for sure as you say. Thanks for the review!

blunderdome:
@bendansby do you currently have a way to get a Percy run for these pre-merge?

PR #1755, opened 2025-05-02
Component tweaks

bendansby:
Some styling tweaks to components, mainly around font weights.


macoca:
Although the only place I can recall we use groups is to define the admin sections, removing the bold makes them quite hard to distinguish from the regular items

Screenshot 2025-05-02 at 3 48 44 PM

macoca:
My idea was to extend the SideNav groups and implement the Catalog sections with them. But we can have groups and sections as separate things if there's a use case for both. WDYT?

macoca (approved):

PR #1810, opened 2026-04-26
Polish Container example page.

bendansby:
## Context

The customizable examples on the Container component catalog page were demoing each variant with a generic "quick brown fox" pangram, which doesn't show off what each variant is actually for. This swaps in realistic NoRedInk-flavored content per variant, adopts our design-system text styles, and tightens the layout.

## :wrench: What changes

- **Realistic per-variant default content:** assignment card (Default), tip note (Gray), "Quick Write" prompt (Pillow), clickable practice card with badge (Buttony), locked unit (Disabled).
- **Content control is now optional** — the realistic defaults render until the user opts into the content control, instead of being permanently overridden by it.
- **Typography uses design-system styles** — `Text.smallBodyGray`, `Text.mediumBody`, `Text.smallBody`, `Text.caption`, and `Heading` instead of inline `Fonts` + `Css.fontSize` / `Css.color`.
- **Tighter, consistent spacing** — 32px between examples, 2px between section heading and blurb, paragraph margins flushed so the flex `gap` controls layout.
- **2-column grid** for the five customizable example cards.

_(add before/after screenshots — Container page showing the customizable examples grid)_

- [x] Changes are clearly documented (no API changes; component code unchanged)
- [x] Changes extend to the Component Catalog (this PR is component-catalog-only)
- [x] The component example still has an accurate preview, valid sample code, correct keyboard behavior (none), and guidance
- [x] No new major version
- [ ] Reviewers
  - [ ] Component library owner

## Notes for reviewers

- Only `component-catalog/src/Examples/Container.elm` changes; the published `Nri.Ui.Container.V2` module is untouched.
- The customizable controls still override the default content when toggled on, so the page stays useful for experimenting.

### DIFF
--- a/component-catalog/src/Examples/Container.elm
+++ b/component-catalog/src/Examples/Container.elm
@@ -16,9 +16,12 @@ import Debug.Control.View as ControlView
 import Example exposing (Example)
 import Html.Styled as Html exposing (Html)
 import Html.Styled.Attributes exposing (css)
+import Nri.Ui.Colors.V1 as Colors
 import Nri.Ui.Container.V2 as Container
+import Nri.Ui.Fonts.V1 as Fonts
 import Nri.Ui.Heading.V3 as Heading
 import Nri.Ui.Spacing.V1 as Spacing
+import Nri.Ui.Text.V6 as Text
 
 
 moduleName : String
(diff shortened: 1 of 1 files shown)


adauguet (approved):
Looks good @bendansby, thanks, sorry for the delay!

